PR #559, opened 2024-10-21
CPU refactoring

rdragos:
This PR mainly does two things:
1. Removes lots of old/dead code.
2. Renames files, gets rid of the `ng_/gr` structs/functions.


iris-mpc-cpu/src/prelude.rs
dkales, line 1: do we keep this empty file around?

dkales (approved):

iris-mpc-cpu/src/prelude.rs
rdragos: Nice catch, I've missed that one! Have deleted it in e701406be018db2473bc2fc0642afb1f3719af4c

PR #599, opened 2024-10-28
simpler heartbeat based on health

philsippl:
The NCCL heartbeat has always been a troublemaker and wasn't anyways testing the health of the NCCL connections that are actually used in the main thread. 

This now just uses the `/health` endpoint and exposes a unique id to the other nodes. If the other nodes notice the unique id has changed they restart.

Note: This requires a new config entry to be added (without port).
`SMPC__NODE_HOSTNAMES='["...", "...", "..."]'`


wojciechsromek:
@philsippl this will require some infrastructure work in the module, I can pick it up. 

It would make sense though to include some checks for DB connectivity / correct working of the queue processing checks, because otherwise this would basically be a liveness check, not healthiness one. Wdyt?

PR #605, opened 2024-10-30
Update Hawk Pack

iliailia:
This update brings additional API necessary to compute vector and distance operations on batches


rdragos (approved):
:1st_place_medal:

PR #606, opened 2024-10-30
Batch evaluation of distances in SMPC

iliailia:
Now `LocalNetAby3NgStoreProtocol` computes distances eagerly in `eval_distance` and `eval_distance_batch`, and supplies the results to `less_than` and `is_match` in secret shared form. As a result, `DistanceRef` contains secret shared distances, which resulted in a bit of refactoring.


iris-mpc-cpu/src/hawkers/galois_store.rs
```diff
@@ -151,22 +181,83 @@ impl VectorStore for LocalNetAby3NgStoreProtocol {
...
+            .map(|player_shares| player_shares.into_iter())
+            .collect::<Vec<_>>();
+        (0..vectors.len())
+            .map(|_| {
+                all_shares
+                    .iter_mut()
+                    .map(|player_shares| player_shares.next().unwrap())
+                    .collect::<Self::DistanceRef>()
+            })
+            .collect::<Vec<Self::DistanceRef>>()
```
dkales, line 249: Just to be clear, this is just an artifact of the local variant where all players are done by this party and this would not happen in a real network right?

iris-mpc-cpu/src/hawkers/galois_store.rs
```diff
@@ -133,10 +139,34 @@ impl LocalNetAby3NgStoreProtocol {
...
+    galois_ring_to_rep3(player_session, ds_and_ts)
+        .await
+        .unwrap()
+}
+
 impl VectorStore for LocalNetAby3NgStoreProtocol {
     type QueryRef = PointId; // Vector ID, pending insertion.
     type VectorRef = PointId; // Vector ID, inserted.
-    type DistanceRef = (PointId, PointId); // Lazy distance representation.
+    type DistanceRef = Vec<DistanceShare<u16>>; // Distance represented as shares.
```
dkales, line 169: This is the shared distance to all rotated copies of query?

iliailia: At the moment, it is only to one specific rotation (actually, to an original iris without rotations). We'll extend it later to pick the best distance among rotated copies.

iliailia: `Vec` here stands for shares of all parties (so it's length is 3).

dkales (approved):

iris-mpc-cpu/src/hawkers/galois_store.rs
```diff
@@ -184,43 +275,29 @@ impl VectorStore for LocalNetAby3NgStoreProtocol {
...
-                    ds_and_ts[1].clone(),
-                    ds_and_ts[2].clone(),
-                    ds_and_ts[3].clone(),
-                )
-                .await
-                .unwrap()
-            });
+        for share1 in distance1.iter() {
+            for share2 in distance2.iter() {
+                if share1.player == share2.player {
```
rdragos, line 282: why is this needed?

rdragos (approved):
Good work! Just a small comment

iris-mpc-cpu/src/hawkers/galois_store.rs
iliailia: We want to be sure that these share belong to the same player before doing local operations.

rdragos: I see. The code will change a bit after I'll finish up the session refactoring. Nevertheless, that's a problem for future us :smile:

PR #659, opened 2024-11-05
Reduce memory usage of plaintext HNSW indexes

bgillesp:
Previous implementation had some inefficiencies in the data representations used for a plaintext HNSW index:
- `PointId` used `usize` which is usually 64 bits, but only 32 bits are needed for a (4-)billion entry database; type is changed to use `u32` instead
- `FormattedIris` represented the "dot product" encoding of masked iris bits directly using `i8` values, plus an extra copy of the binary iris code mask, meaning iris codes represented in this way used 9 bits per geometric index instead of the necessary 2 from using a standard `IrisCode` object; `FormattedIris` is now `PlaintextIris`, a tuple struct wrapper for `IrisCode`
- `DistanceRef` used `(PointId, PointId)` to represent a lazy distance, which previously was 128 bits (64 bits using new `PointId`s), but only requires a `(u16, u16)` to represent a non-lazy distance as a numerator and denominator; implementation is changed to compute and store the plaintext distances directly

Overall reduction in space I believe should be around 4.5x for the iris code storage, and around 3x for the HNSW graph storage.

### DIFF
--- a/iris-mpc-cpu/src/hawkers/plaintext_store.rs
+++ b/iris-mpc-cpu/src/hawkers/plaintext_store.rs
@@ -1,5 +1,5 @@
 use hawk_pack::VectorStore;
-use iris_mpc_common::iris_db::iris::IrisCode;
+use iris_mpc_common::iris_db::iris::{IrisCode, MATCH_THRESHOLD_RATIO};
 
 #[derive(Default, Debug, Clone)]
 pub struct PlaintextStore {
@@ -85,7 +85,7 @@ impl PlaintextStore {
 impl VectorStore for PlaintextStore {
     type QueryRef = PointId; // Vector ID, pending insertion.
     type VectorRef = PointId; // Vector ID, inserted.
-    type DistanceRef = (PointId, PointId); // Lazy distance representation.
+    type DistanceRef = (u16, u16);
 
     async fn insert(&mut self, query: &Self::QueryRef) -> Self::VectorRef {
         // The query is now accepted in the store. It keeps the same ID.
@@ -98,14 +98,14 @@ impl VectorStore for PlaintextStore {
         query: &Self::QueryRef,
         vector: &Self::VectorRef,
     ) -> Self::DistanceRef {
-        // Do not compute the distance yet, just forward the IDs.
-        (*query, *vector)
+        let query_code = &self.points[*query as usize];
+        let vector_code = &self.points[*vector as usize];
+        query_code.data.distance_fraction(&vector_code.data)
     }
 
     async fn is_match(&self, distance: &Self::DistanceRef) -> bool {
-        let x = &self.points[distance.0 as usize];
-        let y = &self.points[distance.1 as usize];
-        x.data.0.is_close(&y.data.0)
+        let (a, b) = *distance; // a/b
+        (a as f64) < (b as f64) * MATCH_THRESHOLD_RATIO
     }
 
     async fn less_than(
@@ -113,8 +113,9 @@ impl VectorStore for PlaintextStore {
         distance1: &Self::DistanceRef,
         distance2: &Self::DistanceRef,
     ) -> bool {
-        let (cross_1, cross_2) = self.distance_computation(distance1, distance2);
-        (cross_1 - cross_2) < 0
+        let (a, b) = *distance1; // a/b
+        let (c, d) = *distance2; // c/d
+        (a as i32) * (d as i32) - (b as i32) * (c as i32) < 0
     }
 }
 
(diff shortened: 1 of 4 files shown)


iris-mpc-cpu/src/hawkers/plaintext_store.rs
```diff
@@ -1,129 +1,95 @@
...
         let point_id = self.points.len() - 1;
-        PointId(point_id)
+        point_id as PointId
     }
 
+    /// Compare two distances between pairs of iris codes
+    ///
+    /// Returns terms obtained by cross-multiplying numerators with opposite
+    /// denominators.
     pub fn distance_computation(
```
iliailia, line 62: It looks that this function isn't used anywhere. Plus, it is inconsistent with the new `DistanceRef` of `PlaintextStore`. You can safely delete it.

iris-mpc-cpu/src/hawkers/plaintext_store.rs
```diff
@@ -132,23 +98,24 @@ impl VectorStore for PlaintextStore {
         query: &Self::QueryRef,
         vector: &Self::VectorRef,
     ) -> Self::DistanceRef {
-        // Do not compute the distance yet, just forward the IDs.
-        (*query, *vector)
+        let query_code = &self.points[*query as usize];
```
iliailia, line 101: This recasts from `PointId` are a bit annoying. It might be worth to write a getter (just named `get`) for points that accepts only `PointId`.

iliailia (commented):
Thanks! Could you find out what is a real improvement in tests/benchmarks? You can run `bench_plaintext_hnsw` and monitor RAM usage with the `top` tool if you use Ubuntu.

iris-mpc-cpu/src/hawkers/plaintext_store.rs
bgillesp: Since `PointId`s are intended to be used as indices, I tried implementing `Index` and `IndexMut` for `Vec` indexed by `PointId`, which seems to simplify the notation as well.  Does this approach address the issue for you?

(I also added `From<u32>` and `From<usize>` to simplify some other conversions a bit.)

iris-mpc-cpu/src/hawkers/plaintext_store.rs
```diff
@@ -1,129 +1,95 @@
...
-    pub fn compute_real_distance(&self, other: &FormattedIris) -> f64 {
-        let hd = self.dot_on_code(other);
-        let mask_ones = (self.mask & other.mask).count_ones();
-        ((mask_ones as f64) - (hd as f64)) / (2. * mask_ones as f64)
+    /// Return the fractional Hamming distance with another PlaintextIris,
+    /// represented as the i16 dot product of associated masked-bit vectors
+    /// and the u16 size of the common unmasked region
+    pub fn dot_distance_fraction(&self, other: &Self) -> (i16, u16) {
+        let (code_distance, combined_mask_len) = self.distance_fraction(other);
+        let dot_product = combined_mask_len.wrapping_sub(2 * code_distance) as i16;
```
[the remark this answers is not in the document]
bgillesp: Not sure if it's described explicitly in this form in public resources, but it's not too hard to see directly:

The value `code_distance` gives the number of common unmasked bits which are different between the iris codes, and `combined_mask_len` gives the total number of unmasked bits.  Relative Hamming distance directly is the ratio of these two.  The dot product formulation of things computes just about this, but the dot product adds 1 for each unmasked bit in common, and subtracts 1 for each unmasked bit that is different.  To get this from the above two quantities, start with 1 for each unmasked bit (`combined_mask_len`), then subtract 2 for each such bit that is different (twice `code_distance`).

I can add a short additional comment to explain this inline, if that's sufficient?

bgillesp:
Just tested the memory usage in an example program `iris-mpc-cpu/examples/hnsw-ex.rs` which generates an index with 1,000 entries, using the `valgrind --tool=massif` command, and memory usage decreased by a factor of 4:
- `main` branch: `15.59 MB`
- This PR: `3.895 MB`

Estimated amounts for the iris code DB plus the HNSW layer 0 graph are respectively 15.168 MB and 3.456 MB, so these measurements align closely with the theory.

bgillesp:
Looking at some benchmarks now, it seems like this PR may also accelerate generation of plaintext HNSW indexes substantially -- I'm guessing it's a result of a) properly cached distance computations, and b) the actual computation of distances is faster because it uses bitwise ops over `u64`-packed bits rather than the `i8` dot product that's currently being computed.  I'm oddly getting some mixed results from the Criterion benchmarks (size 100 and 1k plaintext insertion timings massively accelerated by 1100% and 600%, size 10k showing up consistently slightly slower after the changes for some reason, 100k accelerated by 350%, ...) but anecdotally the 100k plaintext benchmark took about 15 minutes to run on my system with the old implementation, and about 2:15 with the new implementation.

iris-mpc-cpu/src/hawkers/plaintext_store.rs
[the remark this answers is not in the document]
iliailia: Yes, please. I understand the trick, but we'll leave this code at some point and some other guy in the future might find it confusing.

iliailia (approved):
Looks great!

PR #663, opened 2024-11-06
more robust handling of sequence id 

philsippl:
- merge delete and reset of id into a single db transaction
- try to recover from a state with out-of-sync sequence id

### DIFF
--- a/iris-mpc-store/src/lib.rs
+++ b/iris-mpc-store/src/lib.rs
@@ -296,13 +296,22 @@ DO UPDATE SET right_code = EXCLUDED.right_code, right_mask = EXCLUDED.right_mask
             .execute(&self.pool)
             .await?;
 
-        // We also need to reset the sequence to avoid gaps in the IDs.
+        Ok(())
+    }
+
+    pub async fn set_irises_sequence_id(&self, id: usize) -> Result<()> {
         sqlx::query("SELECT setval(pg_get_serial_sequence('irises', 'id'), $1 + 1, false)")
-            .bind(db_len as i64)
+            .bind(id as i64)
             .execute(&self.pool)
             .await?;
-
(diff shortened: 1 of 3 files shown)


iris-mpc-store/src/lib.rs
```diff
@@ -290,21 +290,43 @@ DO UPDATE SET right_code = EXCLUDED.right_code, right_mask = EXCLUDED.right_mask
...
+        tx.commit().await?;
         Ok(())
     }
 
+    pub async fn set_irises_sequence_id(&self, id: usize) -> Result<()> {
+        self.set_sequence_id(id, &self.pool).await
+    }
+
+    pub async fn get_irises_sequence_id(&self) -> Result<usize> {
+        let id: (i64,) = sqlx::query_as("SELECT currval(pg_get_serial_sequence('irises', 'id'))")
```
eaypek-tfh, line 324: when I tried this manually in pgAdmin, I got "currval has not been defined in this session" error. Then I used below query to get it working:

`SELECT last_value FROM "SMPC_prod_1".irises_id_seq;`

Let's test in stage and we can use this one if we see any errors 👍

eaypek-tfh (approved):

iris-mpc-store/src/lib.rs
philsippl: hmm then let's just go your proposed way directly?

philsippl: do we need to use the schema though?

eaypek-tfh: I don't think schema is needed here as we set the search path to schema already. but i needed it in pgAdmin

eaypek-tfh (approved):

PR #664, opened 2024-11-07
handle pod restarts gracefully

eaypek-tfh:
**Change**
- This PR captures pod restart signals and shuts down the service gracefully by waiting for last batch to be completed
- It also waits for all pending batch results to be reported

### DIFF
--- a/iris-mpc/src/bin/server.rs
+++ b/iris-mpc/src/bin/server.rs
@@ -109,6 +109,7 @@ fn preprocess_iris_message_shares(
     ))
 }
 
+#[allow(clippy::too_many_arguments)]
 async fn receive_batch(
     party_id: usize,
     client: &Client,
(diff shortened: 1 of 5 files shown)


iris-mpc/src/bin/server.rs
```diff
@@ -1095,6 +1102,11 @@ async fn server_main(config: Config) -> eyre::Result<()> {
         let dummy_shares_for_deletions = get_dummy_shares_for_deletion(party_id);
 
         loop {
+            if shutdown_handler.is_shutting_down() {
+                tracing::info!("Exiting from main loop gracefully due to shutdown signal...");
+                return Ok(());
+            }
+
```
philsippl, line 1109: Won't this now leave one batch unprocessed like this, since we start fetching the next batch already in the previous loop iteration?

philsippl: Should we maybe rather have this `is_shutting_down()` check around the `receive_batch` call (which fills the next batch), set the `next_batch` None, the the current batch process and exit the loop in the next iteration if `next_batch` is None?

philsippl (approved):

PR #667, opened 2024-11-07
Update with new hawk-pack

iliailia:
This change allows vector operations to mutate a vector store. It's needed for #618.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2121,7 +2121,7 @@ dependencies = [
 [[package]]
 name = "hawk-pack"
 version = "0.1.0"
-source = "git+https://github.com/Inversed-Tech/hawk-pack.git?rev=400db13#400db13435acbdb6500ed3c8584c66281bf9e000"
+source = "git+https://github.com/Inversed-Tech/hawk-pack.git?rev=4e6de24#4e6de24f7422923f8cccd8571ef03407e8dbbb99"
 dependencies = [
  "aes-prng 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "criterion",
--- a/iris-mpc-cpu/Cargo.toml
+++ b/iris-mpc-cpu/Cargo.toml
@@ -17,7 +17,7 @@ bytemuck.workspace = true
 dashmap = "6.1.0"
 eyre.workspace = true
 futures.workspace = true
-hawk-pack = { git = "https://github.com/Inversed-Tech/hawk-pack.git", rev = "400db13" }
+hawk-pack = { git = "https://github.com/Inversed-Tech/hawk-pack.git", rev = "4e6de24" }
 iris-mpc-common = { path = "../iris-mpc-common" }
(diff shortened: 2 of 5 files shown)


bgillesp:
What do the `eval_distance` and `eval_distance_batch` functions need to mutate the vector store for?  My feeling is that these should be considered "read-only" functions on the vector stores, and any state that does need to be updated should be stored somewhere else, or the updates should be handled outside these functions.  (I'm guessing it's an MPC/networking thing?)

iliailia:
> What do the `eval_distance` and `eval_distance_batch` functions need to mutate the vector store for? My feeling is that these should be considered "read-only" functions on the vector stores, and any state that does need to be updated should be stored somewhere else, or the updates should be handled outside these functions. (I'm guessing it's an MPC/networking thing?)

It's related to sessions. Many low-level MPC functions mutate PRNG within a `Session` object to generate randomness, zero-shares etc. If we want to generate as little PRF as possible (in other words, do as many operations as possible within a single Session), then we can attach a session to a vector store and invoke it when needed. This is what Dragos did here #618.

bgillesp:
Okay, I understand now.  The issue is that the `VectorStore` trait represents an object that doesn't just store the vectors, but is also responsible for processing the vectors, hence the MPC/networking state creeps in.  Right now it can't easily represent the idea that "vectors are read-only" for certain operations, because the mutability has to reflect the entire state, including anything on top needed for handling the processing.

I have an inkling that things might be a bit cleaner if we logically separate this into a separate object that handles interpreting/processing the HNSW graphs and vectors using the HNSW logic, so that the vector and graph data are properly independent of the processing -- but I think that's refactoring for another PR if we decide it's necessary.

bgillesp (approved):
Changes look fine -- but interested if you have thoughts on the potential refactoring I mentioned above to decouple the operation state/logic from the underlying data structures.

iliailia:
> Okay, I understand now. The issue is that the `VectorStore` trait represents an object that doesn't just store the vectors, but is also responsible for processing the vectors, hence the MPC/networking state creeps in. Right now it can't easily represent the idea that "vectors are read-only" for certain operations, because the mutability has to reflect the entire state, including anything on top needed for handling the processing.
> 
> I have an inkling that things might be a bit cleaner if we logically separate this into a separate object that handles interpreting/processing the HNSW graphs and vectors using the HNSW logic, so that the vector and graph data are properly independent of the processing -- but I think that's refactoring for another PR if we decide it's necessary.

We can add a mutable evaluator struct argument to all functions that might need interaction. This struct will contain all the mutable parts necessary for evaluation and can be specific for a certain implementation of `VectorStore` via traits.

dkales:
yeah I also think some struct like an IOContext or similar that is passed to functions that need it seems better

iliailia:
> yeah I also think some struct like an IOContext or similar that is passed to functions that need it seems better

Got it, let me do it in some future PR

PR #669, opened 2024-11-07
make comparison <= threshold

philsippl:


rw0x0 (approved):

dkales:
Context for the last fix: This was apparently always broken (passing a u32 to a kernel expecting a usize) and did not surface until we added another parameter

@philsippl Can you take a look at these changes and then merge?

PR #672, opened 2024-11-08
Feat/Python HNSW Bindings

bgillesp:
Add Python bindings for plaintext HNSW search to support data analysis using Python ecosystem tools.

WIP, current implementation supports basic iris code data types and insertion/search, but no serialization and deserialization from file.


iliailia (approved):
Looks good to me. Though I'm unsure what API might be missing; I guess Oded would be the best judge here.

PR #682, opened 2024-11-12
chore: Adjust startupProbe in production

wojciechsromek:
When the Aurora is scaled-down, the application can take as long as ~22 minutes to load the DB.
I wanted to avoid setting the `startupProbe` initialDelay any higher, but rather bumped the amount of accepted failures.

Right now the star threshold is set to 25 minutes, this should prevent us from killing the app immediately after a cold Aurora start.


kszarek (approved):

wojciechsromek:
Delivered in: https://github.com/worldcoin/iris-mpc/pull/683

PR #684, opened 2024-11-14
Add a ReSharing protocol to both re-share to existing parties and create shares for additional parties

dkales:
TODO: 
 - [ ] Tracing
 - [x] Robustness checks
 - [ ] Server does not shutdown once finished
 - [x] Common seed derivation, still hardcoded atm
 - [ ] Testing in staging, should probably work well since it is GRPC and should be able to just be behind a TLS reverse proxy without problems (as long as you enable http2 for the transport between RP and GRPC server (e.g., `schema=h2c` in traefik))

- **refactor: no longer have degree 2 galois ring sharing in library at all**
- **feat: add more points to exceptional sequence of galois sharing**
- **test: more exhaustive tests for galois rings**
- **feat: basic reshare protocol implemented**
- **feat: also send and check eye in reshare message**
- **Revert "feat: also send and check eye in reshare message"**
- **feat: include both eyes in the sent message**
- **feat: basic reshare client**
- **feat: basic server and seed binary for v2 dbs**
- **feat: working basic reshare client and server probably lacks robustness + tracing**


iris-mpc-upgrade/Cargo.toml
```diff
@@ -49,3 +61,7 @@ path = "src/bin/tcp_ssl_upgrade_client.rs"
 [[bin]]
 name = "seed-v1-dbs"
 path = "src/bin/seed_v1_dbs.rs"
+
+[[bin]]
+name = "seed-v2-dbs"
```
wojciechsromek, line 66: nit: probably we could use a better name here, matching the protocol instead of v1,v2

iris-mpc-upgrade/src/bin/reshare-client.rs
```diff
@@ -0,0 +1,119 @@
...
+            );
+        }
+        tracing::info!(
+            "Submitting reshare request for iris codes {} to {}",
+            db_chunk_start,
+            db_chunk_end
+        );
+
+        let request = iris_reshare_helper.finalize_reshare_batch();
+        let mut timeout = tokio::time::Duration::from_millis(100);
```
wojciechsromek, line 91: What do you think about parametrising the timeout here?

iris-mpc-upgrade/Cargo.toml
dkales: This really is a generic seed binary that produces valid, like-production DBs for SMPCv2.

iris-mpc-upgrade/src/bin/reshare-client.rs
```diff
@@ -99,7 +113,7 @@ async fn main() -> eyre::Result<()> {
                 x if x == IrisCodeReShareStatus::FullQueue as i32 => {
                     tokio::time::sleep(timeout).await;
                     // todo: linear backoff strategy ok?
```
carlomazzaferro, line 115: Should be fine

carlomazzaferro (approved):

deploy/stage/common-values-reshare-server-left.yaml
```diff
@@ -0,0 +1,128 @@
...
+      worker_processes  auto;
+      
+      error_log  /dev/stderr notice;
+      pid        /tmp/nginx.pid;
+      
+      events {
+        worker_connections  1024;
+      }
+      
+      http {
```
carlomazzaferro, line 102: @wojciechsromek feel free to take a look - I slightly reworked this from what we had in upgrade server

dkales: Is there a need to have left/right reshare server config? The protocol does both at the same time

Dockerfile
```diff
@@ -41,6 +41,8 @@ COPY --from=build-app /src/gpu-iris-mpc/target/x86_64-unknown-linux-gnu/release/
 COPY --from=build-app /src/gpu-iris-mpc/target/x86_64-unknown-linux-gnu/release/upgrade-server /bin/upgrade-server
 COPY --from=build-app /src/gpu-iris-mpc/target/x86_64-unknown-linux-gnu/release/upgrade-client /bin/upgrade-client
 COPY --from=build-app /src/gpu-iris-mpc/target/x86_64-unknown-linux-gnu/release/upgrade-checker /bin/upgrade-checker
+COPY --from=build-app /src/gpu-iris-mpc/target/x86_64-unknown-linux-gnu/release/reshare-server /bin/reshare-server
+COPY --from=build-app /src/gpu-iris-mpc/target/x86_64-unknown-linux-gnu/release/reshare-server /bin/reshare-client
```
dkales, line 45: this should read client

carlomazzaferro: thanks, fixed!

PR #685, opened 2024-11-15
fix the invalid transformation from < to <= in the threshold check

philsippl:


iris-mpc-gpu/tests/e2e.rs
```diff
@@ -235,14 +249,38 @@ mod e2e_test {
                         db.db[db_index].clone()
                     }
                     2 => {
+                        println!("Sending iris code on the threshold");
+                        let db_index = rng.gen_range(0..DB_SIZE / 10);
+                        if deleted_indices.contains(&(db_index as u32)) {
+                            continue;
+                        }
+                        let variation = rng.gen_range(-10..10);
```
dkales, line 257: idk if this 20 range is too large to reliably hit the exact threshold during the e2e tests?

philsippl: yeah worried the same, should we just make it `(-1,0,+1)`?

dkales (approved):

PR #688, opened 2024-11-15
verify loaded record count equals to db size

eaypek-tfh:
**Change**
- Ensure that we loaded the db into memory correctly by checking their sizes are the same


philsippl (approved):
lgtm, let's test on staging

PR #698, opened 2024-11-20
Basic gRPC networking for SMPC

iliailia:
This is the first stab at gRPC networking, so I expect the overall design might significantly change. At the moment, the design rationale goes as follows:
- each player creates an instance of `GrpcNetworking`, which contains 
  - its gRPC client connections to other parties,
  - queues of received messages from other parties. Those are implemented via `async_channel`.
- `GrpcNetworking` operates as a server to receive messages. See its implementation of the `PartyNode` gRPC service, which is defined in `party_node.proto`. This service contains only one method `send_message`. When invoked on the server side, it retrieves the async channel corresponding to the given sender ID and session ID and reroutes the message there.
- `GrpcNetworking` also implements the `Networking` trait, which provides `send` and `receive` methods to be called within SMPC protocols. The `send` method simply calls `send_message` as a client by wrapping up a message, sender ID and session ID into a gRPC request. The `receive` method simply retrieves the first message in the queue corresponding to the given session ID and sender ID.

The crucial changes are located in the following files (the rest is just refactoring): 
- `proto/party_node.proto`,
- `src/execution/local.rs`,
- `src/network/grpc.rs`

TODO:
- more robust session management
- TLS support
- testing in WAN
- server shutdown logic


dkales:
re TLS support, if you add the `tls` and `tls-native-roots` feature to tonic, it can do TLS out of the box for the client. For the server side I think it would always be deployed in a reverse-proxy setup (e.g., behind traefik) anyways so it would not need TLS support by itself

iris-mpc-cpu/src/execution/local.rs
```diff
@@ -17,6 +17,18 @@ pub fn generate_local_identities() -> Vec<Identity> {
...
+    Ok(listener.local_addr()?.port())
+}
+
+pub fn get_free_local_addresses(num_ports: usize) -> eyre::Result<Vec<String>> {
+    let free_ports = (0..num_ports)
+        .map(|_| get_free_port())
+        .collect::<eyre::Result<Vec<u16>>>()?;
+    Ok(free_ports.iter().map(|p| format!("[::1]:{p}")).collect())
+}
+
```
dkales, line 31: This does not seem to be very robust.
First, it drops the listener after getting its port, so it could be immediately reused, second it grabs ports on IPv4 and returns IPv6 address strings?

iris-mpc-cpu/src/network/grpc.rs
```diff
@@ -0,0 +1,491 @@
...
+use std::sync::Arc;
+use tonic::{
+    async_trait,
+    transport::{Channel, Server},
+    Request, Response, Status,
+};
+
+mod party_node {
+    tonic::include_proto!("party_node");
+}
```
dkales, line 23: I'm always more inclined to put the generated files in version control to be able to look at them, however, I get that this is the official recommended way of tonic.

iris-mpc-cpu/src/network/grpc.rs
```diff
@@ -0,0 +1,491 @@
...
+                .clone();
+            let request = Request::new(SendRequest {
+                sender_id:  self.party_id.0.clone(),
+                session_id: session_id.0,
+                data:       value.clone(),
+            });
+            println!(
+                "Sending message {:?} from {:?} to {:?}",
+                value, self.party_id, receiver
+            );
```
dkales, line 184: these should probably use tracing at a TRACE level

iris-mpc-cpu/src/network/grpc.rs
```diff
@@ -0,0 +1,491 @@
...
+    async fn receive(&self, sender: &Identity, session_id: &SessionId) -> eyre::Result<Vec<u8>> {
+        // Just retrieve the first message from the corresponding queue
+        self.message_queues
+            .get(session_id)
+            .ok_or_eyre(format!(
+                "Session {session_id:?} hasn't been added to message queues"
+            ))?
+            .pop_front(sender)
+            .await
+    }
```
dkales, line 207: Does this get + pop_front combo hold the dashmap entry lock until there is actual data from the network?
I think this only works atm because the async channel used internally does not need any mut references to self at all, right, and would not work with e.g., a tokio channel

iris-mpc-cpu/proto/party_node.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+package party_node;
+
+service PartyNode {
+    rpc SendMessage (SendRequest) returns (SendResponse);
+}
```
dkales, line 7: Some things to think about here:
1) Should we use unidirectional streaming gRPC for this, instead of request/response gRPC?
 (i.e., (`rpc SendMessage (stream SendRequest) returns (SendResponse);`) which can then be drained directly into the async_channel?) Probably a bit more complex with timeouts etc, but should hopefully be more performant.
2) should the message just be the data, and the sender_id and session_id be metadata that is sent via, headers etc? This would then also help a bit in the streaming case, since it is not part of every request?

iris-mpc-cpu/src/network/grpc.rs
iliailia: You're correct. Do you see a case where it might fail in our protocols?

iris-mpc-cpu/proto/party_node.proto
iliailia: Done with the latter. Regarding the former, I can give it a try in the next PR.

iris-mpc-cpu/src/execution/local.rs
iliailia: I added a static variable to keep track of used ports. Please, have a look.

iris-mpc-cpu/src/network/grpc.rs
dkales: I think it is fine due to the design of aysnc channel, but IDK the tradeoff that crate makes to be `&self` only compared to most channel impls that have `&mut self` signatures

iris-mpc-cpu/proto/party_node.proto
dkales: Yeah I think without the streaming it does not make too much sense to do it since it is just more annoying to grab the fields from the metadata, the only real advantage would be to have the metadata only send once for the streaming call instead of for each message

dkales:
re: formatting for the generated file, I would just put `#[rustfmt:skip]` and similar on the module since it will re-generate a new one on build that has changes otherwise.

iris-mpc-cpu/src/execution/local.rs
dkales: Oh I see, this is only intended to be used for testing?

iliailia: Yes

dkales (approved):

bgillesp:
Question about the `Cargo.lock` changes -- did these come from the new dependencies supporting the added functionality, or did the lockfile get regenerated at some point?  (Especially, I see a lot of existing libraries that just have bumped version numbers.)  If the latter, then possibly updating dependency versions might be better in its own PR since then it's easy to identify any incompatibilities that come from the updates.

iliailia:
> Question about the `Cargo.lock` changes -- did these come from the new dependencies supporting the added functionality, or did the lockfile get regenerated at some point? (Especially, I see a lot of existing libraries that just have bumped version numbers.) If the latter, then possibly updating dependency versions might be better in its own PR since then it's easy to identify any incompatibilities that come from the updates.

I didn't do it on purpose, but it looks that these are the updates from `cargo update`.

bgillesp:
> > Question about the `Cargo.lock` changes -- did these come from the new dependencies supporting the added functionality, or did the lockfile get regenerated at some point? (Especially, I see a lot of existing libraries that just have bumped version numbers.) If the latter, then possibly updating dependency versions might be better in its own PR since then it's easy to identify any incompatibilities that come from the updates.
> 
> I didn't do it on purpose, but it looks that these are the updates from `cargo update`.

Got it, makes sense -- thanks for reverting the updates.  Would you mind committing the `Cargo.lock` you get after building as well?  Currently the lockfile doesn't contain the new dependencies from the PR, so it gets modified to include those whenever you first compile.

iliailia:
> > > Question about the `Cargo.lock` changes -- did these come from the new dependencies supporting the added functionality, or did the lockfile get regenerated at some point? (Especially, I see a lot of existing libraries that just have bumped version numbers.) If the latter, then possibly updating dependency versions might be better in its own PR since then it's easy to identify any incompatibilities that come from the updates.
> > 
> > 
> > I didn't do it on purpose, but it looks that these are the updates from `cargo update`.
> 
> Got it, makes sense -- thanks for reverting the updates. Would you mind committing the `Cargo.lock` you get after building as well? Currently the lockfile doesn't contain the new dependencies from the PR, so it gets modified to include those whenever you first compile.

Done

iliailia:
Ok, Bryan agreed to move ahead, so I'll merge this one and push a new one with streaming.

iris-mpc-cpu/proto/party_node.proto
iliailia: Streaming seems to work, so I'll leave this metadata change here.

PR #702, opened 2024-11-21
manually manage serial id

philsippl:


iris-mpc-store/src/lib.rs
```diff
@@ -69,6 +69,7 @@ impl StoredIris {
 
 #[derive(Clone)]
 pub struct StoredIrisRef<'a> {
+    pub id:         usize,
```
dkales, line 72: should this directly be an i64?

dkales (approved):
I have some of these functionalities (inserting directly to chosen ID) also in my reshare protocol branch #684 . I can unify this and use this one once this is merged.

iris-mpc-store/src/lib.rs
philsippl: yeah I guess this should represent it as what it actually is in pg, updated.

PR #707, opened 2024-11-22
gRPC networking with streaming

iliailia:
Now parties stream outgoing messages via `send_message` RPC. In particular, they pass unbounded tokio channels to each other.

This changed the design a bit. Now RPC calls are a part of initialization when parties create sessions. Afterwards, they retrieve and send messages via tokio channels.

Benchmarking results

```bash
gr_ready_made_hnsw/gr-big-hnsw-insertions/1
                        time:   [3.6375 ms 3.6706 ms 3.7157 ms]
                        change: [-70.649% -69.314% -68.196%] (p = 0.00 < 0.05)
gr_ready_made_hnsw/gr-big-hnsw-searches/1
                        time:   [6.5096 ms 6.5808 ms 6.6451 ms]
                        change: [-71.982% -71.344% -70.588%] (p = 0.00 < 0.05)
gr_ready_made_hnsw/gr-big-hnsw-insertions/10
                        time:   [234.79 ms 237.39 ms 239.84 ms]
                        change: [-73.691% -73.370% -73.044%] (p = 0.00 < 0.05)
gr_ready_made_hnsw/gr-big-hnsw-searches/10
                        time:   [135.91 ms 138.04 ms 140.14 ms]
                        change: [-75.074% -74.443% -73.762%] (p = 0.00 < 0.05)
gr_ready_made_hnsw/gr-big-hnsw-insertions/100
                        time:   [3.5658 s 3.5850 s 3.6063 s]
                        change: [-74.067% -73.849% -73.632%] (p = 0.00 < 0.05)
gr_ready_made_hnsw/gr-big-hnsw-searches/100
                        time:   [2.9875 s 3.0093 s 3.0281 s]
                        change: [-73.951% -73.624% -73.345%] (p = 0.00 < 0.05)
```
The request/response network is extremely slow with datasets of >1000 elements. Hence, the below results cannot be compared.

```bash
gr_ready_made_hnsw/gr-big-hnsw-insertions/1000
                        time:   [9.0761 s 9.2220 s 9.3809 s]
gr_ready_made_hnsw/gr-big-hnsw-searches/1000
                        time:   [8.3919 s 8.4949 s 8.5961 s]
gr_ready_made_hnsw/gr-big-hnsw-insertions/10000
                        time:   [20.473 s 20.673 s 20.899 s]
gr_ready_made_hnsw/gr-big-hnsw-searches/10000
                        time:   [19.746 s 20.186 s 20.954 s]
```


iris-mpc-cpu/src/network/grpc.rs
```diff
@@ -43,76 +44,120 @@ impl MessageQueueStore {
...
+// within one session are sent in order and consecutively. Don't send messages
+// to the same player in parallel within the same session. Use batching instead.
 #[derive(Clone)]
 pub struct GrpcNetworking {
-    party_id:       Identity,
+    party_id:         Identity,
     // other party id -> client to call that party
-    clients:        Arc<DashMap<Identity, PartyNodeClient<Channel>>>,
-    message_queues: Arc<DashMap<SessionId, MessageQueueStore>>,
+    clients:          Arc<Mutex<HashMap<Identity, PartyNodeClient<Channel>>>>,
```
dkales, line 106: Is `Dashmap` not better suited for this? why do we move back to `Mutex<Hashmap>`?

dkales (commented):
Generic Comments: what happens when the connection dies e.g. due to a timeout? This could potentially try to re-establish the connection again, right? Maybe something for the future.

Also Idk if you have some benchmarks for this streaming vs normal request/response setup?

iris-mpc-cpu/src/network/grpc.rs
```diff
@@ -25,15 +33,8 @@ fn err_to_status(e: eyre::Error) -> Status {
...
-#[derive(Clone)]
-struct QueueChannel {
-    pub sender:   Arc<async_channel::Sender<Vec<u8>>>,
-    pub receiver: Arc<async_channel::Receiver<Vec<u8>>>,
-}
-
-#[derive(Clone)]
 struct MessageQueueStore {
-    queues: DashMap<Identity, QueueChannel>,
+    queues: DashMap<Identity, Mutex<Streaming<SendRequest>>>,
```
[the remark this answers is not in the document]
iliailia: I tried it without `Mutex`. The problem is that `Streaming` doesn't satisfy the `Sync` trait, so it doesn't know how to synchronize thread access.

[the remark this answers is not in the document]
dkales: Ah I see, maybe could be solved by a another channel and spawning a task that pulls from the channel and puts into streaming, but might be overkill here. Maybe investigate in the future if there are performance problems, especially with many queues.

iliailia:
> Generic Comments: what happens when the connection dies e.g. due to a timeout? This could potentially try to re-establish the connection again, right? Maybe something for the future.
> 
> Also Idk if you have some benchmarks for this streaming vs normal request/response setup?

I have only a basic timeout logic that returns an eyre error. Definitely, there should be something more sophisticated.

The benchmarks will be in a minute.

iris-mpc-cpu/src/network/grpc.rs
```diff
@@ -43,76 +44,120 @@ impl MessageQueueStore {
...
 #[derive(Clone)]
 pub struct GrpcNetworking {
-    party_id:       Identity,
+    party_id:         Identity,
     // other party id -> client to call that party
-    clients:        Arc<DashMap<Identity, PartyNodeClient<Channel>>>,
-    message_queues: Arc<DashMap<SessionId, MessageQueueStore>>,
+    clients:          Arc<Mutex<HashMap<Identity, PartyNodeClient<Channel>>>>,
+    // other party id -> outgoing streams to send messages to that party in different sessions
+    outgoing_streams: Arc<DashMap<Identity, OutgoingStream>>,
```
dkales, line 108: why is this identity -> session_id -> stream and incoming is session_id -> identity -> stream?
I think this should be unified?
also maybe it would be good to make this 1 jump only, by just using a key that combines both? contains the party as well?

iliailia:
> > Generic Comments: what happens when the connection dies e.g. due to a timeout? This could potentially try to re-establish the connection again, right? Maybe something for the future.

And tokio, indeed, does reconnections under the hood

iliailia:
Added benchmark results to the description. There is a 3-4x speed up, but it is still painfully slow.

iris-mpc-cpu/src/network/grpc.rs
iliailia: Yep, good point. Will try to fix it.

iliailia: Ok, I managed to simplify the dashmap of OutgoingStream, but for incoming ones I have to have the current setup to correctly initialize them before any message is sent.

dkales (approved):
Ok let's leave it at that, I think most of the networking speedups should probably come due to batching anyways? If we think gRPC is too much overhead, I can try to bench some TCP only solution to compare the perf.

iliailia:
> Ok let's leave it at that, I think most of the networking speedups should probably come due to batching anyways? If we think gRPC is too much overhead, I can try to bench some TCP only solution to compare the perf.

Yep, let's try to batch as much as possible before optimizing networking.

PR #724, opened 2024-11-25
Send tuples of values over network

iliailia:
This adds a new `NetworkValue` type that can contain different values of this enum. In this way, we can batch values of different types and send them in one message over the network, thus avoiding unnecessary rounds of communication.

### DIFF
--- a/iris-mpc-cpu/src/network/value.rs
+++ b/iris-mpc-cpu/src/network/value.rs
@@ -15,8 +15,6 @@ pub enum NetworkValue {
     VecRing16(Vec<RingElement<u16>>),
     VecRing32(Vec<RingElement<u32>>),
     VecRing64(Vec<RingElement<u64>>),
-    // Can contain network values of different types
-    Tuple(Vec<NetworkValue>),
 }
 
 impl NetworkValue {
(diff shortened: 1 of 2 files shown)


bgillesp:
Gave this a look, and the approach is effective, but the semantics make me a bit nervous.  Specifically, the existing `NetworkValue` enum cleanly represents "one of a fixed collection of concrete data types", while adding the `Tuple` variant now changes that to "a directed graph", which is a lot more difficult to reason about due to the recursive structure.

I stared at this for a bit to see if there was a solution which captures the "tuple de/serialization" goal without bringing that structure into the concrete enum datatype, and came up with an approach which just provides a "vector" version of the `to_network` and `from_network` methods.  This seems to simplify the usage patterns for things as well when you don't have to mix enum variants with tuple parsing logic.

In testing this out, I already completely re-implemented the functionality, so I made a commit in a separate branch that you can merge or adapt if it seems like a good approach: [branch bg/pr-724-alternative](https://github.com/worldcoin/iris-mpc/tree/bg/pr-724-alternative)

iliailia:
> Gave this a look, and the approach is effective, but the semantics make me a bit nervous. Specifically, the existing `NetworkValue` enum cleanly represents "one of a fixed collection of concrete data types", while adding the `Tuple` variant now changes that to "a directed graph", which is a lot more difficult to reason about due to the recursive structure.
> 
> I stared at this for a bit to see if there was a solution which captures the "tuple de/serialization" goal without bringing that structure into the concrete enum datatype, and came up with an approach which just provides a "vector" version of the `to_network` and `from_network` methods. This seems to simplify the usage patterns for things as well when you don't have to mix enum variants with tuple parsing logic.
> 
> In testing this out, I already completely re-implemented the functionality, so I made a commit in a separate branch that you can merge or adapt if it seems like a good approach: [branch bg/pr-724-alternative](https://github.com/worldcoin/iris-mpc/tree/bg/pr-724-alternative)

Very good point, thanks! It's much cleaner.

PR #728, opened 2024-11-26
Add readyness checks checking that the image is consistent

danielle-tfh:
### Notes
* add image name used as a check between nodes
* do not panic as with such long wait times between parties responses we might want to just see errors and not actually panic


iris-mpc/src/bin/server.rs
```diff
@@ -695,26 +697,44 @@ async fn server_main(config: Config) -> eyre::Result<()> {
...
-                            if is_ready_flag.load(Ordering::SeqCst) {
-                                "ready".into_response()
-                            } else {
-                                StatusCode::SERVICE_UNAVAILABLE.into_response()
+            let app =
+                Router::new()
+                    .route(
+                        "/health",
+                        get(move || async move {
+                            general_purpose::STANDARD.encode(&serialized_response)
```
philsippl, line 721: a bit awkward that we base64 the binary serialization, why not json serde json it?

danielle-tfh: Sure! Wanted to take advantage of some more rust quirks but makes sense to use JSON

iris-mpc/src/bin/server.rs
```diff
@@ -759,9 +776,21 @@ async fn server_main(config: Config) -> eyre::Result<()> {
                     );
                 }
 
-                let uuid = res.unwrap().text().await?;
+                let response = res.unwrap().text().await?;
+                let probe_response: ReadyProbeResponse = serde_json::from_str(&response)
+                    .expect("Deserialization of probe response failed");
```
philsippl, line 781: can be just 
```
let probe_response = res.unwrap().json::<ReadyProbeResponse>().await.expect("Deserialization of probe response failed");
```

philsippl (approved):

PR #731, opened 2024-11-28
load db chunks from s3

philsippl:


iris-mpc-store/src/s3_importer.rs
```diff
@@ -0,0 +1,259 @@
...
+        return Ok(vec![]);
+    }
+    let mut bytes = vec![0; byte_len];
+    hex::decode_to_slice(hex, &mut bytes)?;
+    Ok(bytes)
+}
+
+pub async fn last_snapshot_timestamp(store: &impl ObjectStore) -> eyre::Result<i64> {
+    store
+        .list_objects()
```
carlomazzaferro, line 85: Check this from @wojciechsromek 's PR: https://github.com/worldcoin/iris-mpc-db-exporter/pull/2/files#diff-713f849da8ca4d26c1bf4982ce4d41026a05f017bd78ad3c869c1c4bb51f0b08R63-R66 

I am not entirely sure that `listObectsV2` returns everything - if there are more than 1k rows, the result will be truncated, and the code needs to get a continuation token for it

carlomazzaferro: https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html#description

philsippl:
@eaypek-tfh please check if this is in line with the assumptions on the importer side

eaypek-tfh (approved):

PR #750, opened 2024-12-03
Optimize mock setup generation

iliailia:
It speeds up `lazy_random_setup` by generating distance shares in a naive way as additive shares (distance, 0, 0).

The current speed up for hnsw benchmarks is about 2x in the preprocessing stage with async channels.

### DIFF
--- a/iris-mpc-cpu/src/hawkers/galois_store.rs
+++ b/iris-mpc-cpu/src/hawkers/galois_store.rs
@@ -12,7 +12,10 @@ use crate::{
         compare_threshold_and_open, cross_compare, galois_ring_pairwise_distance,
         galois_ring_to_rep3,
     },
-    shares::share::{DistanceShare, Share},
+    shares::{
+        ring_impl::RingElement,
+        share::{DistanceShare, Share},
+    },
 };
 use aes_prng::AesRng;
 use hawk_pack::{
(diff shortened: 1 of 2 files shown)


iris-mpc-cpu/src/hawkers/galois_store.rs
```diff
@@ -253,9 +256,23 @@ impl VectorStore for LocalNetAby3NgStoreProtocol {
...
+            1 => Share::new(zero_elem, distance_elem),
+            2 => Share::new(zero_elem, zero_elem),
+            _ => panic!("Invalid player index"),
+        }
+    }
+
     async fn graph_from_plain(
         &mut self,
         graph_store: &GraphMem<PlaintextStore>,
+        recompute: bool,
```
bgillesp, line 275: Rename to `recompute_distances`?  The variable isn't used in a context that the long variable name is cumbersome, and it makes it clear at a glance what is recomputed.

iris-mpc-cpu/src/hawkers/galois_store.rs
```diff
@@ -286,10 +310,15 @@ impl LocalNetAby3NgStoreProtocol {
...
     /// vector store and graph, which are returned as well.
+    /// The network type is specified by the user.
+    /// A recompute flag is used to determine whether to recompute the distances
+    /// from stored shares. If recompute is set to false, the distances are
+    /// naively converted from plaintext.
     pub async fn lazy_random_setup<R: RngCore + Clone + CryptoRng>(
         rng: &mut R,
         database_size: usize,
         network_t: NetworkType,
+        recompute: bool,
```
bgillesp, line 321: Same as above re: `recompute_distances` argument name.

iris-mpc-cpu/src/hawkers/galois_store.rs
```diff
@@ -323,14 +354,15 @@ impl LocalNetAby3NgStoreProtocol {
     pub async fn lazy_random_setup_with_local_channel<R: RngCore + Clone + CryptoRng>(
         rng: &mut R,
         database_size: usize,
+        recompute: bool,
```
bgillesp, line 357: `recompute_distances` again.

bgillesp (approved):
Change looks fine -- quite refreshing to have the benchmarks run a bit more promptly.

There's one thing I'm confused about -- the `gr_ready_made_hnsw` benchmarks (i.e. the measured timings in the main loop) are running about 10x as fast after this change on my machine compared to before (just the one commit, checked against main).  Are you seeing this as well, and if so is there any explanation as to what's going on?

iliailia:
> Change looks fine -- quite refreshing to have the benchmarks run a bit more promptly.
> 
> There's one thing I'm confused about -- the `gr_ready_made_hnsw` benchmarks (i.e. the measured timings in the main loop) are running about 10x as fast after this change on my machine compared to before (just the one commit, compared against the performance on main). Are you seeing this as well, and if so is there any explanation as to what's going on?

You might have compared this one with benchmarks with gRPC networking from the main branch. This explains a 10x speed up factor.

PR #756, opened 2024-12-04
Add retrieving s3 shares via getObject

danielle-tfh:
### Notes
* add retrieving shares via S3 Objects. This is because URL's expire when the role credentials expire https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html#PresignedUrl-Expiration
* we will only release this once SS sends shares keys only

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -679,9 +679,9 @@ dependencies = [
 
 [[package]]
 name = "aws-smithy-types"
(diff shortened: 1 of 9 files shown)


eaypek-tfh (approved):
Let's also add the config for stage and test this e2e in stage 👍

PR #761, opened 2024-12-06
tls-aware upgrade client + config and docker compose 

carlomazzaferro:
Changelog:
- Deployment files for external party reshare servers
- Local setup with NGINX and TLS
- Removed the auto-generated clients, and implement a TLS aware client
- Documentation


dkales:
@carlomazzaferro Just FYI, there is an official way to pass tls config to the autogenerated client. I would probably do that instead of modifying the autogenerated client, since it is a hassle to do so again if the protobuf definitions change
I can quickly implement this way if you want.

carlomazzaferro:
> @carlomazzaferro Just FYI, there is an official way to pass tls config to the autogenerated client. I would probably do that instead of modifying the autogenerated client, since it is a hassle to do so again if the protobuf definitions change I can quickly implement this way if you want.

@dkales  hat would be wonderful, I struggled a bit with it yesterday. Feel free to push to the same branch

dkales:
@carlomazzaferro  I cannot really test it locally anymore with the new docker files since I do not have pull permissons on the docker repo, could you test this?

carlomazzaferro:
> @carlomazzaferro I cannot really test it locally anymore with the new docker files since I do not have pull permissons on the docker repo, could you test this?

On it

carlomazzaferro:
@dkales managed to get the ssh chain working, and the client you implemented also works flawlessly with it. Thank you! 

Feel free to give it another look, and I'll merge it.

PR #763, opened 2024-12-06
Update importer contract to make it compatible with exporter

eaypek-tfh:
**Change**
- This PR makes sure the contract between importer and exporter are the same
- It sets the necessary configs for buckets
- It adds the logic to avoid overriding already loaded records from DB with S3 records


iris-mpc-store/src/s3_importer.rs
```diff
@@ -95,12 +95,17 @@ fn hex_to_bytes(hex: &str, byte_len: usize) -> eyre::Result<Vec<u8>> {
 }
 
 pub async fn last_snapshot_timestamp(store: &impl ObjectStore) -> eyre::Result<i64> {
-    store
-        .list_objects()
-        .await?
+    let objects = store.list_objects().await?;
+    tracing::info!("All objects in db chunks s3: {:?}", objects);
```
philsippl, line 99: do we really want to log them all?

iris-mpc-store/src/s3_importer.rs
```diff
@@ -112,7 +117,8 @@ pub async fn fetch_and_parse_chunks(
     let chunks = store.list_objects().await.unwrap();
     stream::iter(chunks)
         .filter_map(|chunk| async move {
-            if chunk.ends_with(".bin") {
+            if chunk.ends_with(".csv") {
+                tracing::info!("Processing chunk: {}", chunk);
```
philsippl, line 121: same here

philsippl (approved):

iris-mpc-store/src/s3_importer.rs
eaypek-tfh: let me remove that one 👍

iris-mpc-store/src/s3_importer.rs
eaypek-tfh: I'd like to keep this one as it can help with future debugging. it's printing one by one

PR #774, opened 2024-12-09
Feat/tls aware reshare protocol

carlomazzaferro:


iris-mpc-upgrade/build.rs
```diff
@@ -1,6 +1,6 @@
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
-    println!("cargo:rerun-if-changed=protos/reshare.proto");
+    println!("cargo:rerun-if-changed=proto/reshare.proto");
```
dkales, line 3: the original here should be correct, but actually tonic_build already prints this internally so this line is not needed at all

dkales (approved):

iris-mpc-upgrade/build.rs
carlomazzaferro: thanks, will remove it

PR #779, opened 2024-12-09
Change importer to work with binary

eaypek-tfh:
**Change**
- This PR converts from hex + csv import format into binary format.
- This helps us get half the size of csv file for the same number of items and also avoids time spent on hex decoding.
- In stage, it gives us 66% decrease on import time

**Side change**
- Instead of calling s3 listObjects to get the chunks, we now generate the chunk paths out of timestamp file. Timestamp file's structure is changed to `{unixTime}_{batchSize}_{lastSerialId}`

### DIFF
--- a/.github/workflows/temp-branch-build-and-push.yaml
+++ b/.github/workflows/temp-branch-build-and-push.yaml
@@ -2,6 +2,8 @@ name: Branch - Build and push docker image
 
 on:
   push:
+    branches:
+      - "ertugrul/change-to-binary"
 
 concurrency:
   group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
--- a/.github/workflows/temp-branch-build-and-push.yaml
+++ b/.github/workflows/temp-branch-build-and-push.yaml
@@ -1,0 +1,45 @@
+name: Branch - Build and push docker image
+
+on:
+  push:
+
+concurrency:
+  group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
+  cancel-in-progress: true
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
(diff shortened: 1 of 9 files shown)


.github/workflows/temp-branch-build-and-push.yaml
```diff
@@ -0,0 +1,47 @@
+name: Branch - Build and push docker image
+
+on:
+  push:
+    branches:
+      - "ertugrul/change-to-binary"
```
eaypek-tfh, line 6: I'll keep this file version controlled actually so that we avoid copying it over for stage manual tests

danielle-tfh: Can we not just make this not based on a branch on based on anything but main?

iris-mpc-store/src/lib.rs
```diff
@@ -74,6 +76,43 @@ impl StoredIris {
...
+            id_bytes
+                .try_into()
+                .map_err(|_| eyre!("Failed to convert id bytes to i64"))?,
+        ) as i64;
+
+        // parse codes and masks
+        let left_code = extract_slice(bytes, &mut cursor, 25_600)?;
+        let left_mask = extract_slice(bytes, &mut cursor, 12_800)?;
+        let right_code = extract_slice(bytes, &mut cursor, 25_600)?;
+        let right_mask = extract_slice(bytes, &mut cursor, 12_800)?;
```
philsippl, line 106: maybe use the `IRIS_CODE_LENGTH` and `MASK_CODE_LENGTH` here instead (multiplied by `mem::size_of<u16>()`)

danielle-tfh (approved):

philsippl (approved):

.github/workflows/temp-branch-build-and-push.yaml
eaypek-tfh: This will then push a lot of images on each commit and most of them won't be used for stage testing. Our image size is pretty big 😅

PR #780, opened 2024-12-10
disable importer in prod

eaypek-tfh:
**Change**
- Use `SMPC__ENABLE_S3_IMPORTER` to run the s3 importer or not. Set false in prod, true in stage
- Make `min_last_modified_at` optional in store lib function. So that we pass `None` for only-db import  and execute old query w/o import. And pass `Some` for s3 import.

### DIFF
--- a/iris-mpc-store/src/lib.rs
+++ b/iris-mpc-store/src/lib.rs
@@ -167,7 +167,7 @@ impl Store {
     /// Stream irises in parallel, without a particular order.
     pub async fn stream_irises_par(
         &self,
-        min_last_modified_at: i64,
+        min_last_modified_at: Option<i64>,
         partitions: usize,
     ) -> impl Stream<Item = eyre::Result<StoredIris>> + '_ {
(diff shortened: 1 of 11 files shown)


carlomazzaferro (approved):

iris-mpc-store/src/lib.rs
```diff
@@ -179,14 +179,23 @@ impl Store {
...
+                .fetch(&self.pool)
+                .map_err(Into::into),
+                None => sqlx::query_as::<_, StoredIris>(
+                    "SELECT * FROM irises WHERE id BETWEEN $1 AND $2",
+                )
+                .bind(start_id as i64)
+                .bind(end_id as i64)
+                .fetch(&self.pool)
+                .map_err(Into::into),
+            };
```
eaypek-tfh, line 198: cargo fmt removes the curl brackets around Some and None which makes it less readable :/

danielle-tfh (approved):

PR #804, opened 2024-12-16
Load smaller partitioned chunks from s3 using multiple clients

eaypek-tfh:
**Change**
- This PR improves the loading time by 35%
- It partitions the data in s3 into prefixes - to utilize [this](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html)
- It brings ability to specify the number of s3 clients and lets each partition use a separate s3 client
- It uses smaller chunks 2000 => 200 items in a chunk


philsippl:
let's park this for now, I don't think all changes here are actually required

PR #817, opened 2024-12-16
Send less in cross_mul

iliailia:
When comparing two distances `d1` and `d2` we compute:

`d1.code_dot * d2.mask_dot - d1.mask_dit * d2.code_dot`.

Currently each party sends a message to another party per each product such that the outcome is in the 2-out-of-3 secret sharing form. This can be squashed into one message for the entire expression as follows:
1. Each SMPC product in ABY3 contains two phases - local computation of the 3-out-of-3 sharing and an interactive conversion to a 2-out-of-3 secret sharing.
2. Since 3-out-of-3 sharing is linear we can omit interactive conversion, compute the difference and switch to 2-out-of-3 sharing afterwards.

### DIFF
--- a/iris-mpc-cpu/src/protocol/ops.rs
+++ b/iris-mpc-cpu/src/protocol/ops.rs
@@ -126,15 +126,9 @@ pub(crate) async fn cross_mul(
     t1: Share<u32>,
     d2: Share<u32>,
     t2: Share<u32>,
-) -> eyre::Result<(Share<u32>, Share<u32>)> {
-    // Compute d1 * t2; t2 * d1
-    let mut exchanged_shares_a = Vec::with_capacity(2);
-    let pairs = [(d1, t2), (d2, t1)];
-    for pair in pairs.iter() {
-        let (x, y) = pair;
-        let res = session.prf_as_mut().gen_zero_share() + x * y;
-        exchanged_shares_a.push(res);
-    }
+) -> eyre::Result<Share<u32>> {
+    // Compute d1 * t2 - t2 * d1
+    let res_a = session.prf_as_mut().gen_zero_share() + &d1 * &t2 - &t1 * &d2;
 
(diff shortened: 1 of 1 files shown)


iris-mpc-cpu/src/protocol/ops.rs
```diff
@@ -126,48 +126,28 @@ pub(crate) async fn cross_mul(
...
             next_role,
             &session.session_id(),
         )
         .await?;
 
     let serialized_reply = network.receive(prev_role, &session.session_id()).await;
     let res_b = match NetworkValue::from_network(serialized_reply) {
-        Ok(NetworkValue::VecRing32(element)) => element,
+        Ok(NetworkValue::RingElement32(element)) => element,
         _ => return Err(eyre!("Could not deserialize VecRing16")),
```
bgillesp, line 147: Error message here should be `"Could not deserialize RingElement32"`

iris-mpc-cpu/src/protocol/ops.rs
```diff
@@ -126,48 +126,28 @@ pub(crate) async fn cross_mul(
     t1: Share<u32>,
     d2: Share<u32>,
     t2: Share<u32>,
-) -> eyre::Result<(Share<u32>, Share<u32>)> {
```
bgillesp, line 129: Doc comment on this function should say something like `Computes D1*T2 - D2*T1`

bgillesp (approved):
Nice, good find!  I feel like all of these little improvements in low-level communication costs will tend to add up over the course of an HNSW search.  Just a couple of comment/error message corrections, otherwise looks good to me.

iliailia:
Thanks, this one is tiny, but some other one is coming which reduces communication complexity by about 30%.

bgillesp:
30% improvements I also think will tend to add up. :exploding_head:

PR #878, opened 2025-01-05
[POP-2042] Anonymized stats through buckets

philsippl:


iris-mpc-gpu/src/dot/kernel.cu
```diff
@@ -51,6 +51,41 @@ extern "C" __global__ void openResults(unsigned long long *result1, unsigned lon
...
+
+            // Check if we are out of bounds for the query or db
+            if (queryIdx >= queryLength || dbIdx >= realChunkLen || !match)
+            {
+                continue;
+            }
+
+            // Save the corresponding code and mask dots for later (match distributions)
+            unsigned int match_distances_counter_idx = atomicAdd(&match_distances_counter[0], 1);
+            if (match_distances_counter[0] < 1000)
```
dkales, line 74: apart from the 1000, should this be match_distances_counter_idx here?

philsippl: yes it should, didn't push this yet

iris-mpc-gpu/tests/bitinject.rs
dkales, line 1: all of these tests should also now be able to be run in CI, because Roman aligned them to the e2e test.
Do we want to do that?

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1060,6 +1112,144 @@ impl ServerActor {
...
+
+        let mut total_distance_counter = 0;
+        for (i, device) in self.device_manager.devices().iter().enumerate() {
+            let counter = device.dtoh_sync_copy(&match_distances_counters[i]).unwrap();
+            total_distance_counter += counter[0];
+        }
+
+        tracing::info!("Matching distances collected: {}", total_distance_counter);
+
+        if total_distance_counter >= MATCH_DISTANCES_BUFFER_SIZE as u32 {
```
dkales, line 1143: Meta: Is the buffer size:
- small enough to provide frequent enough bucket stats
- large enough to be private enough

I feel the answer to 2 is yes, for 1 I do not know.

iris-mpc-gpu/src/dot/share_db.rs
```diff
@@ -700,26 +718,33 @@ impl ShareDB {
...
         streams: &[CudaStream],
+        results_idx: usize,
     ) -> CudaSlice<u32> {
         assert_eq!(len & 3, 0);
         let mut rand = unsafe {
             self.device_manager
                 .device(idx)
                 .alloc::<u32>(len >> 2)
-                .unwrap()
+                .unwrap() // TODO: fix, make this async
```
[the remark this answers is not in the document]
philsippl: ah thanks for the pointer, stumbled across this while implementing, some non async leftover

iris-mpc-gpu/tests/bitinject.rs
philsippl: ok great, let's run them then

iris-mpc-gpu/src/server/actor.rs
philsippl: yeah also not quite sure yet, let me get some numbers on "the average amount of partial matches per signup". Then we can estimate how long it takes.
FYI: also the number of buckets is more of a placeholder still, I think a bit higher resolution will be required.

iris-mpc-gpu/Cargo.toml
```diff
@@ -9,7 +9,7 @@ repository.workspace = true
 
 [dependencies]
 bincode = "1.3.3"
-cudarc = { version = "0.12", features = ["cuda-12020", "nccl"] }
+cudarc = { git = "https://github.com/worldcoin/cudarc-fork.git", features = ["cuda-12020", "nccl"] }
```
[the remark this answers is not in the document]
philsippl: The macro only defines up to 12 parameters: https://github.com/coreylowman/cudarc/blob/e768643c98ba0f62e0e1eb954a5bc3ea52c2b825/src/driver/safe/launch.rs#L356-L359

I've instead switched to the underlying implementation, but needed to export the function for that: https://github.com/worldcoin/cudarc-fork/commit/b6ba8da720217736d262c43d39260c7380530567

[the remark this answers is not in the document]
philsippl: We could just push upstream an impl with more params as a quick hack?

iris-mpc-gpu/tests/bitinject.rs
philsippl: @dkales they ran after resolving the merge conflicts, but the bit inject case failed.

dkales: will take a look

iris-mpc-gpu/Cargo.toml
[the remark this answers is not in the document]
dkales: I created a PR in cudarc that extends this to 20

carlomazzaferro (approved):

PR #897, opened 2025-01-09
Feat: even odd import from s3

eaypek-tfh:
**Change**
- This PR makes use of https://github.com/worldcoin/iris-mpc-db-exporter/pull/58 to load data into the limbs without having any other operations

### DIFF
--- a/iris-mpc-gpu/src/dot/share_db.rs
+++ b/iris-mpc-gpu/src/dot/share_db.rs
@@ -313,7 +313,7 @@ impl ShareDB {
 
         let a0_host = record
             .iter()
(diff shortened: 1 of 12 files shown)


iris-mpc-gpu/src/dot/share_db.rs
```diff
@@ -313,7 +313,7 @@ impl ShareDB {
 
         let a0_host = record
             .iter()
-            .map(|&x| ((x as i8) as i32 - 128) as i8)
+            .map(|&x: &u16| ((x >> 8) as i32 - 128) as i8)
```
philsippl, line 316: this now actually extracts also the "high byte" (same as below), which is wrong

philsippl (approved):

PR #898, opened 2025-01-09
Port local copy of HawkSearcher for application-specific customizations

bgillesp:
Creates a local copy of the `HawkSearcher` data structure that can be used for algorithm customizations specific to our MPC setting.  Local version is called just `HnswSearcher` to distinguish it from the `HawkPack` version, and to make the class more legible to anyone reading the codebase with less context.


iris-mpc-cpu/src/hnsw/searcher.rs
```diff
@@ -0,0 +1,515 @@
...
+pub use hawk_pack::data_structures::queue::{
+    FurthestQueue, FurthestQueueV, NearestQueue, NearestQueueV,
+};
+use hawk_pack::{GraphStore, VectorStore};
+use rand::RngCore;
+use rand_distr::{Distribution, Geometric};
+use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
+
+// specify construction and search parameters by layer up to this value minus 1
```
iliailia, line 16: nit: "Specify" as in all such comments below

iris-mpc-cpu/src/hnsw/searcher.rs
```diff
@@ -0,0 +1,515 @@
...
+    /// index of this array if `lc` is larger than the array size.
+    fn get_val(arr: &[usize; N_PARAM_LAYERS], lc: usize) -> usize {
+        arr[lc.min(N_PARAM_LAYERS - 1)]
+    }
+}
+
+/// An implementation of the HNSW algorithm.
+///
+/// Operations on vectors are delegated to a VectorStore.
+/// Operations on the graph are delegate to a GraphStore.
```
iliailia, line 166: "are delegated"

iris-mpc-py/src/py_hnsw/pyclasses/hawk_searcher.rs
```diff
@@ -1,14 +1,13 @@
...
+    hnsw::searcher::{HnswParams, HnswSearcher, N_PARAM_LAYERS},
+    py_bindings,
 };
-use iris_mpc_cpu::py_bindings;
 use pyo3::{exceptions::PyIOError, prelude::*};
 
 #[pyclass]
 #[derive(Clone, Default)]
-pub struct PyHawkSearcher(pub HawkSearcher);
+pub struct PyHawkSearcher(pub HnswSearcher);
```
iliailia, line 10: Does it makes sense to use "PyHnswSearcher" instead?

iris-mpc-py/src/py_hnsw/pyclasses/hawk_searcher.rs
```diff
@@ -20,17 +19,17 @@ impl PyHawkSearcher {
...
     #[staticmethod]
     pub fn new_uniform(M: usize, ef: usize) -> Self {
-        let params = HawkerParams::new_uniform(ef, M);
-        Self(HawkSearcher { params })
+        let params = HnswParams::new_uniform(ef, M);
+        Self(HnswSearcher { params })
     }
 
-    /// Construct `HawkSearcher` with fully general parameters, specifying the
+    /// Construct `IrisSearcher` with fully general parameters, specifying the
```
iliailia, line 32: Should it be "HnswSearcher"?

iliailia (approved):
Thanks! Left a few minor comments.

iris-mpc-py/src/py_hnsw/pyclasses/hawk_searcher.rs
bgillesp: Probably does!  I was hedging on it because this is part of the Python class structure that gets exposed in the external interface, but decided to make the change anyway.  Easy fix for any Python scripts that use it is to add an `as` clause to the import.

iris-mpc-py/src/py_hnsw/pyclasses/hawk_searcher.rs
bgillesp: Left over from a previous naming convention, thanks for catching it!

PR #908, opened 2025-01-10
Add graceful shutdown to all health check failures

danielle-tfh:
### Notes
* enforce sending graceful shutdown signals to all nodes
* argo CD only triggers graceful shutdown every 3 minutes meaning during roll outs we do not see all 3 pods having graceful shutdown
* this shares between pods to start gracefully shutting down

### DIFF
--- a/iris-mpc/src/bin/server.rs
+++ b/iris-mpc/src/bin/server.rs
@@ -663,7 +663,9 @@ async fn main() -> eyre::Result<()> {
 }
 
 async fn server_main(config: Config) -> eyre::Result<()> {
-    let shutdown_handler = ShutdownHandler::new(config.shutdown_last_results_sync_timeout_secs);
+    let shutdown_handler = Arc::new(ShutdownHandler::new(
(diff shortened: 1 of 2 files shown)


iris-mpc/src/bin/server.rs
```diff
@@ -777,27 +779,47 @@ async fn server_main(config: Config) -> eyre::Result<()> {
     let is_ready_flag = Arc::new(AtomicBool::new(false));
     let is_ready_flag_cloned = Arc::clone(&is_ready_flag);
 
-    #[derive(Serialize, Deserialize)]
+    #[derive(Debug, Serialize, Deserialize, Clone)]
     struct ReadyProbeResponse {
         image_name: String,
         uuid:       String,
+        shutdown:   bool,
```
eaypek-tfh, line 786: a nit suggestion on naming:
```suggestion
        shutting_down:   bool,
```

eaypek-tfh (approved):

PR #932, opened 2025-01-14
Batch hash

philsippl:
Calculates a hash out of the signup ids of the current batch and syncs this hash across the node. This makes sure the queues are actually in sync before starting processing. This just throws an error and the actor can't recover it on its own requiring manual intervention.

### DIFF
--- a/iris-mpc-gpu/src/server/actor.rs
+++ b/iris-mpc-gpu/src/server/actor.rs
@@ -1499,7 +1499,7 @@ impl ServerActor {
             .collect();
 
         // Only keep entries that are valid on all nodes
-        let mut valid_merged = vec![false; max_batch_size];
+        let mut valid_merged = vec![true; max_batch_size];
         for i in 0..self.comms[0].world_size() {
             for j in 0..max_batch_size {
                 valid_merged[j] &= results[i][j] == 1;
--- a/iris-mpc-gpu/src/server/actor.rs
+++ b/iris-mpc-gpu/src/server/actor.rs
@@ -27,6 +27,7 @@ use eyre::eyre;
 use futures::{Future, FutureExt};
 use iris_mpc_common::{
     galois_engine::degree4::{GaloisRingIrisCodeShare, GaloisRingTrimmedMaskCodeShare},
+    helpers::sha256::sha256_bytes,
     iris_db::iris::IrisCode,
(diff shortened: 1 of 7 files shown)


iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1458,53 +1465,60 @@ impl ServerActor {
...
             .collect();
 
-        tracing::info!(party_id = self.party_id, "sync_batch_entries end");
+        // Only keep entries that are valid on all nodes
+        let mut valid_merged = vec![false; max_batch_size];
+        for i in 0..self.comms[0].world_size() {
+            for j in 0..max_batch_size {
+                valid_merged[j] &= results[i][j] == 1;
+            }
+        }
```
danielle-tfh, line 1507: Should we still keep this or fail? seems if we remove some entries than we risk them not being the same signupIds?

philsippl: But this can happen all the time. A client can even force this by sending us a wrongly encrypted share.

philsippl: This code has been there actually also before, just slightly rewritten, I can also restore the old version

danielle-tfh (approved):

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1458,53 +1465,60 @@ impl ServerActor {
...
         let results: Vec<_> = results
             .chunks_exact(results.len() / self.comms[0].world_size())
             .collect();
 
-        tracing::info!(party_id = self.party_id, "sync_batch_entries end");
+        // Only keep entries that are valid on all nodes
+        let mut valid_merged = vec![false; max_batch_size];
+        for i in 0..self.comms[0].world_size() {
+            for j in 0..max_batch_size {
+                valid_merged[j] &= results[i][j] == 1;
```
eaypek-tfh, line 1505: Given that we first set everything to `false` and then we apply `&=` operation, wouldn't it always resolve to `false`? 🤔

philsippl: you're right, and this also shows an issue with our e2e tests. Since this means we don't return any results, there are no mismatches.

eaypek-tfh (approved):

PR #948, opened 2025-01-16
[POP-2033] config and initial impl for LUC enablement

carlomazzaferro:


iris-mpc/src/bin/server.rs
```diff
@@ -138,6 +138,7 @@ async fn receive_batch(
     shares_encryption_key_pairs: SharesEncryptionKeyPairs,
     shutdown_handler: &ShutdownHandler,
     error_result_attributes: &HashMap<String, MessageAttributeValue>,
+    store_len: usize,
```
eaypek-tfh, line 141: I think `store_len` is calculated only once during the app startup. For LUC case, it'd be better to update it before receiving the batch, right?

iris-mpc-common/src/helpers/smpc_request.rs
```diff
@@ -111,6 +111,7 @@ pub struct UniquenessRequest {
     pub signup_id:               String,
     pub s3_key:                  String,
     pub iris_shares_file_hashes: [String; 3],
+    pub or_rule_serial_ids:      Option<Vec<u32>>,
```
eaypek-tfh, line 114: Just a reminder: I think it would be good to calculate the maximum number of serial ids that we can pass given that SQS has 256 MB limit

eaypek-tfh (approved):

iris-mpc/src/bin/server.rs
```diff
@@ -285,6 +286,24 @@ async fn receive_batch(
...
                             tracing::info!("Updating batch size to {}", batch_size);
                         }
+                        if config.luc_enabled {
+                            if config.luc_lookback_records > 0 {
+                                batch_query.or_rule_serial_ids.push(
+                                    ((store_len as u32 - config.luc_lookback_records as u32)
+                                        ..store_len as u32)
+                                        .collect::<Vec<u32>>(),
+                                );
+                            }
```
philsippl, line 296: as @eaypek-tfh points out, `store_len` is not updated. However we also don't have access here to `current_db_sizes` in the actor. I think i's probably simplest to just pass the `luc_lookback_records` as a parameter in the batch and push to the records in the actor?

iris-mpc/src/bin/server.rs
philsippl: re `store_len`: yes that's right, commented below.
re batch matches: hmm I guess the argument here would be then to just always switch to the OR rule in the batch?

philsippl: (assuming lookback is always larger than 64)

eaypek-tfh: > just always switch to the OR rule within the batch?
i think this would be a nice way to solve this 👍

iris-mpc/src/bin/server.rs
carlomazzaferro: Going with this suggestion, will implement the fix. Ideally we have a way of passing both (a set of serial ids and a lookback) and then merge them

PR #949, opened 2025-01-16
Fix creation of gRPC sessions

iliailia:
This checks outgoing channels instead of incoming ones. Otherwise, the asynchronous creation of a session by different parties would fail. A test with consecutive (and thus inherently asynchronous) session creation is added.


dkales:
we have actually hit this while experimenting with this GRPC networking implementation as well, wanted to bring this to you today, nice that you already found this already

PR #951, opened 2025-01-16
[POP-2034] LUC implementation on iric-mpc

carlomazzaferro:
We introduce the ability to perform or-rule matching on left / right results. We add the possibility of the batch processing to receive a parameter `or_rule_serial_ids`, which instructs the kernel code to perform or-rule matching against those serial ids.

The layout of the `or_rule_serial_ids` is a Vector of size batch_size, where each element is a Vector of serial ids. This is then allocated into GPU memory as a 1D bitmap of size batch_size * db_length, where each bit corresponds to a serial id against which the comparisons are carried out. If the bit is set, we perform OR rule. Otherwise, the default AND rule is used.


iris-mpc-gpu/src/dot/distance_comparator.rs
```diff
@@ -347,4 +408,16 @@ impl DistanceComparator {
...
             .collect::<Vec<_>>()
     }
+
+    pub fn prepare_luc_bitmap(&self, db_size: usize) -> Vec<CudaSlice<u64>> {
+        let row_stride64 = (db_size + 63) / 64;
+        (0..self.device_manager.device_count())
+            .map(|i| {
+                self.device_manager
+                    .device(i)
+                    .alloc_zeros(row_stride64 * self.query_length)
```
philsippl, line 418: why multiply by `self.query_length` here?

iris-mpc-gpu/src/dot/kernel.cu
```diff
@@ -93,6 +93,56 @@ extern "C" __global__ void mergeDbResults(unsigned long long *matchResultsLeft,
...
+                if (queryMatchCounter < MAX_MATCHES_LEN)
+                    partialResultsLeft[MAX_MATCHES_LEN * queryIdx + queryMatchCounter] = dbIdx;
+            }
+            if (matchRight)
+            {
+                unsigned int queryMatchCounter = atomicAdd(&matchCounterRight[queryIdx], 1);
+                if (queryMatchCounter < MAX_MATCHES_LEN)
+                    partialResultsRight[MAX_MATCHES_LEN * queryIdx + queryMatchCounter] = dbIdx;
+            }
+            size_t rowIndex = queryIdx * rowStride64;
```
philsippl, line 125: why is this `rowIndex` needed? If I'm not very mistaken, we iterate over the full db here, so no chunks anymore. 
`dbIndex / 64` should give you already the right element, and `dbIndex % 64` the correct bit in it.

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -340,6 +341,9 @@ impl ServerActor {
         let batch_match_list_left = distance_comparator.prepare_db_match_list(n_queries);
         let batch_match_list_right = distance_comparator.prepare_db_match_list(n_queries);
 
+        let or_policy_bitmap =
+            distance_comparator.prepare_luc_bitmap(max_db_size / device_manager.device_count());
```
philsippl, line 345: That's already the more optimized version, we just need to watch out later to also split the ranges for the indices correctly. The most trivial solution would otherwise be to just allocate `max_db_size` on all devices and have the same (unnecessarily) large bitmap on each.

iris-mpc-gpu/src/dot/distance_comparator.rs
carlomazzaferro: As discussed: we allocate a 2D array, flattened to a 1D bitmap for each of the queries, as each query will have (in the future) its own set of serial ids against which the OR ruling is used

iris-mpc-gpu/src/dot/kernel.cu
carlomazzaferro: Same reason as previous comment

iris-mpc-gpu/src/dot/kernel.cu
```diff
@@ -93,6 +93,67 @@ extern "C" __global__ void mergeDbResults(unsigned long long *matchResultsLeft,
     }
 }
 
+extern "C" __global__ void mergeDbResultsWithOrPolicyBitmap(unsigned long long *matchResultsLeft, unsigned long long *matchResultsRight, unsigned int *finalResults, size_t queryLength, size_t dbLength, unsigned int *matchCounter, unsigned int *allMatches, unsigned int *matchCounterLeft, unsigned int *matchCounterRight, unsigned int *partialResultsLeft, unsigned int *partialResultsRight, const unsigned long long *orPolicyBitmap) // 2D bitmap stored as 1D
+{
+
+    size_t rowStride64 = (dbLength + 63) / 64;
+
+    size_t totalBits   = dbLength * queryLength;
+    size_t numElements = (totalBits + 63) / 64; //  div_ceil
```
carlomazzaferro, line 102: @philsippl check if this is acceptable, was the only way I found to remove another argument from the function call

iris-mpc-gpu/src/dot/distance_comparator.rs
philsippl: as discussed, this should be divided by `ROTATIONS` because `query_length` is multiplied by it: https://github.com/worldcoin/iris-mpc/blob/f4dbe47c9beaf2806369ee0bec5ffb60b4c9d222/iris-mpc-gpu/src/server/actor.rs#L221

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1565,6 +1586,37 @@ impl ServerActor {
...
+                    }
+                }
+            }
+
+            // Copy the host-side bitmap into the pre-allocated GPU memory
+            self.device_manager
+                .device(device_idx)
+                .htod_copy_into(bitmap, &mut self.or_policy_bitmap[device_idx])
+                .expect("Failed to populate OR policy bitmap");
+        }
```
philsippl, line 1618: I think you need here to htod copy on the same as the join (`streams[0]`). You can use this helper for it: `htod_on_stream_sync`.
Otherwise you might not await it.

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1565,6 +1586,37 @@ impl ServerActor {
...
+                for &db_idx in db_indices {
+                    if db_idx < db_length as u32 {
+                        let row_start = query_idx * row_stride64;
+                        let word_idx = row_start + (db_idx as usize / 64);
+                        let bit_offset = db_idx as usize % 64;
+                        bitmap[word_idx] |= 1 << bit_offset;
+                    }
+                }
+            }
+
```
philsippl, line 1612: so here this comment becomes relevant: https://github.com/worldcoin/iris-mpc/pull/951/files#r1919178799

The db is sharded across the available devices, so we can't just allocate `db_length` (the sharded size) and then set db indices that are out of range. 

We can either allocate the full db size on each device and waste space or try to be smart about which index can even appear on which device. Serial ids are on arrange on the devices like this:

| Device 0 | Device 1 | Device 2 | Device 3 | Device 4 | Device 5 | Device 6 | Device 7 |
|----------|----------|----------|----------|----------|----------|----------|----------|
| 0        | 1        | 2        | 3        | 4        | 5        | 6        | 7        |
| 8        | 9        | 10       | 11       | 12       | 13       | 14       | 15       |
| ...      | ...      | ...      | ...      | ...      | ...      | ...      | ...      |

iris-mpc-gpu/src/server/actor.rs
carlomazzaferro: can `htod_sync_copy_into` instead be used?

iris-mpc-gpu/src/server/actor.rs
carlomazzaferro: Alright, this makes lots of sense. Will proceed with allocating max_db_size everywhere and then we optimize 👍

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1586,8 +1583,23 @@ impl ServerActor {
...
+        &mut self,
+        or_rule_serial_ids: Vec<Vec<u32>>,
+        batch_size: usize,
+    ) -> Vec<CudaSlice<u64>> {
+        let streams = self
+            .device_manager
+            .devices()
+            .iter()
+            .map(|dev| dev.fork_default_stream().unwrap())
+            .collect::<Vec<_>>();
```
philsippl, line 1596: no need to create new streams, you can just pass them into the function or use directly `self.streams[0]`. If you create new streams, it will just run async with the other stuff, which we don't want.

iris-mpc-gpu/tests/e2e.rs
```diff
@@ -249,13 +249,16 @@ mod e2e_test {
...
             for idx in 0..batch_size {
                 let request_id = Uuid::new_v4();
+                let mut use_or_rule_for_serial_ids: Vec<u32> = vec![];
                 // Automatic random tests
                 let options = if responses.is_empty() {
                     3
                 } else if deleted_indices_buffer.is_empty() {
                     4
-                } else {
+                } else if rng.gen() {
```
dkales, line 258: is there a reason why this is only taken 50% of the time?

iris-mpc-gpu/tests/e2e.rs
```diff
@@ -360,15 +363,54 @@ mod e2e_test {
...
-                        // one of them below the threshold and the other
-                        // above the threshold against the other codes. It will
-                        // also be required to set the param BatchQuery.threshold_bitmap_or to true
-                        // for the specific iris codes to be matched against
-                        // 5 => {
-                        // println!("Sending iris codes that match on right but not left with the OR
-                        // rule set");
+                        5 => {
+                            println!(
+                                "Sending iris codes that match on right but not left with the OR \
```
dkales, line 368: This println no longer fits with the logic below

iris-mpc-gpu/tests/e2e.rs
```diff
@@ -360,15 +363,54 @@ mod e2e_test {
...
+                            assert_eq!(code_left.mask, IrisCodeArray::ONES);
+                            assert_eq!(code_right.mask, IrisCodeArray::ONES);
+
+                            // apply variation to either right of left code
+                            let flip_right: bool = rng.gen();
+                            let variation = 1;
+
+                            if flip_right {
+                                // Flip bits to below threshold
+                                for i in 0..(THRESHOLD_ABSOLUTE as i32 - variation) as usize {
```
dkales, line 400: I think this actually flips to few bits, since the variation is positive so we only flip less than the Threshold. This means that these would also match using the AND rule if I am not mistaken.

dkales: Maybe a better test case would be to add the query to the batch twice, once with AND rule, once with OR rule and check the two results are as expected. However, IDK what the handling of the OR rule with the inter-batch checks is yet, so IDK if this would even work out.

iris-mpc-gpu/tests/e2e.rs
```diff
@@ -360,15 +363,54 @@ mod e2e_test {
...
+                                    code_right.code.flip_bit(i);
+                                }
+                            } else {
+                                for i in 0..(THRESHOLD_ABSOLUTE as i32 - variation) as usize {
```
dkales, line 404: same here

dkales (commented):
leaving a few comments now, still reviewing rest

iris-mpc-gpu/tests/e2e.rs
carlomazzaferro: See above: https://github.com/worldcoin/iris-mpc/pull/951/files/424926dbd45f45c6bb1ae96a948cdab185c42aad#diff-0c1888f81c1fbf6d412eaebb8144b3f97d10ae89d279ca4153b4f900003a69bdR321-R330

When variation is < 0 (-1) we run:

```
for i in 0..(THRESHOLD_ABSOLUTE as i32 + variation) as usize {
                                code.code.flip_bit(i);
                            }
```

Which is the same as I have here, since I set variation fixed to `let variation = 1`

I did verify that the results do indeed come as `true` when flipping the variation value when running the tests, fwiw.

But I will think a bit more deeply on how to implement a more comprehensive test

iris-mpc-gpu/tests/e2e.rs
carlomazzaferro: i'll improve the wording to match the implementation, thanks1

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1565,6 +1583,36 @@ impl ServerActor {
...
+                    let bit_offset = db_idx as usize % 64;
+                    bitmap[word_idx] |= 1 << bit_offset;
+                }
+            }
+            // Transfer the bitmap to the device
+            let _bitmap = htod_on_stream_sync(&bitmap, dev, &self.streams[0][device_idx]).unwrap();
+            or_policy_bitmap.push(_bitmap);
+        }
+        or_policy_bitmap
+    }
```
dkales, line 1615: I do not really get why this has to be a bitmap that is essentially as large as the total DB. Since the IDs->GPU memory mapping is deterministic and consistent, would it not just be more efficient to just push the raw u32 indices to the GPU and just compare them internally?

iris-mpc-gpu/tests/e2e.rs
dkales: but that is my point, no? If variation is -1, then in the branch for 2, we go into the else path which is an expected match, since we flip less than the threshold. I thought what you want here is to flip enough bits in one half so that it no longer is a match?

iris-mpc-gpu/tests/e2e.rs
dkales: but this is not doing the actual generation, this is just the available options for the rng in line 277. I think it is just fine to have this always be an option. The only reason this is as convoluted as it is here, is that we cannot do some of the options unless certain events happened on previous queries (e.g, deletion, insertion)

dkales: Probably this should be refactored to just be a more descriptive enum + random.choice

iris-mpc-gpu/src/server/actor.rs
carlomazzaferro: Unless I am misunderstanding, I don't think that'd be possible because we'd need to pass a mapping of queryIdx -> [dbIdx], since every query will have a different set of dbIdx (and with different lengths, too). 

Though looking directly at the code, I fail to see how we'd use those u32s:

```
extern "C" __global__ void mergeDbResults(...)
{
    size_t idx = blockIdx.x * blockDim.x + threadIdx.x;
    if (idx < numElements)
    {
        for (int i = 0; i < 64; i++)
        {
            unsigned int queryIdx = (idx * 64 + i) / dbLength;
            unsigned int dbIdx = (idx * 64 + i) % dbLength;
            bool matchLeft = (matchResultsLeft[idx] & (1ULL << i));
            bool matchRight = (matchResultsRight[idx] & (1ULL << i));
           ....

```
What we need here is the boolean from each of the comparisons, for each of the queryIdxs

dkales: yeah but is the bool you grab from the bitmap not equivalent to some C-equivalent version of `or_rule_indices[queryIdx].contains(dbIdx)`. I get your point with the different lengths of the `dbIdx` arrays but surely there is some upper limit for this?. I feel that the bitmap is a bit overkilll in terms of memory usage for this, but I guess in the end it is just a few (hunded) MB at most, which should be fine?

philsippl: @dkales this is trading off memory and compute, other options were sorting the list, but we assumed we might get some larger list of indices

philsippl: @dkales yes, you're exactly right, we thought it's worth doing the memory tradeoff here, because it's only up to ~160MB for the whole batch (per device). Other option would would be also binary searching some sorted list of indices.
The bitmap is also not the most efficient yet, we could actually reduce it to at max 160MB total by splitting up the indices modulo device count.

I think the memory allocation should be fine, but let's see the actual performance on the prod machines.

iris-mpc-gpu/tests/e2e.rs
carlomazzaferro: Makes sense, I'll think about refactoring this in a separate PR. It was indeed a bit convoluted

iris-mpc-gpu/tests/e2e.rs
carlomazzaferro: Though at the threshold it should match anyways right? We use =< IIRC, so flipping one side to less than the threshold will leave the other matching  and the one with the flip bits non-matching. 

It could be that my reasoning was mistaken, but the results were correct? I'll refactor to flip one side to threshold + 1 and other side threshold - 1 for clarity

iris-mpc-gpu/tests/e2e.rs
dkales: can also do that in a follow up PR

iris-mpc-gpu/tests/e2e.rs
dkales: An iris code is a match if the fractional distance is less-or-equal than the threshold 0.375 or 3/8. 
When copying an iris code from the DB, this means that the default distance between this query and the position in the DB is 0, so it is a match. To make it a non-match, we need to flip more than 3/8 of the bits to cause it to not match. This threshold_absoute value is exactly 3/8*IRIS_CODE_LEN, so flipping bits up to and including this value will leave the distance less-or-equal to 3/8, which still is a match. Only flipping more bits than this value will result in a non-match. (This also assumes that all of the iris code masks are set to 1, which I think is the case here.

iris-mpc-gpu/Cargo.toml
```diff
@@ -9,7 +9,10 @@ repository.workspace = true
 
 [dependencies]
 bincode = "1.3.3"
-cudarc = { version = "0.12", features = ["cuda-12020", "nccl"] }
+cudarc = { git = "https://github.com/worldcoin/cudarc-fork.git", features = [
```
carlomazzaferro, line 12: Necessary to be able to pass an arbitrary amount of arguments to the kernel function

iris-mpc-gpu/src/dot/kernel.cu
```diff
@@ -93,6 +94,67 @@ extern "C" __global__ void mergeDbResults(unsigned long long *matchResultsLeft,
     }
 }
 
+extern "C" __global__ void mergeDbResultsWithOrPolicyBitmap(unsigned long long *matchResultsLeft, unsigned long long *matchResultsRight, unsigned int *finalResults, size_t queryLength, size_t dbLength, size_t numElements, size_t maxDbLength, unsigned int *matchCounter, unsigned int *allMatches, unsigned int *matchCounterLeft, unsigned int *matchCounterRight, unsigned int *partialResultsLeft, unsigned int *partialResultsRight, const unsigned long long *orPolicyBitmap) // 2D bitmap stored as 1D
+{
+
+    size_t rowStride64 = (maxDbLength + 63) / 64;
```
carlomazzaferro, line 100: Row stride needs to be calculated in the same way as it is in the host: using the `max_db_size`

iris-mpc-gpu/Cargo.toml
philsippl: this has been merged now, you can switch back again: https://github.com/coreylowman/cudarc/pull/321

dkales: sadly I cannot count right and it is actually not enough for the kernel in ps/buckets and there is a follow-up PR :P
https://github.com/coreylowman/cudarc/pull/322

For this one it should be fine to use cudarc >= 0.13.3

carlomazzaferro: Done, thanks!

adukadanny (approved):

adukadanny (approved):

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -844,16 +847,48 @@ impl ServerActor {
...
+
+            // Populate the pre-allocated OR policy bitmap with the serial ids
+            let host_or_policy_bitmap = prepare_or_policy_bitmap(
+                self.max_db_size,
+                batch.or_rule_serial_ids.clone(),
+                batch_size,
+            );
+
+            let device_or_policy_bitmap =
+                self.allocate_or_policy_bitmap(host_or_policy_bitmap.clone());
```
philsippl, line 872: let's maybe add some timing around this, so we now if this is worth improving more? (we had some ideas to improve this)

philsippl (approved):

PR #953, opened 2025-01-17
Add returning error result for skipped requests due to nodes not being synced

danielle-tfh:
### Notes
* follow up from investigation on why nodes self healed

```
Hey Philipp,
We figured out why the queues sync up again and its due to the sync table where we pass the recently deleted signup ids.
Ie: on the next run, the ids that were already read by the desynced queue and communicated to all nodes. Therefore they are ignored by the rest of the nodes when processing. The nodes sync up again.
It is a feature then!
Follow ups:
add error handling to send back errors to signup-service. this should avoid manual intervention.
metric when the hashing fails with an alarm. just a slack message when this occurs (pod restarts should wake us up).
```

### DIFF
--- a/iris-mpc/src/bin/server.rs
+++ b/iris-mpc/src/bin/server.rs
@@ -29,7 +29,8 @@ use iris_mpc_common::{
         },
         smpc_response::{
             create_message_type_attribute_map, IdentityDeletionResult, UniquenessResult,
-            ERROR_FAILED_TO_PROCESS_IRIS_SHARES, SMPC_MESSAGE_TYPE_ATTRIBUTE,
+            ERROR_FAILED_TO_PROCESS_IRIS_SHARES, ERROR_SKIPPED_REQUEST_PREVIOUS_NODE_BATCH,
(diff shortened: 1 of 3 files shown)


iris-mpc/src/bin/server.rs
```diff
@@ -271,6 +271,17 @@ async fn receive_batch(
...
                             );
+                            // shares
+                            send_error_results_to_sns(
+                                smpc_request.signup_id,
+                                &batch_metadata,
+                                sns_client,
+                                config,
+                                error_result_attributes,
+                                UNIQUENESS_MESSAGE_TYPE,
+                                ERROR_FAILED_TO_PROCESS_IRIS_SHARES,
```
eaypek-tfh, line 282: should we maybe have another message for this event? Such as:`ERROR_SKIPPED_REQUEST`

eaypek-tfh (approved):

iris-mpc/src/bin/server.rs
danielle-tfh: Good call 📞

PR #956, opened 2025-01-20
generate masks and templates for left and right

carlomazzaferro:
Addressing the TODO: instead of copying the left batch values into the right batch, we generate them separately. This will be essential for testing a future feature of the protocol, e.g. the ability to pass a custom set of serial ids against which to perform OR-rule (between left and right codes) matching.

See the PR that implements such logic: https://github.com/worldcoin/iris-mpc/pull/951

See the TODO comment: https://github.com/worldcoin/iris-mpc/pull/956/files#diff-0c1888f81c1fbf6d412eaebb8144b3f97d10ae89d279ca4153b4f900003a69bdR363-R371

### DIFF
--- a/iris-mpc-gpu/tests/e2e.rs
+++ b/iris-mpc-gpu/tests/e2e.rs
@@ -28,6 +28,20 @@ mod e2e_test {
     const MAX_DELETIONS_PER_BATCH: usize = 10;
     const THRESHOLD_ABSOLUTE: usize = 4800; // 0.375 * 12800
 
+    #[derive(Clone)]
+    pub struct E2ETemplate {
+        left:  IrisCode,
+        right: IrisCode,
+    }
+
+    #[derive(Clone)]
+    pub struct E2ESharedTemplate {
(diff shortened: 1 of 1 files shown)


iris-mpc-gpu/tests/e2e.rs
```diff
@@ -28,6 +28,20 @@ mod e2e_test {
...
+    #[derive(Clone)]
+    pub struct E2ETemplate {
+        left:  IrisCode,
+        right: IrisCode,
+    }
+
+    #[derive(Clone)]
+    pub struct E2ESharedTemplate {
+        pub left_shared_code:  [GaloisRingIrisCodeShare; 3],
+        pub left_shared_mask:  Vec<GaloisRingTrimmedMaskCodeShare>,
```
dkales, line 40: this could also be [GaloisRingTrimmedMaskCodeShare; 3] to unify and reduce confusion

dkales (approved):

dkales (approved):

PR #957, opened 2025-01-21
Debug/e2e tests fix

carlomazzaferro:
Fixes issues with e2e tests by which actual tests weren't running


dkales (approved):
ah yeah the result was not used

PR #963, opened 2025-01-23
aurel/hawk-startup: More reliable startup and jobs

naure:
Follow-up to #958.

- Introduce `wait_for_session` that tells when the protocols can start using `send / receive`.
- Retry the initial connections to allow other nodes to start.
- Introduce a job queue to interact with `HawkActor` (same as `ServerActorHandle::submit_batch_query`). This simplifies problems of ownership/lifetime.
- Split insertion as "plan" and "apply" phases and return the plans to be persisted.


iris-mpc-cpu/src/hnsw/searcher.rs
```diff
@@ -181,6 +200,49 @@ impl Default for HnswSearcher {
...
+            links.insert(vector_store, q.clone(), nq.clone()).await;
+            links.trim_to_k_nearest(max_links);
+            n_links.push(links);
+        }
+
+        ConnectPlanLayer { neighbors, n_links }
+    }
+
+    /// Two-step variant of `connect_bidir`: execute.
+    async fn connect_apply<V: VectorStore, G: GraphStore<V>>(
```
iliailia, line 230: nit: Would it make sense to leave a to-do to refactor `connect_bidir` using these new functions?

iris-mpc-cpu/src/hnsw/searcher.rs
```diff
@@ -422,6 +484,53 @@ impl HnswSearcher {
         (links, set_ep)
     }
 
+    /// Two-step variant of `insert_from_search_results`: prepare.
```
iliailia, line 487: nit: Would it make sense to leave a to-do to refactor `insert_from_search_results` using these new functions?

iliailia (approved):
Looks good, thank you!

iris-mpc-cpu/src/hnsw/searcher.rs
naure: Done!

iris-mpc-cpu/src/hnsw/searcher.rs
naure: Done!

PR #968, opened 2025-01-27
Improve load time (mpsc channel, post page-lock, directory buckets)

eaypek-tfh:
**Change**

This PR improves load time via:
1. Switching from (returning vector of streams followed by `select_all` and `try_next`) to (spawning a separate tokio task for each `get_object` and sending items to a tokio mpsc channel). Instead of polling multiple dynamically changed list of streams, we now wait for a single channel.
2. Having page locking later than import: Working on memory at the same time with page-lock and import writes is affecting the performance. When we move the page-lock after the import, we saw that page-lock time decreases by 70% while import time remains almost the same
3. Infra change:  Directory buckets (s3express) bring more stable s3 throughput which used to be varying a lot for general purpose buckets. Overall, it improved the performance in stage by around 20-25%.

**Background & More Details**
- We need page-lock for better batch performance and this is why we do not start processing them before page lock finishes. It takes around 7,75s to page-lock 1M records in memory. Parallelizing it does not help as it's probably serialized on OS level.
- We used to start page-lock in the background and immediately start s3 and aurora streams to write into the memory. This used to affect memory write and stream performance indirectly.
- Since we move page-locked after importing everything, the real s3 throughput became obvious where we exceeded aws defined network throughput by around 20-25% in our stage instances.
- In order not to waste the whole time for page-lock, we tried to divide it into chunks and start writing into the memory chunk which is already page-locked. However, chunked page locks resulted in `CUDA_ERROR_INVALID_VALUE` errors. So, we abandoned them.
- With these changes, current bottleneck seems to be memory writing
- Please note that we can further avoid the time to send items to tokio mpsc channel by directly writing into the memory in importer.

### DIFF
--- a/iris-mpc-store/src/s3_importer.rs
+++ b/iris-mpc-store/src/s3_importer.rs
@@ -1,19 +1,9 @@
-use crate::StoredIris;
 use async_trait::async_trait;
 use aws_sdk_s3::{primitives::ByteStream, Client};
 use eyre::eyre;
-use futures::{stream, Stream, StreamExt};
 use iris_mpc_common::{IRIS_CODE_LENGTH, MASK_CODE_LENGTH};
-use std::{
-    mem,
-    pin::Pin,
-    sync::{
-        atomic::{AtomicUsize, Ordering},
-        Arc,
-    },
-    time::Instant,
-};
-use tokio::io::AsyncReadExt;
+use std::{collections::VecDeque, mem, sync::Arc};
+use tokio::{io::AsyncReadExt, sync::mpsc::Sender, task};
 
 const SINGLE_ELEMENT_SIZE: usize = IRIS_CODE_LENGTH * mem::size_of::<u16>() * 2
     + MASK_CODE_LENGTH * mem::size_of::<u16>() * 2
@@ -187,7 +177,7 @@ impl ObjectStore for S3Store {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct LastSnapshotDetails {
     pub timestamp:      i64,
     pub last_serial_id: i64,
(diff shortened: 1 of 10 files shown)


iris-mpc-gpu/src/helpers/mod.rs
```diff
@@ -173,24 +173,35 @@ pub fn htod_on_stream_sync<T: DeviceRepr>(
...
+    chunk_offset: usize,
     code_length: usize,
 ) {
-    let max_size = max_db_length / device_manager.device_count();
+    tracing::info!(
+        "Page-locking chunk: [{}-{}]",
+        chunk_offset,
+        chunk_offset + chunk_length
+    );
+    let size = chunk_length / device_manager.device_count();
```
danielle-tfh, line 185: out of interest - why do we divide by the count?

philsippl: Because we actually don't allocate one blob of memory, but device_count many. Elements are allocated to devices by `serial_id % device_index`.  This is mostly a leftover from when the memory was loaded onto the devices, we could also do it differently now.

iris-mpc-store/src/s3_importer.rs
```diff
@@ -232,85 +222,73 @@ pub async fn last_snapshot_timestamp(
...
-            })
-            .boxed();
 
-    result_stream
+                Ok(())
+            }
+        }));
+    }
+
+    drop(tx);
```
philsippl, line 284: @eaypek-tfh what's the reason for the manual drop?

philsippl (approved):
lgtm!

iris-mpc-store/src/s3_importer.rs
eaypek-tfh: Yeah, I don't need to drop it expicitly as it'd be implicitly dropped after all clones are dropped, right? I think I just wanted to make it obvious but let me remove it 👍

philsippl: Yeah should be all fine, if anything this is a bit awkward since you drop it before the handles are awaited.

PR #970, opened 2025-01-27
aurel/complex-request

naure:
Following #963.

This PR illustrates how to include the complex matching logic (both eyes, rotations, etc). It is a step towards the logic of `ServerActor::process_batch_query`.

- Generalize to left/right eyes database and requests.
- Specify the iris and graph stores in `HawkSession`.
- Match both eyes policy (AND).
- Insert non-matching vectors.


iliailia (approved):
Looks good to me, thanks!

PR #989, opened 2025-01-30
[POP-2073] add buckets as structured response

carlomazzaferro:
This will be parsed by the signup service and will be then stored in a mongodb collection.

The only nuance is how timestamps are set. On initializing the `ServerActor`, we initialize the `BucketStatistics` object with a `DateTime<Utc>` (now) in the `start_timestamp`. When we collect enough matches, we will then log the results and set the `end_timestamp` (now), and pass the filled `BucketStatistics` as a `ServerJobResponse`, with the `end_timestamp` set.


iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1269,19 +1275,10 @@ impl ServerActor {
...
-                results.push_str(&format!(
-                    "    {:.3}-{:.3}: {:?}\n",
-                    previous_threshold, threshold, count
-                ));
-            }
-            tracing::info!("Bucket results:\n{}", results);
+            self.anonymized_bucket_statistics
+                .fill_buckets(&buckets, MATCH_THRESHOLD_RATIO);
+
+            tracing::info!("Bucket results:\n{}", self.anonymized_bucket_statistics);
```
carlomazzaferro, line 1281: Keeping the logging here for now

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1269,19 +1275,10 @@ impl ServerActor {
                 .phase2_buckets
                 .open_buckets(&self.buckets, batch_streams);
 
-            let mut results = String::new();
```
carlomazzaferro, line 1272: Moved this logic to the `BucketStatistics` struct

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -406,6 +407,9 @@ impl ServerActor {
             dev.synchronize().unwrap();
         }
 
+        let anonymized_bucket_statistics =
+            BucketStatistics::new(match_distances_buffer_size, n_buckets, party_id);
```
philsippl, line 411: the `start_timestamp` is then always the start time of the server, is this intended?

philsippl (approved):

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1083,6 +1088,7 @@ impl ServerActor {
                 store_right: query_store_right,
                 deleted_ids: batch.deletion_requests_indices,
                 matched_batch_request_ids,
+                anonymized_bucket_statistics: self.anonymized_bucket_statistics.clone(),
```
dkales, line 1091: since this is never emptied, is this not sent on every response once the buckets have been calculated?
Maybe having `self.anonymized_bucket_statistics` being an option instead and `take`ing it here if full could be simpler?

carlomazzaferro: The inner property `buckets` content is emptied: https://github.com/worldcoin/iris-mpc/pull/989/files#diff-cfedd06c9bac2141ada3a2a2664595d46a77fbb2811d94c2c0a3a15302141cb0R66 

Though I am working in a fix to set the correct start and end timestamp, as currently they are not exactly what we want. I think making it an option won't work, because we'll never be able to set the start_timestamp (unless we store it as ServerActor property)

dkales: yeah but if it is only ever emptied in fill_buckets, then will it not be sent until it is cleared (and then subsequently filled again immediately) in the next call to fill_buckets (which only happens every 2^19 matches atm)?

I think you also need to clear it after a successful send to SNS, unless I am misunderstanding something?

carlomazzaferro: you definitely have a point there

iris-mpc/src/bin/server.rs
```diff
@@ -1456,6 +1458,36 @@ async fn server_main(config: Config) -> eyre::Result<()> {
...
+                            .wrap_err("failed to serialize anonymized statistics result")
+                    })
+                    .collect::<eyre::Result<Vec<_>>>()?;
+
+                send_results_to_sns(
+                    anonymized_statistics_results,
+                    &metadata,
+                    &sns_client_bg,
+                    &config_bg,
+                    &identity_deletion_result_attributes,
```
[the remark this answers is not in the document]
carlomazzaferro: Ops, missed a changeset in the commit. Thanks for spotting it

Issue #985, opened 2025-01-29
Prepare server binary for GPU feature gating


dkales:
- [x] remove dependance on GPU networking in server_main for DB sync (#990)
  - this should be moved to CPU-networking (probably GRPC)
- [x] abstract away loading of Iris shares into memory representation behind traits (#1014)
  - currently, the shares are loaded into specially prepared host memory which is intended for later DMA operations
  - this is more expensive in many ways and also kind of prohibits us to use nice rust types for this, working with raw pointers, which is not what we want for the CPU version
- [ ] abstract actor initialization away such that there is essentially only a few places that would have to be changed
  - initialization of the IrisMatchingActor implementation
  - loading of the database (using a `HNSWLoader` instead of the `GPULoader`)
  - maybe storing back to persistent storage, or should this be internal to the actor in the end?
    - could also use a `Store` trait that is analogous to the `Loader` trait (or just be one combined one)

dkales:
rough plan from the meeting @philsippl @naure @carlomazzaferro @eaypek-tfh @wojciechsromek @bgillesp 

Will try to prepare PR(s) addressing these, happy for more input

carlomazzaferro:
Just to be sure about how this would work at a slightly higher level:

- We'd have a single `server.rs` implementation, which can be run *either* with the GPU features, or without. 
- The same docker image can be used, but we could also use an image without any of the gpu-related stuff
- We can deploy each of the implementations separately, consuming from completely different queues and producing results to also different queues altogether

I am asking mostly to convince myself that we don't necessarily need a completely separate `server_gpu.rs` and `server_cpu.rs` implementation, which at first seems to be an easier approach

dkales:
I think the above steps would be required for both approaches, otherwise there would be a lot of code duplication.

I think the consensus in the end was to have 2 binaries, but move a lot of the common code out of them.

philsippl:
Also not server related, but e2e tests are currently in the gpu crate, they should also be agnositic (are already?)

dkales:
I think the e2e tests for the actor could share some common core (the test case generator), the setup of the actors will probably stay specific to the individual actor impl

dkales:
With #990, #1014, #1020 this seems now relatively done, seemed like a good base based on our meeting today.

PR #990, opened 2025-01-30
Move initial DB sync between nodes to HTTP instead of NCCL

dkales:
This is the first step in addressing #985, making the DB sync call independent of the underlying networking implementation.
The simplest strategy for me here was to just re-use the exsisting http server infrastructure also used for the health and ready checks.


iris-mpc/src/bin/server.rs
```diff
@@ -1012,6 +1016,66 @@ async fn server_main(config: Config) -> eyre::Result<()> {
...
+        let res = reqwest::get(format!("http://{}:3000/sync", host)).await;
+        match res {
+            Ok(res) => {
+                let state: SyncState = match res.json().await {
+                    Ok(state) => state,
+                    Err(e) => {
+                        tracing::error!("Failed to parse sync state from party {}: {:?}", host, e);
+                        panic!(
+                            "could not get sync state from party {}, trying to restart",
+                            host
```
dkales, line 1039: what should be the behavior on this? Since we already did a heartbeat once, this should be expected to always work, so restarting should be fine? Do we need some graceful shutdown already at this point?

iris-mpc/src/bin/server.rs
```diff
@@ -1012,6 +1016,66 @@ async fn server_main(config: Config) -> eyre::Result<()> {
...
+                        );
+                    }
+                };
+                states.push(state);
+            }
+            Err(e) => {
+                tracing::error!("Failed to fetch sync state from party {}: {:?}", host, e);
+                panic!(
+                    "could not get sync state from party {}, trying to restart",
+                    host
```
dkales, line 1049: what should be the behavior on this? Since we already did a heartbeat once, this should be expected to always work, so restarting should be fine? Do we need some graceful shutdown already at this point?

iris-mpc/src/bin/server.rs
```diff
@@ -1012,6 +1016,66 @@ async fn server_main(config: Config) -> eyre::Result<()> {
...
+            db_len
+        );
+        store.rollback(db_len).await?;
+        metrics::counter!("db.sync.rollback").increment(1);
+    }
+
+    if download_shutdown_handler.is_shutting_down() {
+        tracing::warn!("Shutting down has been triggered");
+        return Ok(());
+    }
```
dkales, line 1077: is this connected to the rollback or the db loading?
i.e., should this be here or left with the DB loading procedure

dkales:
Also these changes are a little bit harder to test locally, so would probably need a staging deployment?

iris-mpc/src/bin/server.rs
philsippl: Since this is before receiving any batch, should be find to just panic (cc @danielle-tfh @eaypek-tfh)

iris-mpc/src/bin/server.rs
philsippl: same as above?

iris-mpc/src/bin/server.rs
philsippl: should be fine to leave it here with pg consistency?

philsippl:
changes look good to me, @eaypek-tfh @danielle-tfh could one of you test this on staging if you have time?

eaypek-tfh:
**Stage test results:**
1. Deleted an entry from party1 and restarted the servers. They synced back correctly
    image

2. Deleted three entries from party2 and sent a uniqueness request. Strangely, party2 skipped the serial id. I'm not sure if this issue was there before. At this point, restarting the parties end up with "Detected inconsistency between max serial id" error and it does not even go into sync state. So, it needs manual intervention.
    image

**My comments:**
1. If one of the parties die, they never get back in sync. Because when partyX is down, the others start graceful shutdown after some time. Then, partyX sees the others shutting down and it manually triggers shut down too. We end up with "Failed to fetch sync state from party". Current main could recover from this after 1-2 restarts but this branch did not recover. Tbh, it becomes a bit complex with lots of different k8s probes, heartbeats, initial shutdown handlers, etc.
2. I'd love to see the deleted tests in [sync_nccl.rs file](https://github.com/worldcoin/iris-mpc/pull/990/files#diff-51df368df465a5060a6abe708d961e4909250088a223db4588e5b89e3a928c9f) to be restored
3. We need to refetch `store_len` after a rollback because otherwise it will fail on data loading thinking not all serial ids were loaded - needless restart. This is a problem in current impl too.

iris-mpc/src/bin/server.rs
eaypek-tfh: yes, i think graceful shutdown shouldn't be needed here but shall we consider some sort of retries?

naure:
LGTM!

dkales:
> **Stage test results:**
> 
>     1. Deleted an entry from party1 and restarted the servers. They synced back correctly
>        image
> 
>     2. Deleted three entries from party2 and sent a uniqueness request. Strangely, party2 skipped the serial id. I'm not sure if this issue was there before. At this point, restarting the parties end up with "Detected inconsistency between max serial id" error and it does not even go into sync state. So, it needs manual intervention.
>        image
> 

Thanks for the tests. I will look more into the issues.

> 
> **My comments:**
> 
>     1. If one of the parties die, they never get back in sync. Because when partyX is down, the others start graceful shutdown after some time. Then, partyX sees the others shutting down and it manually triggers shut down too. We end up with "Failed to fetch sync state from party". Current main could recover from this after 1-2 restarts but this branch did not recover. Tbh, it becomes a bit complex with lots of different k8s probes, heartbeats, initial shutdown handlers, etc.

I think the current startup sequence could use a "starting" or "syncing" state in the ready check to synchronize on that.
Atm I think there is a bit of a break in functionality since the previous check only worked once and then never again and also synced on the 3 parties broadcasting at the same time due to using NCCL. Another option would be to disable the sync endpoint after the sync is done and return 404 instead here, handling as appropriate in the startup sequence.

> 
>     2. I'd love to see the deleted tests in [sync_nccl.rs file](https://github.com/worldcoin/iris-mpc/pull/990/files#diff-51df368df465a5060a6abe708d961e4909250088a223db4588e5b89e3a928c9f) to be restored

I can certainly bring them back, although some of them related to the serialization size are no longer needed due to using http instead for transport,

> 
>     3. We need to refetch `store_len` after a rollback because otherwise it will fail on data loading thinking not all serial ids were loaded - needless restart. This is a problem in current impl too.

I see, makes sense, will do that.

dkales:
> I think the current startup sequence could use a "starting" or "syncing" state in the ready check to synchronize on that. Atm I think there is a bit of a break in functionality since the previous check only worked once and then never again and also synced on the 3 parties broadcasting at the same time due to using NCCL. Another option would be to disable the sync endpoint after the sync is done and return 404 instead here, handling as appropriate in the startup sequence.

One option would be to use the existing ready flag for that:
* During startup, even before doing the health-check loop, we wait for other systems to come up and wait until ready=false (503) which indicates that the service is restarting. SInce our server is in startup mode, we force other servers to restart since our health-check UUID changed and then we wait for them to also reach this state?

dkales:
@eaypek-tfh do you think the last commit would solve the first issue?

eaypek-tfh:
> @eaypek-tfh do you think the last commit would solve the first issue?

just deployed the latest image in stage (sorry for polluting your branch). after one node dies, all the others restart and get back healthy. so, it solved the first issue 👍

iris-mpc/src/bin/server.rs
danielle-tfh: @dkales what are your thoughts on moving the sync as part of the health check request? We already sync the UUID if i am not mistaken.

This way we have less http endpoints to keep track off. And one could maybe argue that the health includes the current state?

danielle-tfh:
> @eaypek-tfh do you think the last commit would solve the first issue?



> Member

what do you think about adding it to the heartbeat response? we would know at least that all the nodes started and have the info if its synced between them?

iris-mpc/src/bin/server.rs
dkales: hmm the synced state is actually not the current db len, but the one at startup. It can also change if we do a roll-back/forward. Since this information is a one-time exchange during startup I would rather keep it separate but I am open to do it in this way if you think it is best. In any case the route should probably be called `startup-sync` or something similar

dkales:
I think with the last 2 commits, this should address the issues @eaypek-tfh brought up:
* refetching of the state after potential sync
* better rendezvous procedure of nodes during startup

I have not added the deleted tests back since they tested the NCCL specific functionality that are no longer used.
Generic sync state tests are still present in  https://github.com/worldcoin/iris-mpc/blob/8870e835fd675b34a1d3f46d60ce4b040f4a637c/iris-mpc-common/src/helpers/sync.rs#L46-L47

Open questions:
* Changes to github workflow files should be reverted?
* Staging deployment for remaining changes would be great
* @danielle-tfh commented about merging sync and health routes, should we do this?

danielle-tfh:
> I think with the last 2 commits, this should address the issues @eaypek-tfh brought up:
> 
> * refetching of the state after potential sync
> * better rendezvous procedure of nodes during startup
> 
> I have not added the deleted tests back since they tested the NCCL specific functionality that are no longer used. Generic sync state tests are still present in
> 
> https://github.com/worldcoin/iris-mpc/blob/8870e835fd675b34a1d3f46d60ce4b040f4a637c/iris-mpc-common/src/helpers/sync.rs#L46-L47
> 
> Open questions:
> 
> * Changes to github workflow files should be reverted?
> * Staging deployment for remaining changes would be great
> * @danielle-tfh commented about merging sync and health routes, should we do this?
> * I changed some of the "ANCHOR" logs, are they load-bearing in some way 😅 ?

Thanks @dkales !

I am a fan of merging them - adding a separate sync does a lot of what the initial health check does.

It would also simplify the syncing and the various route logic.

eaypek-tfh (approved):

dkales:
@danielle-tfh I would aim to merge this now since @eaypek-tfh tested it in this configuration and I would need to think about it a bit more if merging the two routes is ok since one is a one-time thing and the heartbeat is continuous and tracks the state. I will revisit this later on if that is ok.

danielle-tfh:
> @danielle-tfh I would aim to merge this now since @eaypek-tfh tested it in this configuration and I would need to think about it a bit more if merging the two routes is ok since one is a one-time thing and the heartbeat is continuous and tracks the state. I will revisit this later on if that is ok.

sounds reasonable! I can also take a look at changing this once your other PRs are merged.

PR #991, opened 2025-01-31
Add support for counting operations declared in spans

bgillesp:
Previously the `OpCountersLayer` type only supported counting of operations declared at `tracing` event callsites.  This PR extends this functionality to also handle operations declared at span callsites, where counting is handled by a call to `on_new_span`.

The PR also fixes an issue where the `OpCountersLayer` did not gracefully handle event callsites which did not specify an `event_type` field.  Now, instead of panicking, such callsites are silently ignored.


iliailia (approved):
This PR looks good, thanks!

One remark though is that we might be interested in the number of operations happening while a span is active. For example, the number of `less_than` operations happened within `search_sorted` or `search_layer`.  It looks that this change doesn't provide this functionality, right?

bgillesp:
> One remark though is that we might be interested in the number of operations happening while a span is active. For example, the number of `less_than` operations happened within `search_sorted` or `search_layer`. It looks that this change doesn't provide this functionality, right?

Yes, that's correct -- this implementation just allows `new_span` to trigger incrementing a counter directly. 
 For support of counting within spans, do you imagine this to be "log counts of operations per span of interest" or "filter operations according to inclusion within one or more spans of interest"?  (Or something else?)

iliailia:
> > One remark though is that we might be interested in the number of operations happening while a span is active. For example, the number of `less_than` operations happened within `search_sorted` or `search_layer`. It looks that this change doesn't provide this functionality, right?
> 
> Yes, that's correct -- this implementation just allows `new_span` to trigger incrementing a counter directly. For support of counting within spans, do you imagine this to be "log counts of operations per span of interest" or "filter operations according to inclusion within one or more spans of interest"? (Or something else?)

The former.

bgillesp:
Based on our conversation earlier @iliailia, merging this PR as is, with the plan to implement span-organized counting for logical operations and data transfer amounts using a custom formatter for `tracing-forest` functionality, possibly in #1007 or a follow-up PR.

PR #992, opened 2025-01-31
accept reauth requests

eaypek-tfh:
**Main Change**
- In this PR, we introduce a new primitive: `Reauth` (along with `Enrollment` and `Deletion`)
- We see `Reauth` as part of the batch because we'll need all the matches for these requests to decide if they're successful or not.
- In a following PR, we'll process the `Reauth` requests in actor and mark them as success only if they match with given index

**Refactors**
- Moving `get_iris_data_by_party_id`, `decrypt_iris_share`, and `validate_iris_share` methods out of `UniquenessRequest` member functions so that they can be used by both Reauth and Uniqueness requests.
- Adding `request_types` entry to the `BatchQuery` to distinguish various request types keeping in mind that it can expand in the future.
- Moving the logic to download & parse iris shares future from uniqueness_request's body to a common function. Also, adding `GaloisShares` and `ParseSharesTaskResult` type aliases for better readability.

**Notes**
- Please note that `use_or_rule` will be consumed after the LUC PR is merged. So, that's why it's unused for now.

### DIFF
--- a/iris-mpc-common/src/helpers/smpc_request.rs
+++ b/iris-mpc-common/src/helpers/smpc_request.rs
@@ -105,6 +105,7 @@ pub const IDENTITY_DELETION_MESSAGE_TYPE: &str = "identity_deletion";
 pub const ANONYMIZED_STATISTICS_MESSAGE_TYPE: &str = "anonymized_statistics";
 pub const CIRCUIT_BREAKER_MESSAGE_TYPE: &str = "circuit_breaker";
 pub const UNIQUENESS_MESSAGE_TYPE: &str = "uniqueness";
+pub const REAUTH_MESSAGE_TYPE: &str = "reauth";
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct UniquenessRequest {
@@ -122,6 +123,16 @@ pub struct CircuitBreakerRequest {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct IdentityDeletionRequest {
     pub serial_id: u32,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct ReAuthRequest {
+    pub reauth_id:               String,
+    pub batch_size:              Option<usize>,
+    pub s3_key:                  String,
+    pub iris_shares_file_hashes: [String; 3],
+    pub serial_id:               u32,
+    pub use_or_rule:             bool,
 }
 
 #[derive(Error, Debug)]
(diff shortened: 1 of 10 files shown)


iris-mpc/src/bin/server.rs
```diff
@@ -299,78 +305,97 @@ async fn receive_batch(
...
+                        batch_query
+                            .request_ids
+                            .push(reauth_request.reauth_id.clone());
+                        batch_query
+                            .request_types
+                            .push(REAUTH_MESSAGE_TYPE.to_string());
+                        batch_query.metadata.push(batch_metadata);
+                        batch_query.reauth_target_indices.insert(
+                            reauth_request.reauth_id.clone(),
+                            reauth_request.serial_id - 1,
```
carlomazzaferro, line 378: this is because the db indices are 0-index, correct?

carlomazzaferro (approved):
clean!

iris-mpc/src/bin/server.rs
eaypek-tfh: yes, definitely. in actor.rs, the world is all 0-indexed - insertion indexes, matched indexes and so on. 

while we send requests, we apply -1 to each serial id. 

after we get responses from it, we apply +1 to each returned index to bring it back to 1-indexed serial id world so that contract is consistent. [Example](https://github.com/worldcoin/iris-mpc/blob/main/iris-mpc/src/bin/server.rs#L1355)

iris-mpc-gpu/src/server/mod.rs
```diff
@@ -66,21 +66,29 @@ pub struct BatchMetadata {
...
+    pub query_right:              BatchQueryEntries,
+    pub db_right:                 BatchQueryEntries,
+    pub store_right:              BatchQueryEntries,
+    pub query_right_preprocessed: BatchQueryEntriesPreprocessed,
+    pub db_right_preprocessed:    BatchQueryEntriesPreprocessed,
+    pub valid_entries:            Vec<bool>,
+
+    // Only reauth specific fields
+    // Map from reauth request id to the index of the target entry to be matched
+    pub reauth_target_indices: BTreeMap<String, u32>,
```
philsippl, line 87: any specific reason here instead of just HashMap?

philsippl (approved):

iris-mpc-gpu/src/server/mod.rs
eaypek-tfh: yes. `Hash` trait [here](https://github.com/worldcoin/iris-mpc/pull/992/files#diff-9b0f591ea4930013253441d5dc05178ec720e20e6d7728feb5f5ce43a9b5fbdcR67) is not implemented for `HashMap` and I avoided to implement myself

eaypek-tfh: completely removed `Hash` trait from `BatchQuery` struct and updated the field to `HashMap` as well. Thank you! 🙌 

relevant commit: https://github.com/worldcoin/iris-mpc/pull/992/commits/0637d34a9d0d6d5769aa8c507f704e4a2338da7e

iris-mpc-common/src/helpers/smpc_request.rs
```diff
@@ -124,6 +125,16 @@ pub struct IdentityDeletionRequest {
     pub serial_id: u32,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct ReAuthRequest {
+    pub reauth_id:               String,
+    pub batch_size:              Option<usize>,
+    pub s3_key:                  String,
+    pub iris_shares_file_hashes: [String; 3],
```
danielle-tfh, line 133: So recently, we actually added the hashes to the S3 https://github.com/worldcoin/signup-service/commit/35b8514448be9af1c73ce2653a74362a627cb55b. This opens up the possibility of retries and keeps any information about the signup in one place.

I would like to move the uniqueness request to using this as well.

How about for re-auth we already use the one from S3? will this change much?

eaypek-tfh: Yeah, good call. But i'd do it in a separate PR as it's not really related to just reauth 👍 

Thank you!

iris-mpc/src/bin/server.rs
```diff
@@ -417,13 +447,18 @@ async fn receive_batch(
                 tracing::error!("Failed to process iris shares: {:?}", e);
                 // Return error message back to the signup-service if failed to process iris
                 // shares
+                let result_attributes = match batch_query.request_types[index].as_str() {
```
danielle-tfh, line 450: Do we use the same results SNS for re-auth and uniqueness requests?

eaypek-tfh: yes, we use the same SNS for deletion, uniqueness, reauth, and anonymized bucket stats. they're then forwarded into different SQSes based on the msg attribute.

PR #997, opened 2025-02-01
handle connection reset during s3 load

eaypek-tfh:
**Change**
- Introducing exponential backoff based retry on s3 import to mitigate below error:

> Panic occurred with message: Couldn't fetch and parse chunks from s3: streaming error
Caused by:
   0: error reading a body from connection: Connection reset by peer (os error 104)
   1: Connection reset by peer (os error 104)
Location:
    /src/gpu-iris-mpc/iris-mpc-store/src/s3_importer.rs:275:48

### DIFF
--- a/iris-mpc-store/src/s3_importer.rs
+++ b/iris-mpc-store/src/s3_importer.rs
@@ -321,7 +321,6 @@ async fn read_range_in_chunk(
     range_size: usize,
     tx: Sender<S3StoredIris>,
 ) -> eyre::Result<()> {
-    // Open the range for the chunk
     let mut stream = store
         .get_object(
             key,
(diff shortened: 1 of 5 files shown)


philsippl (approved):

carlomazzaferro (approved):

iris-mpc-store/src/s3_importer.rs
```diff
@@ -227,6 +227,8 @@ pub async fn fetch_and_parse_chunks(
     prefix_name: String,
     last_snapshot_details: LastSnapshotDetails,
     tx: Sender<S3StoredIris>,
+    max_retries: usize,
```
danielle-tfh, line 230: Any reason to not use whats built in with the AWS SDK?

eaypek-tfh: Oh yes, I have a very good reason :slightly_smiling_face:

We call the SDK via get_object for each 200 items (15MB) and it returns us a stream. We then consume this stream to parse all these items. If the stream is later dropped on s3 side, we get this error (we see connection closed error). So, fine tuning our existing max 5 retry config on the sdk won't help in this case. we need to retry for the whole range.

PR #999, opened 2025-02-02
Fix non-deterministic buckets

philsippl:
The e2e tests were failing seemingly random since including a check for total counts of buckets.
The reason for this was another source of non-determinism across the three nodes in the kernel that saves distances if they are a match. It can happen that the threads on the different nodes / devices arrive at this maximum in different order, such that we end up saving different distances (the sorting can't not solving this).
https://github.com/worldcoin/iris-mpc/blob/1dd0bc686de22ae93a0ffa043fac02f2083732eb/iris-mpc-gpu/src/dot/kernel.cu#L74

It's hard to solve this with the current approach bc the actual number of matches is unpredictable. This fix simply tries to capture a few more distances than needed, do the sorting, and then truncate to the desired length, such that this should very likely contain the same elements.

### DIFF
--- a/iris-mpc-common/src/config/mod.rs
+++ b/iris-mpc-common/src/config/mod.rs
@@ -118,6 +118,9 @@ pub struct Config {
     #[serde(default = "default_match_distances_buffer_size")]
     pub match_distances_buffer_size: usize,
 
+    #[serde(default = "default_match_distances_buffer_size_extra_percent")]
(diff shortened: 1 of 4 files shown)


carlomazzaferro (approved):

dkales (approved):
makes sense, 10% should be enough to cover most cases I think

PR #1002, opened 2025-02-03
aurel/graph_store: copy and adapt GraphPg

naure:


iris-mpc-cpu/src/hnsw/graph/graph_store.rs
```diff
@@ -0,0 +1,399 @@
...
+            let mut d = vec![];
+            for v in vectors.iter() {
+                d.push(vector_store.eval_distance(&vectors[0], v).await);
+            }
+            d
+        };
+
+        let ep = graph.get_entry_point().await;
+
+        let ep2 = EntryPoint {
```
iliailia, line 324: Shouldn't you compare the content of `ep` with `ep2` as well?

iliailia (approved):
Looks great, thanks!

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
naure: `ep` should be None so I added an `assert`.

PR #1004, opened 2025-02-04
feat: fix check_bucket_statistics and disable faulty WithOrRuleSet testcases in e2e

rw0x0:
The e2e testcase should work now again, however it shows that something is wrong with the OR rule implementation

### DIFF
--- a/iris-mpc-gpu/src/dot/distance_comparator.rs
+++ b/iris-mpc-gpu/src/dot/distance_comparator.rs
@@ -287,6 +287,7 @@ impl DistanceComparator {
             );
 
             self.device_manager.device(i).bind_to_thread().unwrap();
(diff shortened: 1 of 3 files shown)


philsippl (approved):

iris-mpc-gpu/src/dot/distance_comparator.rs
```diff
@@ -298,7 +299,7 @@ impl DistanceComparator {
                             &matches_bitmap_left[i],
                             &matches_bitmap_right[i],
                             &final_results[i],
-                            self.query_length,
+                            (self.query_length / ROTATIONS) as u64,
```
carlomazzaferro, line 302: Missed this, even if used only for boundary checks

PR #1005, opened 2025-02-05
Process reauth messages

eaypek-tfh:
**Change**
- Process the reauth requests in actor. 
- Only consider the AND rule for now. 
  - If there's exactly one match to the target serial id -> consider it as successful reauth. write into memory and db
  - Else -> consider is as failed reauth. do not persist anything and send result to SNS.

**Notes**
- OR rule will be handled in a follow-up PR
- Sync mechanism will be handled in a follow-up PR
- `ServerJobResult` could probably be refactored to separate reauth and uniqueness results but i didn't want to spend time on it for now.
- There'll be an e2e PR as well targeting this branch. I'll merge e2e into this branch and then branch to main.

### DIFF
--- a/iris-mpc-gpu/src/server/actor.rs
+++ b/iris-mpc-gpu/src/server/actor.rs
@@ -1095,24 +1095,26 @@ impl ServerActor {
             })
             .collect::<Vec<_>>();
 
-        let successful_reauth_positions = match_ids_filtered
+        let successful_reauths = match_ids_filtered
             .iter()
             .enumerate()
-            .filter(|(idx, matches)| {
-                let reauth_id = batch.request_ids[*idx].clone();
+            .map(|(idx, matches)| {
+                let reauth_id = batch.request_ids[idx].clone();
                 // expect exactly one match to the target reauth index
-                batch.request_types[*idx] == REAUTH_MESSAGE_TYPE
+                batch.request_types[idx] == REAUTH_MESSAGE_TYPE
                     && matches.len() == 1
                     && matches[0] == *batch.reauth_target_indices.get(&reauth_id).unwrap()
             })
-            .map(|(idx, _)| idx)
-            .collect::<Vec<_>>();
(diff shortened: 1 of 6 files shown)


philsippl (approved):

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1088,9 +1095,34 @@ impl ServerActor {
...
+                    && matches[0] == *batch.reauth_target_indices.get(&reauth_id).unwrap()
+            })
+            .collect::<Vec<bool>>();
+        let mut reauth_updates_per_device = vec![vec![]; self.device_manager.device_count()];
+        for (reauth_pos, success) in successful_reauths.clone().iter().enumerate() {
+            if !*success {
+                continue;
+            }
+            let reauth_id = batch.request_ids[reauth_pos].clone();
+            let reauth_index = *batch.reauth_target_indices.get(&reauth_id).unwrap();
```
carlomazzaferro, line 1115: out of curiosity / nit, is it safe to call unwrap on this?

carlomazzaferro (approved):
just a comment to be sure we're safe

iris-mpc-gpu/src/server/actor.rs
dkales: should be due to it being built like this here: https://github.com/worldcoin/iris-mpc/blob/daa7e76cecde589fd9a7f54100174251bb6fcf0c/iris-mpc/src/bin/server.rs#L394

eaypek-tfh: yeap, we discussed this async as well. let me write here too:

We populate the batch.reauth_target_indices [here](https://github.com/worldcoin/iris-mpc/blob/main/iris-mpc/src/bin/server.rs#L394-L397) ourselves. So, if it can not be found for a given reauth request, then there should be smt seriously wrong and IMO, crushing would be the best option.

PR #1008, opened 2025-02-05
[WIP] Implement reauth e2e tests

eaypek-tfh:


carlomazzaferro (approved):

iris-mpc-gpu/tests/e2e.rs
```diff
@@ -290,18 +293,26 @@ mod e2e_test {
...
 
                 for (
-                    (((((req_id, &was_match), &idx), partial_left), partial_right), match_id),
+                    (
+                        (
+                            ((((req_id, &was_match), &was_reauth_success), &idx), partial_left),
+                            partial_right,
+                        ),
+                        match_id,
+                    ),
```
dkales, line 310: we should probably be using `itertools::izip` for this, the large amount of tuple nesting is adding so much noise here.

iris-mpc-gpu/tests/e2e.rs
```diff
@@ -571,7 +606,7 @@ mod e2e_test {
         /// The rng that is used internally
         rng:                    StdRng,
 
-        // info for current batch, will be cleared at the start of a new batch
+        /// info for current batch, will be cleared at the start of a new batch
```
dkales, line 609: that was actually intended to be a non-doc comment, being a deliminator between persistent and per-batch storage

dkales:
I would also probably add the case of testing a previous re-auth message:
* check that a previously re-authed iris code no longer matches
* check that a newly re-authed iris-code matches

eaypek-tfh:
> * check that a previously re-authed iris code no longer matches
>   
>   * also what is the intended behavior on this? Since it is overwritten it should no longer match at least in E2E tests.
>   * however, in practice it would be usually close enough to probably match?

The idea behind reauth is to improve iris code quality for users. So, the new codes will be very similar to the older codes but just with an improved quality. So, previously re-authed code should match again. I don't think this would make a good test case, so, I'll skip it. But lmk if you think othwerwise 👍 

> * check that a newly re-authed iris-code matches

That's a good idea. I'll already extend the test cases for `OR` rule as well. Let me do it with that follow-up. I think your suggestion is basically this flow: `Successful Reauth -> Enrollment (with new iris codes from reauth's) -> expect match`.

PR #1014, opened 2025-02-06
Encapsulate GPU related functionality for DB loading behind trait

dkales:
Building upon #990, this moves most of the GPU related functionality behind a trait describing operations on the InMemoryStore.
Still some of the methods on that trait are bound to the current layout of the GPU inmemory representation (e.g., loading in 2 limbs instead of whole u16).

The loading functionality in server.rs then uses this trait.

TODOs:
* `config.fake_db_size` branch seems broken to me already in current main, since the memory is never registered at all? Just remove it?
* Only really GPU relevant imports are now the Actor, Handle, and BatchQuery structs. These still contain the query preprocessing that is splitting it in two limbs, which is not needed for the CPU variant. Should we keep this for now or move this preprocessing into the actor or a trait?


dkales:
@naure have a look at the `InMemoryStore` trait that encapsulates the functionality that is GPU specific, I guess this interface would be somewhat easy for your HSNW DB to implement right? (Mapping back the limb representation that is saved to S3 to full u16 is annoying but not that complicated). Still need to think on how to fit the Graph loading in this trait as well?

iris-mpc/src/bin/server.rs
```diff
@@ -1910,3 +1806,158 @@ async fn process_identity_deletions(
 
     Ok(())
 }
+
+#[allow(clippy::too_many_arguments)]
```
danielle-tfh, line 1810: So nice to see this in a method! thank u!

carlomazzaferro (dismissed):

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -150,6 +148,7 @@ impl ServerActor {
         disable_persistence: bool,
         enable_debug_timing: bool,
     ) -> eyre::Result<(Self, ServerActorHandle)> {
+        tracing::info!("GPU Actor: Starting Device Manager");
         let device_manager = Arc::new(DeviceManager::init());
```
danielle-tfh, line 152: Do we still need to keep this as an arc? or is there only one copy now.

iris-mpc-gpu/src/dot/share_db.rs
```diff
@@ -304,6 +304,7 @@ impl ShareDB {
                     CU_MEMHOSTALLOC_PORTABLE,
                 );
             }
+            self.device_manager.track_locked_db(db.code_gr.clone());
```
danielle-tfh, line 307: `register_host_memory` thoughts on moving this to deviceManager for consistency with this and drop?

iris-mpc-gpu/src/server/actor.rs
dkales: That is true, will change that

iris-mpc-gpu/src/dot/share_db.rs
dkales: makes sense

carlomazzaferro (approved):

PR #1016, opened 2025-02-10
Implement or rule on reauth requests

eaypek-tfh:
**Main Change**
- Change `ReauthResult` to include information on or rule usage and result.
- As agreed with AI team, finalize reauth as success if any side of the target serial id is matching. Do not consider AND matches with the other serial ids but just report those back to SNS.
- Write reauth result as 1-based serial id insead of 0-based index to SNS.
- Add e2e tests for reauth or rule matching and non-matching.
  - Refactor existing or rule e2e tests by extracting `prepare_flipped_codes` function.

**Side Changes**
- Fix returning correct result message on reauth share parsing errors -> it used to be uniqueness result.
- Create `new_error_results` method on `Reauth` and `Uniqueness` results for easier construction.

### DIFF
--- a/iris-mpc-gpu/tests/e2e.rs
+++ b/iris-mpc-gpu/tests/e2e.rs
@@ -1068,7 +1068,7 @@ mod e2e_test {
 
         /// Returns a template with flipped bits of given `db_index`.
         ///
-        /// If `will_match` is true, the template will be flipped to above
+        /// If `will_match` is false, the template will be flipped to above
         /// threshold on both sides If `flip_right` is true, the right
         /// side will be flipped to above threshold
         fn prepare_flipped_codes(
--- a/iris-mpc-gpu/tests/e2e.rs
+++ b/iris-mpc-gpu/tests/e2e.rs
@@ -326,9 +326,6 @@ mod e2e_test {
 
                     if !test_case_generator.or_rule_matches.contains(req_id) {
                         assert_eq!(partial_left, partial_right);
-                    }
-
-                    if !test_case_generator.or_rule_matches.contains(req_id) {
                         assert_eq!(partial_left, match_id);
                     }
                     test_case_generator.check_result(
@@ -375,6 +372,9 @@ mod e2e_test {
         match maybe_reauth_target_index {
             Some(target_index) => {
                 batch.request_types.push(REAUTH_MESSAGE_TYPE.to_string());
+                batch
+                    .reauth_use_or_rule
+                    .insert(request_id.clone(), !or_rule_serial_ids.is_empty());
                 batch
                     .reauth_target_indices
                     .insert(request_id.clone(), *target_index);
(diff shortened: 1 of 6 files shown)


iris-mpc-gpu/tests/e2e.rs
```diff
@@ -1015,28 +998,118 @@ mod e2e_test {
...
+                        template
                     }
                 }
             };
             (request_id, e2e_template, or_rule_indices)
         }
 
+        /// Returns a template with flipped bits of given `db_index`.
+        ///
+        /// If `will_match` is true, the template will be flipped to above
```
carlomazzaferro, line 1071:
```suggestion
        /// If `will_match` is false, the template will be flipped to above
```

carlomazzaferro: Both above threshold == neither matches

iris-mpc-gpu/tests/e2e.rs
```diff
@@ -1015,28 +998,118 @@ mod e2e_test {
...
+                        self.or_rule_matches.push(request_id.to_string());
+                        self.reauth_target_indices
+                            .insert(request_id.to_string(), db_index as u32);
+                        or_rule_indices = vec![db_index as u32];
+                        let will_match = true;
+                        let flip_right = Some(self.rng.gen());
+                        let template = self.prepare_flipped_codes(db_index, will_match, flip_right);
+                        self.expected_results
+                            .insert(request_id.to_string(), ExpectedResult {
+                                db_index:             None,
```
carlomazzaferro, line 1039: If we expect this to match, shouldn't the `db_index` be set to the `db_index` from which we get the template?

iris-mpc-gpu/tests/e2e.rs
```diff
@@ -1015,28 +998,118 @@ mod e2e_test {
...
+        /// Returns a template with flipped bits of given `db_index`.
+        ///
+        /// If `will_match` is false, the template will be flipped to above
+        /// threshold on both sides If `flip_right` is true, the right
+        /// side will be flipped to above threshold
+        fn prepare_flipped_codes(
+            &mut self,
+            db_index: usize,
+            will_match: bool,
+            flip_right: Option<bool>,
```
carlomazzaferro, line 1078: Thoughts on making `flip_right` a `bool` and pushing the responsibility of making it `rng.gen()` to the caller? Makes things a bit more explicit

carlomazzaferro (approved):
Minor comments, the test one is possibly the only high-importance one, otherwise lovely PR

iris-mpc-gpu/tests/e2e.rs
eaypek-tfh: The reason I made `flip_right` optional was because it was only considered if `will_match` is `true`. So, for cases with `will_match=false`, we do not even use that flag. In that case, we'd enforce caller to populate a field which won't be used. 

I actually thought if I can pass both `flip_left` and `flip_right`as bool instead of `will_match` but then it ended up being relatively ugly in caller side. 

Let me know if you still have another preference 👍

Issue #959, opened 2025-01-22
Keeping `Ref` of dashmap over await points can lead to deadlocks


dkales:
This might be relevant for you @bgillesp @naure @iliailia: We did some experiments/benchmarks with the GPRC networking and used the impl in this repository as a starting point.

In our scenarios we managed to deadlock it quite consistently, due to it holding the Dashmap `Ref` you get from `.get` over await points.

e.g. https://github.com/worldcoin/iris-mpc/blob/d11df4bfd043e979f0bfa3eb682890cc61897fad/iris-mpc-cpu/src/network/grpc.rs#L56

Changing it to `tokio::sync::RwLock<HashMap>` does fix it, but might not be the greatest for perf.

IDK if you have hit that in your current development

naure:
That’s correct. I’m reviewing all these usages at the moment and probably move to async mutexes.

PR #1019, opened 2025-02-11
async SharedIrisesRef

naure:
- Replace `sync::RwLock` with `tokio::RwLock` to avoid blocking threads and deadlock hazards.
- Share the lock acquisition in `eval_distance_batch`.
- Remove verbose Debug.

Maybe related to https://github.com/worldcoin/iris-mpc/issues/959.


iliailia (approved):
Looks good!

PR #1020, opened 2025-02-11
Encapsulate GPU preprocessing

dkales:
This is the final step to address #985. This PR moves the GPU-relevant part of the preprocessing completely into the GPU Actor.
This step might cause slight performance regressions since this part was previously handled by the main server loop, so this should be cross-examined.

I also moved the BatchQuery structs and its relevant child structs into the common crate and created a trait that provides an interface to submit jobs to whatever implementation is backing the Iris Uniqueness service.
This means that there is only a single import leftover from the `iris_mpc_gpu` crate, which is the concrete implementation of the actor itself, which can now more easily be switched out.

### DIFF
--- a/iris-mpc-gpu/benches/preprocessing.rs
+++ b/iris-mpc-gpu/benches/preprocessing.rs
@@ -1,0 +1,77 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use iris_mpc_common::{
+    galois_engine::degree4::{
+        FullGaloisRingIrisCodeShare, GaloisRingIrisCodeShare, GaloisRingTrimmedMaskCodeShare,
+    },
+    helpers::smpc_request::UNIQUENESS_MESSAGE_TYPE,
+    iris_db::iris::IrisCode,
+    job::BatchQuery,
+};
+use iris_mpc_gpu::server::PreprocessedBatchQuery;
+use rand::thread_rng;
+use uuid::Uuid;
+
+pub fn criterion_benchmark_preprocessing(c: &mut Criterion) {
+    let mut group = c.benchmark_group("Preprocessing");
+    for batch_size in [1, 4, 16, 64] {
+        let mut batch_query = BatchQuery::default();
+        for _ in 0..batch_size {
+            let iris = IrisCode::random_rng(&mut thread_rng());
+            let [shares, _, _] =
+                FullGaloisRingIrisCodeShare::encode_iris_code(&iris, &mut thread_rng());
+            let code_shares_store = shares.code;
+            let mask_shares_store = shares.mask;
+            let mut code_shares_query = code_shares_store.clone();
+            let mut mask_shares_query = mask_shares_store.clone();
+            GaloisRingIrisCodeShare::preprocess_iris_code_query_share(&mut code_shares_query);
+            GaloisRingTrimmedMaskCodeShare::preprocess_mask_code_query_share(
+                &mut mask_shares_query,
+            );
+            let code_shares_query = code_shares_query.all_rotations();
+            let mask_shares_query = mask_shares_query.all_rotations();
+            let code_shares_db = code_shares_store.all_rotations();
+            let mask_shares_db = mask_shares_store.all_rotations();
+            batch_query.request_ids.push(Uuid::new_v4().to_string());
+            batch_query
+                .request_types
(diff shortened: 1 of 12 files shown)


iris-mpc-gpu/src/server/actor.rs
```diff
@@ -590,6 +590,10 @@ impl ServerActor {
         return_channel: oneshot::Sender<ServerJobResult>,
     ) -> eyre::Result<()> {
         let now = Instant::now();
+
+        // TODO: since we now moved this to the actor again, is this a performance
+        // bottleneck and should we parallelize this, etc?
+        let batch = PreprocessedBatchQuery::from(batch);
```
philsippl, line 596: let's add a quick bench for this? this it's not parallelized anyways, local bench should be somewhat representative

philsippl (approved):
lgtm! let's just bench the preprocessing first

iris-mpc-gpu/benches/preprocessing.rs
```diff
@@ -0,0 +1,77 @@
...
+            batch_query.db_right.code.extend(code_shares_db);
+            batch_query.db_left.mask.extend(mask_shares_db.clone());
+            batch_query.db_right.mask.extend(mask_shares_db);
+        }
+
+        group.bench_function(format!("batch_size={}", batch_size), |b| {
+            b.iter_batched(
+                || batch_query.clone(),
+                |batch| {
+                    let _ = PreprocessedBatchQuery::from(batch);
```
naure, line 64: It may be necessary to use [black_box](https://docs.rs/criterion/latest/criterion/fn.black_box.html) here.

iris-mpc-gpu/src/server/actor.rs
dkales: It's not free. Maybe just throwing a few `par_iters` at it is good enough?

naure (approved):
Very helpful!

iris-mpc-gpu/benches/preprocessing.rs
dkales: criterion actually uses that internally already, so it should not be needed in most cases

iris-mpc-gpu/src/server/actor.rs
dkales: par_iters actually slow this down for smaller batch sizes < 16.

iris-mpc-gpu/benches/preprocessing.rs
naure: Right, maybe it also works to just return the result (vs `let _`).

dkales:
```
Preprocessing/batch_size=1
                        time:   [761.52 µs 770.09 µs 781.14 µs]
                        change: [-84.465% -84.263% -84.015%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) high mild
  7 (7.00%) high severe
Preprocessing/batch_size=4
                        time:   [13.804 ms 13.888 ms 13.985 ms]
                        change: [-43.554% -43.064% -42.565%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  3 (3.00%) high mild
  8 (8.00%) high severe
Benchmarking Preprocessing/batch_size=16: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 8.6s, or reduce sample count to 50.
Preprocessing/batch_size=16
                        time:   [56.308 ms 56.669 ms 57.087 ms]
                        change: [-47.513% -46.930% -46.369%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe
Benchmarking Preprocessing/batch_size=64: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 29.6s, or reduce sample count to 10.
Preprocessing/batch_size=64
                        time:   [180.56 ms 180.97 ms 181.46 ms]
                        change: [-69.071% -68.994% -68.908%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe
```

PR #1022, opened 2025-02-12
Aurel/wal: Generalize recovery / rollback

naure:
Draft!

This PR is the beginning of a generalization of our "sync" or "rollback" concept. The iris store sync works by simply truncating the iris table therefore rolling back the latest insertions. But that does not work for deletions, updates, and certainly not for HNSW graph changes.

The proposal here is to migrate the current mechanism to something similar but more formal / general. Then we can tack our HNSW updates to it and have everything stay consistent.

Closely related to the ["roll forward" proposal](https://www.notion.so/worldcoin/Proposal-for-iris-mpc-modifications-sync-safeguards-18a8614bdf8c80ca8617cb0134e0d2e3).

The idea explored in this PR is a 2 phases prepare/commit mechanism:

1. After the MPC, nodes prepare their state mutations. After this a node is certain that it is capable of committing; e.g. the mutation contains iris codes, graph updates, results, etc.
2. Continue with the next batch of requests, the next MPC. If we reach this point, it means everybody has prepared the previous batch so everybody can commit. The background "result sender" thread can do that.
3. A node crashes, restarts, and everybody synchronizes about their highest prepared mutation. Everybody commits the highest common denominator, but rolls back the higher non-consensual mutations, if any.

The state is represented by a single monotonically increasing ID ("mutation_id") because we want consensus about the whole database (no granularity of individual entries).


iris-mpc-store/src/lib.rs
```diff
@@ -440,6 +444,60 @@ DO UPDATE SET right_code = EXCLUDED.right_code, right_mask = EXCLUDED.right_mask
...
+        let mut query = sqlx::QueryBuilder::new("INSERT INTO wal (action)");
+        query.push_values(actions, |mut query, action| {
+            query.push_bind(Json(action));
+        });
+        query.build().execute(tx.deref_mut()).await?;
+        Ok(())
+    }
+
+    pub async fn wal_get_state_id(&self) -> Result<u64> {
+        let id: (i64,) = sqlx::query_as("SELECT MAX(state_id) FROM wal")
```
naure, line 461: TODO: A separate table to track the latest applied action.

PR #1023, opened 2025-02-12
Aurel/server hawk

naure:
Goals of the PR is enabling smooth local development of the HNSW-based implementation of the SMPC protocol with all its infrastructure dependencies, as well as having a self-contained way of running the server without GPU dependencies.

Parts of the PR are:
- AWS stack through docker / LocalStack
- DBs using docker / postgres
- A self-contained server implementation that has no GPU dependencies 

2 options to run locally:

### Step 1: run ancillary services

```
>>> docker-compose up localstack dev_db
```

### Step 2: run service with the init script for Party 0

```
>>> ./scripts/run-server.sh 0 --init-servers # change to 1 or 2 for other parties
```

The script must run with `--init-servers` flag at least once. It will create some AWS resources, but after that it can run without it. 

```
>>> ./scripts/run-server.sh 1
>>> ./scripts/run-server.sh 2 # in a separate shell
```

## 2. Run everything with docker

Just run

```
>>> docker-compose up
```

It will bring up the 3 parties plus all the needed AWS/DB resources

# What else is missing
- [ ] SQS and SNS setup in LocalStack to send and receive messages, plus a lightweight client that can generate queries and send them to SNS/SQS
- [ ] Mapping of current inputs of `iris-mpc-gpu` to what `iris-mpc-cpu` expects. Namely, ensure that `BatchQuery` can be consumed by the `HawkActor`
- [ ] Mapping of the current outputs of `iris-mpc-cpu` to what we need to produce to be sent to SQS. In `iris-mpc-gpu`, this is encapsulated by the `ServerJobResult`. In `iris-mpc-cpu`, the construct used is `HawkResult`


iris-mpc-common/src/config/mod.rs
```diff
@@ -156,6 +156,9 @@ pub struct Config {
 
     #[serde(default)]
     pub enable_reauth: bool,
+
+    #[serde(default = "default_hawk_load_parallelism")]
+    pub hawk_load_parallelism: usize,
```
naure, line 161: This should be `hawk_request_parallelism`.

carlomazzaferro:
Linear task for tracking: https://linear.app/worldcoin/issue/POP-2157/local-environment-initial-setup-for-hnsw

carlomazzaferro (approved):

PR #1025, opened 2025-02-14
Aurel/persist-graph: Mutate the graph in the background thread

naure:
_On top of #1023._

- [x] Put each eye side into its own graph (with column `graph_id`).
    - [x] Test SQL migrations (`test_graph_migrations`).
- [x] Load the graph on startup (method `load_graph_store`).
    - [x] Test of the loading (`test_graph_load`).
- [x] Update the graph with results (`hawk_mutation.persist`).
- [x] Minimize code in `server_hawk.rs`.
- [x] Consolidate update of graph and iris store (same `tx` in `server_hawk.rs`).
- [x] Consolidate with the iris store (same schema, same `migrations` scripts, same connections).


iliailia:
Could you merge it with main to have a better view of relevant changes?

naure:
> Could you merge it with main to have a better view of relevant changes?

Done.

iris-mpc-common/src/job.rs
```diff
@@ -64,6 +64,8 @@ pub struct ServerJobResult {
     pub successful_reauths: Vec<bool>, // true if request type is reauth and it's successful
     pub reauth_target_indices: HashMap<String, u32>,
     pub reauth_or_rule_used: HashMap<String, bool>,
+    /// Actor-specific data (e.g. graph mutations).
+    pub actor_data: A,
```
dkales, line 68: I think this looks fine, just a design question:
Should we aim for the db insertion stuff to also be internal to the actor? This again would now add code that is unique to the CPU impl to the `server_hawk.rs`

iliailia:
Looks good to me so far.

iliailia (commented):
Looks good to me so far.

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
```diff
@@ -15,7 +15,11 @@ use sqlx::{
     types::{Json, Text},
     Executor, Postgres, Row, Transaction,
 };
-use std::{marker::PhantomData, str::FromStr};
+use std::{
+    marker::PhantomData,
+    ops::{Deref, DerefMut},
```
iliailia, line 20: `clippy` complains on this unused `Deref`

iris-mpc-common/src/job.rs
naure: Good point to try to minimize change in `server*.rs`. As it stands the GraphStore needs a plain old data structure from the Hawk search, but a complex interaction with the startup, database update, and crash recovery mechanisms.

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
naure: Fixed.

naure:
> * [ ]  Consolidate with the iris store (same schema).
> * [ ]  Consolidate update of graph and iris store (same `tx`).

Any opinions about this?

iliailia:
> > * [ ]   Consolidate with the iris store (same schema).
> > * [ ]   Consolidate update of graph and iris store (same `tx`).
> 
> Any opinions about this?

Sounds like a good idea to me. It's a bit annoying to have both a graph and an iris store as arguments for any HNSW searcher method. Plus, semantically these two objects must relate to each other anyway.

iris-mpc-common/src/job.rs
naure: I got it down to pure method calls - one for startup loading and one for updates.

naure:
> > > * [ ]    Consolidate with the iris store (same schema).
> 
> Sounds like a good idea to me. It's a bit annoying to have both a graph and an iris store as arguments for any HNSW searcher method. Plus, semantically these two objects must relate to each other anyway.

This question is about whether the config, SQL schema, and connection pools are the same for irises and graph, or separate, potentially two different physical databases behind them.

iliailia:
> > > > * [ ]     Consolidate with the iris store (same schema).
> > 
> > 
> > Sounds like a good idea to me. It's a bit annoying to have both a graph and an iris store as arguments for any HNSW searcher method. Plus, semantically these two objects must relate to each other anyway.
> 
> This question is about whether the config, SQL schema, and connection pools are the same for irises and graph, or separate, potentially two different physical databases behind them.

I see. IMHO, it'd be more robust to consolidate.

iliailia (approved):
Looks good to me!

naure:
> I see. IMHO, it'd be more robust to consolidate.

I merged the iris and graph stores (schemas, connection, test utils, etc).

PR #1028, opened 2025-02-14
sync mechanism for modifications - baseline

eaypek-tfh:
## Change
- Create the modifications table which will store reauth and deletion requests for now. We can expand it in the future with uniqueness as well.
- The next PR will implement the sync mechanism based on collected data.
- Store each deletion/reauth request when they're received with `status=IN_PROGRESS` and `persisted=false`.
- After the batch, update each deletion request with `status=COMPLETED` and `persisted=true`.
- After the batch, update each reauth request with `status=COMPLETED` and `persisted={success_result}`.

## Side Changes
- Introduce a `max_deletions_per_batch` config so that we can safely define a lookback size for modifications sync.
  - Note: Currently, deletion processor cron in signup-service sends all the deletion requests for which the cooldown has ended. Since deletions are not counted towards the batch size, we can potentially executed 100s of them. We'll change it so that it can send the same `max_deletions_per_batch` per cron execution.
 - Do not allow a serial id to have multiple modifications in a single batch.
 
## Links
- Design doc: https://www.notion.so/worldcoin/Proposal-for-iris-mpc-modifications-sync-safeguards-18a8614bdf8c80ca8617cb0134e0d2e3?pvs=4


iris-mpc-store/src/lib.rs
```diff
@@ -440,6 +464,98 @@ DO UPDATE SET right_code = EXCLUDED.right_code, right_mask = EXCLUDED.right_mask
...
+                persisted = data.persisted
+            FROM (
+                SELECT
+                    unnest($1::bigint[])  as id,
+                    unnest($2::text[])    as status,
+                    unnest($3::bool[])    as persisted
+            ) as data
+            WHERE modifications.id = data.id
+            "#,
+        )
```
eaypek-tfh, line 549: needed to use unnest to get the tests working. strangely, query builder did not work. i get [this error](https://github.com/worldcoin/iris-mpc/actions/runs/13331080536/job/37235110211) before [this commit](https://github.com/worldcoin/iris-mpc/pull/1028/commits/9f8630384d4cfce4abeba881e1f89dd0be281bc8)

let me know if you know of a way to avoid the `unnest` if you think it looks ugly 🙏

iris-mpc/src/bin/server.rs
```diff
@@ -229,6 +230,22 @@ async fn receive_batch(
                             })?;
                         metrics::counter!("request.received", "type" => "identity_deletion")
                             .increment(1);
+                        if modifications.contains_key(&identity_deletion_request.serial_id) {
+                            tracing::warn!(
+                                "Received another modification operation in batch: {}. Skipping",
+                                identity_deletion_request.serial_id
+                            );
+                            continue;
```
naure, line 238: This `continue` skips SQS `delete_message`. So the message will show up again, maybe going into a loop over the same message, or maybe breaking the deterministic ordering of messages.

eaypek-tfh: very good catch, thank you! i fixed it with [this commit](https://github.com/worldcoin/iris-mpc/pull/1028/commits/16cb5a93813c7b06de590e67a3677210b513cd37)

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -515,6 +515,7 @@ impl JobSubmissionHandle for HawkHandle {
                 successful_reauths: vec![],                             // TODO.
                 reauth_target_indices: Default::default(),              // TODO.
                 reauth_or_rule_used: Default::default(),                // TODO.
+                modifications: HashMap::new(),                          // TODO.
```
naure, line 518: Same as in the GPU actor:

```suggestion
                modifications: batch.modifications,
```

iris-mpc-store/src/lib.rs
naure: This seems to be the recommended way in sqlx doc. [How can I bind an array…](https://github.com/launchbadge/sqlx/blob/main/FAQ.md#how-can-i-bind-an-array-to-a-values-clause-how-can-i-do-bulk-inserts)

naure (approved):
Looks like a good direction to me.

iris-mpc-common/src/helpers/sync.rs
```diff
@@ -13,6 +14,26 @@ pub struct SyncResult {
     all_states: Vec<SyncState>,
 }
 
+pub const STATUS_IN_PROGRESS: &str = "IN_PROGRESS";
+pub const STATUS_COMPLETED: &str = "COMPLETED";
```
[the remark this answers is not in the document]
eaypek-tfh: I'll add a rust enum. i first opted for postgres enum and then abandoned it because it comes with lots of serializer/deserializer overhead between `Modification` (in sync.rs) and `StoredModification` (in store/lib.rs)

PR #1035, opened 2025-02-17
use shares from s3

eaypek-tfh:
## Change
- As a follow-up from https://github.com/worldcoin/iris-mpc/pull/1034, this PR makes use of the shares from s3 instead of the ones from SNS payload.
- We'll keep sending both S3 + SNS until this change is deployed. After that, we'll abandon SNS

## Side Change
- Rename `IrisCodesJSON` to `IrisCodeSharesJSON` to better represent underlying data

### DIFF
--- a/iris-mpc-common/src/helpers/smpc_request.rs
+++ b/iris-mpc-common/src/helpers/smpc_request.rs
@@ -109,11 +109,10 @@ pub const REAUTH_MESSAGE_TYPE: &str = "reauth";
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct UniquenessRequest {
-    pub batch_size:              Option<usize>,
-    pub signup_id:               String,
-    pub s3_key:                  String,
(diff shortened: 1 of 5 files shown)


iris-mpc-common/src/helpers/smpc_request.rs
```diff
@@ -195,12 +197,12 @@ pub struct IrisCodesJSON {
 }
 
 impl SharesS3Object {
-    pub fn get(&self, party_id: usize) -> Option<&String> {
+    pub fn get(&self, party_id: usize) -> (Option<&String>, Option<&String>) {
```
dkales, line 200: Should this be a single `Option<(&str, &str)>` instead?

eaypek-tfh: yeah, would be cleaner. updated 👍

danielle-tfh (approved):

PR #1036, opened 2025-02-17
CPU integration testgs: dockerfile and compose for server-client

carlomazzaferro:
In this PR we introduce a new GH Workflow for performing full integration tests against the hawk server. Main changes:
- A working docker-compose file for starting up infra, 3 nodes, and a client
- A new GH Workflow for running integration tests

Currently the integration tests won't succeed, since the server does not have the capabilities of actually processing batches. For now, all the client does is hits the healthcheck endpoints of the 3 servers after they have started up correctly (and established grpc, as well as HTTP connectivity between themselves)

Other QOL improvements:

- A new `.envrc` and `direnv.toml` file for loading test environment variables (using `direnv`)
- A new set of `.test.hawkX.env` files for setting up test environment variables

This should make it more consistent to run things with either docker or via the server startup script, as we managed to reduce a good amount of duplication in the env var setup

Tests run quite fast: docker build and compose up in total take ~4 minutes, so there's very little overhead added.


.github/workflows/cpu-integration-tests.yaml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Log in to the Container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and export to Docker
```
carlomazzaferro, line 20: We first build the docker image so we don't have to build it thrice when spinning up the docker compose

.test.hawk0.env
```diff
@@ -0,0 +1,5 @@
+SMPC__DATABASE__URL=postgres://postgres:postgres@dev_db:5432/SMPC_dev_0
+SMPC__REQUESTS_QUEUE_URL=http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/smpcv2-0-dev.fifo
+SMPC__NODE_HOSTNAMES='["0.0.0.0","hawk_participant_1","hawk_participant_2"]'
```
carlomazzaferro, line 3: need to pass local interface for grpc server for its own address due to the hawk main loading mechanics

iris-mpc-cpu/src/network/grpc.rs
```diff
@@ -152,8 +152,8 @@ impl GrpcNetworking {
     // TODO: from config?
     fn backoff(&self) -> ExponentialBackoff {
         ExponentialBackoff {
-            max_elapsed_time: Some(std::time::Duration::from_secs(2)),
-            max_interval: std::time::Duration::from_secs(1),
+            max_elapsed_time: Some(std::time::Duration::from_secs(60)),
```
carlomazzaferro, line 155: Increased to ensure images do not get killed on docker startup

scripts/run-client-docker.sh
```diff
@@ -0,0 +1,31 @@
...
+do
+  echo "Checking hawk\_participant\_${i} at <url>:300${i}/health..."
+  curl -f hawk_participant_${i}:300${i}/health
+done
+
+echo "All endpoints are healthy."
+
+
+#/bin/client \
+#    --request-topic-arn arn:aws:sns:$AWS_REGION:000000000000:iris-mpc-input.fifo \
```
carlomazzaferro, line 22: Once we have a functioning server, able to process requests, we can re-enable this

Dockerfile.hawk
```diff
@@ -1,4 +1,4 @@
-FROM  ubuntu:22.04 as build-image
+FROM  ubuntu:22.04 AS build-image
```
wojciechsromek, line 1: Let's maybe mark this file as `dev` version for clarity?

docker-compose.yml
```diff
@@ -8,6 +8,13 @@ services:
       POSTGRES_PASSWORD: "postgres"
     volumes:
       - ./scripts/init-db.sh:/docker-entrypoint-initdb.d/init-db.sh
+    healthcheck:
```
wojciechsromek, line 11: Re the file in general, what do you think about already preparing both versions, with and without the actual `hawk` services in the docker-compose? A standalone one with only the services bootstrapped will probably come in handy.

docker-compose.yml
```diff
@@ -18,56 +25,91 @@ services:
...
+      retries: 10
+
     volumes:
       - ./scripts/init-localstack.sh:/etc/localstack/init/ready.d/init-localstack.sh
 
   hawk_participant_0:
-    build:
-      context: .
-      dockerfile: Dockerfile.hawk
+    image: iris-mpc-hawk-server:latest
```
wojciechsromek, line 42: Should this be the `ghcr` link?

docker-compose.yml
```diff
@@ -18,56 +25,91 @@ services:
...
     ports:
-      - "4001:4001"
+      - "127.0.0.1:4001:4001"
       - "3001:3001"
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:3001/health"]
+      interval: 10s
+      retries: 10
+    entrypoint: ./run-server-docker.sh 1
+#    entrypoint: tail -f /dev/null
```
wojciechsromek, line 78: Can we remove the comments?

docker-compose.yml
```diff
@@ -18,56 +25,91 @@ services:
...
 
   hawk_participant_2:
-    build:
-      context: .
-      dockerfile: Dockerfile.hawk
+    image: iris-mpc-hawk-server:latest
+    env_file:
+      - .test.env
+      - .test.hawk2.env
```
wojciechsromek, line 84: Should this participant load both the env files?

scripts/run-server-docker.sh
```diff
@@ -1,8 +1,13 @@
...
 
-# needs to run twice to create the keys with both AWSCURRENT and AWSPREVIOUS states
-/bin/key-manager --region us-east-1 --node-id "$NODE_ID" --env dev rotate --public-key-bucket-name wf-dev-public-keys
-/bin/key-manager --region us-east-1 --node-id "$NODE_ID" --env dev rotate --public-key-bucket-name wf-dev-public-keys
+NODE_ID="$1"
+if [ -z "$NODE_ID" ]; then
+  echo "Usage: run-server.sh <node_id>"
+  exit 1
+fi
+
```
wojciechsromek, line 9: Let's leave the comment explaining why we are running it twice, wdyt?

docker-compose.yml
carlomazzaferro: I'll investigate this, or otherwise use the extending functionality https://docs.docker.com/compose/how-tos/multiple-compose-files/extends/

docker-compose.yml
carlomazzaferro: no, this is a locally built image as per the readme. Avoids re-building the image 3 times

docker-compose.yml
carlomazzaferro: all participants load 2 env files now, as it was simpler to have participant-specific env files instead of polluting the docker-compose

docker-compose.yml
wojciechsromek: Excerpt from readme, `docker build -f Dockerfile.hawk -t iris-mpc-hawk:latest .` the difference in naming confused me. Let's maybe name it something descriptive, like `locally-built-hawk` or something?

docker-compose.yml
wojciechsromek: ah, the diff did not display it very well, comment withdrawn 👍

docker-compose.yml
carlomazzaferro: sure, sounds good

docker-compose.test.yaml
```diff
@@ -0,0 +1,101 @@
+include:
+  - docker-compose.dev.yaml
```
carlomazzaferro, line 2: See this @wojciechsromek

README.md
```diff
@@ -133,8 +138,9 @@ The script must run with `--init-servers` flag at least once. It will create som
 
 Just run
 
-```
-docker-compose up
+```bash
+docker build -f Dockerfile.hawk -t hawk-server-local-build:latest .
```
carlomazzaferro, line 142: Name change here as well @wojciechsromek

Dockerfile.hawk
carlomazzaferro: Made this the prod image, built by a separate action

Dockerfile.dev.hawk
```diff
@@ -0,0 +1,50 @@
+FROM  ubuntu:22.04 AS build-image
```
carlomazzaferro, line 1: @wojciechsromek dev image

.test.hawk0.env
carlomazzaferro: probably would be good 👍 would be best to just pass the real hostnames to al of them instead of having to override their own

scripts/run-client-docker.sh
naure: When this command succeeds then the test is complete, correct?

carlomazzaferro: Yup. Locally we should be able to verify this by running the 3 services, and issuing this command. The server should be able to process all the SQS messages being generated.

Dockerfile.dev.hawk
```diff
@@ -0,0 +1,50 @@
...
+    debhelper \
+    ca-certificates \
+    protobuf-compiler \
+    wget
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
+ENV PATH "/root/.cargo/bin:${PATH}"
+ENV RUSTUP_HOME "/root/.rustup"
+ENV CARGO_HOME "/root/.cargo"
+RUN rustup toolchain install nightly-2024-07-10
```
wojciechsromek, line 22: @naure could we maybe start using a stable version here? this got quite old actually

.test.hawk0.env
naure: Done here: https://github.com/worldcoin/iris-mpc/pull/1038

wojciechsromek (approved):

Dockerfile.dev.hawk
naure: I’m all for it. There are some errors `use of unstable library feature` (`array_chunks` and `int_roundings`).

PR #1049, opened 2025-02-21
Add skipping persistance into request query

danielle-tfh:
### Notes
* before releasing e2e tests in prod - we want to ensure that tests can never result in any insertions
* add skip-persistence flag that ensures if the result is unique it is not inserted into the DB
* the result is sent back with no serial id

### Testing
* tested in staging with both unique and non-unique signups

### DIFF
--- a/iris-mpc-gpu/src/server/actor.rs
+++ b/iris-mpc-gpu/src/server/actor.rs
@@ -631,7 +631,8 @@ impl ServerActor {
                 && batch_size * ROTATIONS == batch.query_left_preprocessed.len()
                 && batch_size * ROTATIONS == batch.query_right_preprocessed.len()
                 && batch_size * ROTATIONS == batch.db_left_preprocessed.len()
-                && batch_size * ROTATIONS == batch.db_right_preprocessed.len(),
+                && batch_size * ROTATIONS == batch.db_right_preprocessed.len()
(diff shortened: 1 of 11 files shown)


iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1020,6 +1022,16 @@ impl ServerActor {
...
         );
 
+        // create a seperate matches list that includes the matches for skip persistence
+        let mut matches_with_skip_persistence = matches.clone();
+        skipped_unique_insertions.iter().for_each(|&idx| {
+            matches_with_skip_persistence[idx] = false;
+            tracing::info!(
+                "Matches with skip insertion request ID {}",
+                batch.request_ids[idx],
+            );
```
eaypek-tfh, line 1032: In theory, this should not happen anytime with e2e integration, right? What do you think of making this more obvious by logging it as error, or emitting a metric, etc?

eaypek-tfh (approved):
Looks clean to me! Nice job! 🚀

iris-mpc-gpu/src/server/actor.rs
danielle-tfh: I actually thought to instead allow this to be apart of the SMPC flow and not be something bad ie: SMPC is not aware of a test being bad but rather accepts certain functionality

PR #1053, opened 2025-02-22
Implement batch insertions for HNSW sorted lists

bgillesp:
PR implements batched insertions into sorted HNSW graph neighborhoods using a Batcher odd-even merge sort sorting network.  Generation of sorting networks and integration with sorted neighborhoods is complete, but the HNSW layer search algorithm still needs to be adjusted to make use of this batch insertion functionality, and `Aby3Store` still needs to provide a custom implementation for the `less_than_batch` function of the `VectorStore` trait to take advantage of parallel operations to reduce round complexity.


naure:
This can be used to parallelize insertions.

See the current serial code: [insert / insert_one / insert_prepare](https://github.com/worldcoin/iris-mpc/blob/c67aa45a1816a6636fa57fc8bdd6d7bd4da4dd6b/iris-mpc-cpu/src/execution/hawk_main.rs#L318).

iris-mpc-cpu/src/hnsw/sorting/batcher.rs
```diff
@@ -0,0 +1,239 @@
+use super::sorting_network::{SortingNetwork, SortingNetworkLayer};
+
+/// Generate sort tuples for specified stage of batcher merge network
+fn batcher_merge_step(deg: usize, stage: usize) -> SortingNetworkLayer {
+    let (offset, scale) = (1 << (deg - stage), 1 << (stage - 1));
```
iliailia, line 5: perhaps, it would be easier to read this function if the special case `stage = 1` is handled right at the start, i.e.,
```rust
let offset = 1 << (deg - stage);
if stage == 1 {
    return (0..offset).map(|start| (start, start + offset)).collect();
}
...
`

File: iris-mpc-cpu/src/hnsw/sorting/sorting_network.rs [hunk 2]
```diff
@@ -0,0 +1,96 @@
...
+
+    /// Uniformly shift indices of an input sorting network.  Panics if integer
+    /// overflow occurs during a shift operation.
+    pub fn shift(&mut self, shift_amount: isize) -> &mut Self {
+        self.map_indices(|x| x.checked_add_signed(shift_amount).unwrap())
+    }
+
+    /// Apply a filter to wires of the sorting network, optionally removing any
+    /// layers which are empty in the output
+    pub fn filter_wires<F>(&mut self, predicate: F, purge_empty: bool) -> &mut Self
```
iliailia, line 44: Do you need `purge_empty` at all? You use it being set only to `true` and empty layer seem useless anyway.

iliailia (approved):
Thanks, Bryan, amazing work given all the intricacies of merging sorted and unsorted arrays. I left a couple of minor comments.

iliailia:
It might be worth to add a reference description to Batcher sorting network. For example,

https://math.mit.edu/~shor/18.310/batcher.pdf

iris-mpc-cpu/src/hnsw/sorting/sorting_network.rs
bgillesp: Probably not necessary for our use -- my idea here was that there could be cases where empty layers need to be preserved when building up larger sorting networks from smaller ones to ensure that each layer consists of wires that don't interfere with each other, and so can be run in parallel.  As you say, this isn't something that comes up in our usage here, so probably it's worth eliminating the flag for readability.  I'll make the change, and we can bring back the functionality later if we really need it.

PR #1054, opened 2025-02-24
use sns message id to calculate batch hash

eaypek-tfh:
## Change
- We are currently calculating a `batch_hash` in actor.rs out of request_ids. This only encapsulates the enrollment and reauth request ids. It does not contain the deletion request ids.
- To have a generic way of calculating the `batch_hash` across all message types, this PR uses the `sns_message_ids` to calculate the `batch_hash`. This way, for any new message type addition, we just care about populating their sns message id and not change the hashing logic.

## Testing
- Tested in stage

### DIFF
--- a/iris-mpc-gpu/src/server/actor.rs
+++ b/iris-mpc-gpu/src/server/actor.rs
@@ -699,8 +699,8 @@ impl ServerActor {
         let tmp_now = Instant::now();
         tracing::info!("Syncing batch entries");
 
-        // Compute hash of the request ids concatenated
-        let batch_hash = sha256_bytes(batch.request_ids.join(""));
+        // Compute hash of the SNS message ids concatenated
+        let batch_hash = sha256_bytes(batch.sns_message_ids.join(""));
         tracing::info!("Current batch hash: {}", hex::encode(&batch_hash[0..4]));
 
         let valid_entries =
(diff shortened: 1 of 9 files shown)


iris-mpc-gpu/src/server/actor.rs
```diff
@@ -652,6 +652,23 @@ impl ServerActor {
...
+
+        // Compute hash of the SNS message ids concatenated
+        let batch_hash = sha256_bytes(batch.sns_message_ids.join(""));
+        tracing::info!("Current batch hash: {}", hex::encode(&batch_hash[0..4]));
+
+        let valid_entries =
+            self.sync_batch_entries(&batch.valid_entries, self.max_batch_size, &batch_hash)?;
+        let valid_entry_idxs = valid_entries.iter().positions(|&x| x).collect::<Vec<_>>();
+        batch_size = valid_entry_idxs.len();
+        batch.retain(&valid_entry_idxs);
```
philsippl, line 669: You should also remove those `sns_message_ids` from here to be consistent: https://github.com/worldcoin/iris-mpc/blob/e85d8d4c52a96302fb947708a1ef629b696ab8e6/iris-mpc-gpu/src/server/mod.rs#L174

philsippl (approved):

PR #1056, opened 2025-02-24
aurel/fix-plans-order: Decouple set_entry_point and set_links

naure:
Previously there was a requirement to set an entry point at a layer before inserting links (edges). But it doesn’t really make sense to "create" layers and that’s inconvenient with parallelism or batching.

- Accept links before entry point in new layers.
- Remove confusing reordering of plans (`swap`).


iliailia (approved):
LGTM!

PR #1057, opened 2025-02-24
Hawk mode config extensions

siajasl:
Until HNSW has proved itself, the server will be deploy-able in various modes, i.e. `gpu` | `cpu` | `cpu-shadow`.  The SPMC server software will adapt it's behavior according to currently configured mode.  This PR thus encapsulates introducing config settings for specifying mode.


siajasl:
Have added new config settings:
- New settings: `mode_of_compute` & `mode_of_deployment`.  
- Defaults: `ModeOfCompute::GPU` & `ModeOfDeployment::STANDARD`.
- Overrides in Hawk related .env & .sh files: `mode_of_compute=ModeOfCompute::CPU`.

Could merge this into main but keep the PR open for next phase.

iris-mpc/src/bin/server.rs
```diff
@@ -864,6 +864,19 @@ async fn server_main(config: Config) -> eyre::Result<()> {
     ));
     shutdown_handler.wait_for_shutdown_signal().await;
 
+    // Validate modes of compute/deployment.
+    if config.mode_of_compute == ModeOfCompute::GPU
```
naure, line 868: Actually only GPU / STANDARD is valid here.

iris-mpc/src/bin/server_hawk.rs
```diff
@@ -830,6 +830,19 @@ async fn server_main(config: Config) -> eyre::Result<()> {
     ));
     shutdown_handler.wait_for_shutdown_signal().await;
 
+    // Validate modes of compute/deployment.
+    if config.mode_of_compute == ModeOfCompute::GPU
```
naure, line 834: Only CPU is valid here (STANDARD or SHADOW).

naure (approved):
See comments for validation of the status quo.

siajasl:
@naure

Just updated config validation logic as per your comments ... I should have grokked that before submitting the PR.  I feel that formally the config loader should implicitly validate but for the moment we will proceed as is.

PR #1065, opened 2025-02-25
Shadow Mode: Iris Postgres R/O

siajasl:
When running in CPU-SHADOW mode, the Iris PostgreSQL dB service will be shared with the **production GPU dB**.  Other services will be isolated, i.e. SNS, SQS, S3 ...etc.  As a consequence, when running in CPU-SHADOW mode, the server process will no longer need to write to the Iris PostgreSQL dB.


siajasl:
Added skipping of persisting Iris postgres store state changes.

iris-mpc/bin/server_hawk.rs
```diff
@@ -12,7 +12,7 @@ use clap::Parser;
 use eyre::{eyre, Context, Report};
 use futures::{stream::BoxStream, StreamExt};
 use iris_mpc_common::{
-    config::{Config, ModeOfCompute, Opt},
+    config::{Config, ModeOfCompute, ModeOfDeployment, Opt},
```
danielle-tfh, line 15: Also, worth noting that we have https://github.com/worldcoin/iris-mpc/blob/main/iris-mpc-common/src/config/mod.rs#L83 `disable_persistence` - we previously used it in the actor for shadow mode for iris-mpc

danielle-tfh: I see you use it bellow and i wonder if HNSW should only use ModeOfDeployment to not have 2 configs influencing the modes

danielle-tfh (approved):

iris-mpc/bin/server_hawk.rs
```diff
@@ -1505,61 +1509,68 @@ async fn server_main(config: Config) -> eyre::Result<()> {
...
-
-            store_bg
-                .insert_results(&mut tx, &uniqueness_results)
-                .await?;
+            // Initialise dB transaction.
+            let mut tx = iris_pgres_store_bg.tx().await?;
 
-            // TODO: update modifications table to store reauth and deletion results
+            // Persist Iris postgres store state changes.
+            if config_bg.mode_of_deployment == ModeOfDeployment::STANDARD {
```
naure, line 1516: I think the shadow mode should still do these operations in its own database, for the purpose of exercising as much production functionality as possible.

Instead, the difference could be only at startup time - load data from the master read-only db instead of its own db. That could be a new (optional) db config and a separate Store object.

danielle-tfh: This makes me wonder if we should rather setup HNSW having its own database with a copy? Lets discuss in the sync today

iris-mpc/bin/server_hawk.rs
```diff
@@ -1360,63 +1360,64 @@ async fn server_main(config: Config) -> eyre::Result<()> {
...
```
carlomazzaferro, line 1367: It seems like this parens is never closed. Otherwise, let's resolve the merge conflicts / fmt / etc. and make a final decision on this?

PR #1068, opened 2025-02-26
aurel/fix-hawk-request: Increase stack size for integration tests [POP-2168]

naure:
- [x] Increase stack size for integration tests
  - Otherwise stack overflow in [SQS receive here](https://github.com/worldcoin/iris-mpc/blob/cc4d1198f3fec6e42d4b917e05e8a1933a9b6072/iris-mpc/bin/server_hawk.rs#L161).

- [x] Populate `ServerJobResult` with the right structure (`Vec` with a result for each request).

- [x] Convert `HawkMutation` to that same `Vec` format of the rest of `ServerJobResult`.

- [x] Produce valid `serial_id` in `merged_results`.

- [x] Put valid SNS Metadata in the test client.

- [x] Create SNS topic for results.

PR #1072, opened 2025-02-27
aurel/ci-client-request: Enable client request in CI

naure:
Follow-up to #1068.

### DIFF
--- a/scripts/run-client-docker.sh
+++ b/scripts/run-client-docker.sh
@@ -20,10 +20,10 @@ echo "All endpoints are healthy."
 
 # Prepare the results queue.
 SQS_IRIS_MPC_RESULTS_QUEUE_NAME=iris-mpc-results-us-east-1.fifo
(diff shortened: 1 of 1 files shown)


danielle-tfh:
@aurel-fr should be fixed with this: https://github.com/worldcoin/iris-mpc/pull/1076

naure:
`An error occurred (InvalidClientTokenId) when calling the GetQueueUrl operation: The security token included in the request is invalid.`

@danielle-tfh I am starting to think I am on the wrong track with the AWS CLI. Maybe I’ll do it in rust.

carlomazzaferro:
This might be fine for now, though we probably want to use the mechanics in the client to verify the messages that are produced: https://github.com/worldcoin/iris-mpc/blob/main/iris-mpc/src/client.rs#L146-L225 

@naure make sure to align also with @danielle-tfh as she was thinking about working on this

danielle-tfh:
@naure I am currently implementing it. Will send a PR when its complete

PR #1079, opened 2025-02-27
[POP-2213] initial modularization work

carlomazzaferro:
A simple attempt at keeping in `server_hawk.rs` only the main server function (and some leftover db loading) which can be further refactored later


iris-mpc/src/utils.rs
```diff
@@ -0,0 +1,158 @@
...
+        kms_dh::derive_shared_secret,
+        smpc_request::{CircuitBreakerRequest, ReceiveRequestError, SQSMessage},
+    },
+};
+use metrics_exporter_statsd::StatsdBuilder;
+use std::{backtrace::Backtrace, panic};
+use telemetry_batteries::tracing::{datadog::DatadogBattery, TracingShutdownHandle};
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
+
+pub async fn handle_circuit_breaker_message(
```
danielle-tfh, line 14: why is the `handle_circuit_breaker_message` here?

danielle-tfh: Yeah - maybe a good place would be the batch.rs

danielle-tfh:
Looking great Carlo! This is def a super good step in the refactoring direction 🚀

carlomazzaferro:
Superseeded by: https://github.com/worldcoin/iris-mpc/pull/1104

PR #1084, opened 2025-02-28
Aurel/matching logic

naure:
[Spec](https://www.notion.so/inversed/Design-left-right-matching-18fa3b540bfe8006a51ed2eaa8a5f902).

- [x] Classify nearest neighbors with MPC `is_match`.
    - [x] Using a bisection on sorted list.
    - [ ] Check whether it is true that sorted distances imply a bisection of `is_match` (see `match_count`).
- [x]  Join left and right results.
- [x]  AND rule on inner-join results.
- [x]  Anti-join results.
    - [x] Fetch the iris of the missing side (`calculate_is_match`).
    - [x] Parallelize.
    - [x] Call MPC `is_match` on them.
    - [x] Include these in the AND rule.
- [x]  Optimize is_match with bisection.


siajasl (approved):
Looks clean.  Nice comments.

PR #1086, opened 2025-03-03
Implement batched HNSW distance comparisons

bgillesp:
This PR implements a batched mode for HNSW layer search algorithm, which optimistically compares neighbor distances to current farthest neighbor in candidate list in batches, and inserts filtered elements into the candidate list in batches by sorting.  Altogether this should produce a marked improvement in the total number of communication rounds required by HNSW searches/insertions.

A few items which may call for additional attention in follow-up work:
- More systematic method for choosing target batch size for insertions
- Migrate Batcher odd-even merge sort network to parallel quicksort to reduce total number of comparisons

A small additional optimization has been included in this PR: layer search using `ef = 1` is now executed using a separate direct greedy algorithm that finds the nearest neighbor from a list using a binary tree sorting network to filter the minimum distance element to the start of the list in logarithmically many rounds.


iris-mpc-cpu/src/hnsw/sorting/tree_min.rs
```diff
@@ -0,0 +1,104 @@
...
+
+/// Generates a sorting network layer with wires between every other multiple of
+/// `2^stage` indices, for indices up to `2^deg` exclusive.
+fn tree_min_layer(deg: usize, stage: usize) -> SortingNetworkLayer {
+    if stage >= deg {
+        return Default::default();
+    }
+
+    // Here 0 <= stage < deg
+    let (increment, count) = (1 << stage, 1 << (deg - stage));
```
iliailias, line 37: Should `count` be equal to `1 << (deg - stage - 1)`? Otherwise, you generate twice more comparisons, which are later filtered by `filter_wires`.

iris-mpc-cpu/src/hnsw/sorting/tree_min.rs
```diff
@@ -0,0 +1,104 @@
+//! Implements a sorting network which identifies the minimum element of an
+//! unsorted list in logarithmically many rounds of parallel comparisons using
+//! a binary tree design.
+
+use super::sorting_network::{SortingNetwork, SortingNetworkLayer};
+
+/// Generates a sorting network which moves the minimum element of an unsorted
+/// list of size `length` to the starting index in logarithmically many rounds
+/// using a binary tree of comparison operations.
+pub fn tree_min(length: usize) -> SortingNetwork {
```
iliailias, line 10: It isn't a sorting network as it doesn't really sort array values. Perhaps, it would be more concise to rename `SortingNetwork` into ``SwapNetwork` or alike.

iris-mpc-cpu/src/hnsw/sorting/tree_min.rs
```diff
@@ -0,0 +1,104 @@
...
+    match length {
+        0 | 1 => SortingNetwork::new(),
+        _ => {
+            let deg = (usize::ilog2(length - 1) + 1) as usize;
+            let mut network = tree_min_base(deg);
+            network.filter_wires(|&(_, idx2)| idx2 < length);
```
iliailias, line 16: I think you can remove it once `tree_min_layer` is fixed.

naure:
Is it possible to keep the previous algorithm so we can test them against each other eventually?

iris-mpc-cpu/src/hnsw/sorting/tree_min.rs
bgillesp: Yep, nice catch!

iris-mpc-cpu/src/hnsw/sorting/tree_min.rs
bgillesp: Really good point!  I guess the technical name for these is "Comparator Networks", but `SwapNetwork` rolls off the tongue a bit better.  Will revise.

iris-mpc-cpu/src/hnsw/sorting/tree_min.rs
bgillesp: Probably should keep it -- the `tree_min_base` makes a swap network for a 2-power sized input list, while the `filter_wires` call removes wires that aren't needed for indices between `length` and the next 2-power, when `length` itself isn't a 2-power.

bgillesp:
> Is it possible to keep the previous algorithm so we can test them against each other eventually?

Already in there!  See the `layer_search_std` function in `HawkSearcher`.  Right now it actually still gets called for searches with `2 <= ef <= 31` because I wanted to include an implicit assumption of moderately large `ef` values for the current batched implementation.  In this "small `ef`" range it probably still would be more optimal to do some batching of operations, but it's not a high priority for us to handle since all of our `ef` values will be either larger than this or 1.

iris-mpc-cpu/src/hnsw/searcher.rs
```diff
@@ -354,16 +367,377 @@ impl HnswSearcher {
...
+
+        // TODO optimize batch size selection
+        let insertion_batch_size = ef / 6;
+        let mut insertion_queue: Vec<(V::VectorRef, V::DistanceRef)> =
+            Vec::with_capacity(4 * insertion_batch_size);
+
+        let mut comparison_queue: Vec<(V::VectorRef, V::DistanceRef)> = Vec::new();
+
+        // Numerator and denominator of estimated current insertion rate
+        const INS_RATE_NUM: usize = 64;
```
iliailia, line 486: Do you use 64 here to take into account that we do bit packing into u64 in bit operations?

iliailia (approved):
Looks good, Bryan! Great work!

Do you have an idea how this PR affects runtime over plaintext? Asking for a friend generating graphs...

iris-mpc-cpu/src/hnsw/searcher.rs
bgillesp: Nope, that's a total coincidence -- the insertion rate estimate we can think of as a multiplier on the insertion batch size to decide the number of new elements to inspect to get enough for insertion, so the encoding is `comparison batch size = insertion batch size / estimated insertion rate`, the latter of which can be rewritten as `insertion batch rate * (ins_rate_denom / 64)`.  So the 64 denominator in this reciprocal just gives enough granularity on the fixed precision multiplier to have some flexibility at smaller values.

bgillesp:
> Do you have an idea how this PR affects runtime over plaintext? Asking for a friend generating graphs...

I would expect the difference over plaintext to be fairly minimal -- everything on plaintext is still single-threaded, so batching doesn't give us a speed-up unless we leverage it with multithreading or optimized hardware ops.  If anything the changes do introduce a small overhead in terms of extra book-keeping in the layer search, some small inefficiencies in the graph traversal (a minimal increase in opened nodes), and an increase in the total number of distance comparisons.  These all should be low-order effects though, especially as distance comparisons already are lightweight in plaintext (check: `ad - bc < 0`).

PR #1088, opened 2025-03-03
Update Rust toolchain: nightly-2024-07-10 -> stable 1.81.0

carlomazzaferro:
- [x] remove upgrade protocol, which depends on `mpc-uniqueness-check`, which uses rust unstable features
- [x] bump rust version to stable 1.81.0 

There's one issue however. Our `rustfmt` depends on pretty much all nightly features. We can use nightly to lint and keep those configurations, but that will require everyone to have 2 channels locally, or we can just remove those configs and re-lint the projects. Thoughts?

Linting with stable channel: https://github.com/worldcoin/iris-mpc/pull/1092


iris-mpc-store/src/lib.rs
```diff
@@ -1,5 +1,3 @@
-#![feature(int_roundings)]
```
carlomazzaferro, line 1: Appeared unused. @philsippl mind double checking?

eaypek-tfh (approved):

PR #1089, opened 2025-03-03
Replay modification results upon startup

eaypek-tfh:
## Change
- Store `sns_message_body` field in modifications table and pass this around during start-up sync phase
- If a node is behind the most advanced modification, update the `node_id` field in sns message body and store the modification so that we replay the same result message with indication that it comes from current node
- After the start-up modification sync among nodes, replay the last `2 * (max_batch_size + max_deletions)` modifications to SNS


iris-mpc/bin/server.rs
```diff
@@ -693,6 +694,79 @@ fn get_iris_shares_parse_task(
...
+        "Sending {} last modifications to SNS. {} deletion, {} reauth",
+        last_modifications.len(),
+        deletion_messages.len(),
+        reauth_messages.len(),
+    );
+
+    if !deletion_messages.is_empty() {
+        send_results_to_sns(
+            deletion_messages,
+            &Vec::new(), // Not sure what these "empty" messages represent
```
dkales, line 746: this is the request metadata (node_id, trace_id, span_id) that is added to the sns message in case it is present.
I guess since it is not serialized into the DB we cannot send the original here (also it might not even make sense to send the orginal trace information)

iris-mpc/bin/server.rs
```diff
@@ -1171,13 +1245,23 @@ async fn server_main(config: Config) -> eyre::Result<()> {
...
     if config.enable_modifications_sync {
-        let (to_update, to_delete) = sync_result.compare_modifications();
+        let (mut to_update, to_delete) = sync_result.compare_modifications();
         tracing::info!(
             "Modifications to update: {:?}, to delete: {:?}",
             to_update,
             to_delete
         );
-        let to_update: Vec<&Modification> = to_update.iter().collect();
+        // Update node_id in each modification because they are coming more another more advanced node
```
dkales, line 1254: there is a `from` missing here I guess

iris-mpc/bin/server.rs
```diff
@@ -1225,6 +1309,17 @@ async fn server_main(config: Config) -> eyre::Result<()> {
         tx.commit().await?;
     }
 
+    // replay last 2 x `max_modification_lookback` modifications to SNS
```
dkales, line 1312: this `2x` does not seem to agree with the passed param?

iris-mpc/bin/server.rs
```diff
@@ -1446,25 +1537,32 @@ async fn server_main(config: Config) -> eyre::Result<()> {
                         match_ids[i].iter().map(|x| x + 1).collect::<Vec<_>>(),
                         *reauth_or_rule_used.get(&reauth_id).unwrap(),
                     );
-                    serde_json::to_string(&result_event)
+                    let result_string = serde_json::to_string(&result_event)
                         .wrap_err("failed to serialize reauth result")
+                        .expect("failed to serialize reauth result");
```
dkales, line 1542: I think wrap_err and expect are kind of redundant here.

dkales (approved):

dkales (approved):

PR #1098, opened 2025-03-05
aurel/refactor-vector-store: Use references for iris shares

naure:
Task https://www.notion.so/inversed/Robust-iris-memory-db-loading-1aca3b540bfe80cdacbec02845f62286?pvs=4

- [x] Move `insert` out of the `VectorStore` trait.
    - This clarifies that it is supposed to be read-only.
    - This is a step towards moving the `AbyStore` type to references instead of locks.
- [x] Use references to move around giant structures `GaloisRingSharedIris`.
- [x] Change the iris storage to `Vec<Arc>`.
    - This avoids (potential) allocation limits and copies of a flat `Vec`.
    - This makes it easier to handle shares in async contexts.
    - This introduces a layer of indirection which we will use for deletions and updates; e.g. change `Vec[index]` to `HashMap.get(vector_id)`.


iliailia (approved):
Thanks!

PR #1099, opened 2025-03-05
smaller scope refactor

carlomazzaferro:
refactor AWS config common functionality into its own module. Removed standalone functions to `utils.rs`

This is a self-contained PR with the goal of having an incremental improvement towards a larger scope refactor as outlined in https://github.com/worldcoin/iris-mpc/pull/1079/


iris-mpc-common/src/helpers/aws.rs
```diff
@@ -36,41 +33,3 @@ pub fn construct_message_attributes(
 
     Ok(message_attributes)
 }
-
-// This would only ever be leveraged if the code had isolated flows for every
-// message, leaving for now, maybe it will happen in the future
-pub fn trace_from_message_attributes(
```
carlomazzaferro, line 42: unused

iris-mpc/src/utils.rs
```diff
@@ -0,0 +1,119 @@
+use iris_mpc_common::{config::Config, helpers::kms_dh::derive_shared_secret};
+use metrics_exporter_statsd::StatsdBuilder;
+use std::{backtrace::Backtrace, panic};
+use telemetry_batteries::tracing::{datadog::DatadogBattery, TracingShutdownHandle};
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
+
+pub fn initialize_tracing(config: &Config) -> eyre::Result<TracingShutdownHandle> {
```
danielle-tfh, line 7: Thoughts on maybe calling this some form of initialisation mod instead?

carlomazzaferro: I will do that now, but I will also follow up with a PR that restructures a bit the code into something like this:

```
├── bin
│   ├── server
|   │   ├── iris_mpc_gpu.rs
│   │   └── iris_mpc_hnsw.rs
│   └── client.rs
└── src
    ├── server
    │   ├── init
    |   │   ├── mod.rs
    |   │   ├── tracing.rs
    |   │    .... # other initialization modules as needed, e.g. healthchecks
    │   │   └── aws.rs
    │   ├── init.rs
    └── lib
        ├── aws
        │   ├── mod.rs
        │   ├── clients.rs
        │   ├── s3.rs
        │   └── sns.rs
         .... # othes modules as needed once we refactor more 
        ├── gpu
        │   ├── mod.rs
        │   └── gpu.rs # gpu-specific server code
        └── hnsw
            ├── mod.rs
            └── hnsw.rs # hnsw-specific server code
```

Wdyt?

siajasl: You could factor out the AWS code into a top level library crate named: `iris-mpc-services`.  This crate would then have sub-modules: `aws`, `pgres` ...etc.

carlomazzaferro: I think at this point since the `services` will be mostly something used within the `iris-mpc` crate, and more specifically by the `iris_mpc_hnsw.rs`, I'll adopt your suggestion but make it a submodule of `iris-mpc`.

See this follow-up PR: https://github.com/worldcoin/iris-mpc/pull/1104

siajasl (approved):
Nicely scoped.  Moving in right direction.

iris-mpc/src/aws/sns.rs
```diff
@@ -0,0 +1,78 @@
+use aws_sdk_sns::{types::MessageAttributeValue, Client as SNSClient};
```
danielle-tfh, line 1: As a follow up - lets consider moving this to some form of results module. I think an AWS module should not be aware of the error or other application types of messages it is sending.

PR #1101, opened 2025-03-05
Add e2e client to read from file data

danielle-tfh:
### Notes
* refactor client tests to have different configs either reading from file or generation
* removed the use of db since it needs to be replaced with reading from postgres
* removed config params to make it cleaner

### DIFF
--- a/iris-mpc-store/src/lib.rs
+++ b/iris-mpc-store/src/lib.rs
@@ -412,10 +412,10 @@ DO UPDATE SET right_code = EXCLUDED.right_code, right_mask = EXCLUDED.right_mask
     }
 
     pub async fn get_max_serial_id(&self) -> Result<usize> {
-        let id: (i64,) = sqlx::query_as("SELECT MAX(id) FROM irises")
+        let id: (Option<i64>,) = sqlx::query_as("SELECT MAX(id) FROM irises")
             .fetch_one(&self.pool)
             .await?;
-        Ok(id.0 as usize)
+        Ok(id.0.unwrap_or(0) as usize)
     }
 
     pub async fn insert_results(
--- a/iris-mpc/src/client.rs
+++ b/iris-mpc/src/client.rs
@@ -1,8 +1,7 @@
 #![allow(clippy::needless_range_loop)]
-
 use aws_config::retry::RetryConfig;
 use aws_sdk_s3::Client as S3Client;
-use aws_sdk_sns::{config::Region, types::MessageAttributeValue, Client};
+use aws_sdk_sns::{config::Region, types::MessageAttributeValue, Client as SnsClient};
 use aws_sdk_sqs::Client as SqsClient;
 use base64::{engine::general_purpose, Engine};
 use clap::Parser;
@@ -22,11 +21,14 @@ use iris_mpc_common::{
         smpc_response::{create_message_type_attribute_map, UniquenessResult},
         sqs_s3_helper::upload_file_to_s3,
     },
-    iris_db::{db::IrisDB, iris::IrisCode},
+    iris_db::iris::IrisCode,
 };
 use rand::{rngs::StdRng, Rng, SeedableRng};
+use serde::Deserialize;
 use serde_json::to_string;
 use sodiumoxide::crypto::{box_::PublicKey, sealedbox};
+use std::fs::File;
+use std::io::Read;
 use std::{collections::HashMap, sync::Arc, time::Duration};
 use tokio::{
     sync::{Mutex, Semaphore},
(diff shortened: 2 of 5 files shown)


iris-mpc-store/src/lib.rs
```diff
@@ -412,10 +412,10 @@ DO UPDATE SET right_code = EXCLUDED.right_code, right_mask = EXCLUDED.right_mask
     }
 
     pub async fn get_max_serial_id(&self) -> Result<usize> {
-        let id: (i64,) = sqlx::query_as("SELECT MAX(id) FROM irises")
```
danielle-tfh, line 415: This is to allow for the clients to start up with an empty DB

iris-mpc/src/client.rs
```diff
@@ -85,370 +78,581 @@ pub struct Opt {
...
+
+                        match rng.gen_range(0..options) {
+                            0 => {
+                                println!("Sending new iris code for request id {}", request_id);
+                                {
+                                    let mut tmp = expected_results.lock().await;
+                                    tmp.insert(request_id.to_string(), None);
+                                }
+                                if use_data_from_file {
+                                    // TODO in future find a way to ensure no re-use of data
```
carlomazzaferro, line 309: These data contain an idComm, right? We could use that as our request id, or just populate each of the s3 entries with a new UUID and store it in a hashmap (future PR would be OK)

carlomazzaferro:
In general, it looks good. But I'll provide some overall feedback as it was fairly hard to follow the logic. My main issue with the how the client is structured is the amount of mutable properties the object has, and the amount of functions that mutate its state.

Particularly the states and contents of these 3:

```
            expected_results: Arc::new(Mutex::new(HashMap::new())),
            requests: Arc::new(Mutex::new(HashMap::new())),
            responses: Arc::new(Mutex::new(HashMap::new()))
```

Are hard to follow and know exactly what they are supposed to be (maybe more descriptive types would help). Something that can be done about this would be splitting the configuration of the client and its initialization from its actual runtime.

Something that would help would be extracting the the logic starting from https://github.com/worldcoin/iris-mpc/blob/fdade49324353622f27c2ad8b02c35d5beea0b4a/iris-mpc/src/client.rs#L292C21-L292C89 

```
let party_shares: IrisCodePartyShares = if random.unwrap_or(false) {
```

Into a separate function that gets called on initialization or on the `run` function, that returns all the inputs and outputs needed for the e2e test.

Another aspect worth thinking about is removing from the E2EClient struct some utility functions. `send_enrollment_request` and `send_enrollment_request` are good candidates, as they are fairly self contained and would require only the aws clients to be passed as parameters. Ideally even, put in a separate helper module.

iris-mpc/src/client.rs
danielle-tfh: Goood idea! let me do that now. It actually has a signup_id

danielle-tfh:
> In general, it looks good. But I'll provide some overall feedback as it was fairly hard to follow the logic. My main issue with the how the client is structured is the amount of mutable properties the object has, and the amount of functions that mutate its state.
> 
> Particularly the states and contents of these 3:
> 
> ```
>             expected_results: Arc::new(Mutex::new(HashMap::new())),
>             requests: Arc::new(Mutex::new(HashMap::new())),
>             responses: Arc::new(Mutex::new(HashMap::new()))
> ```
> 
> Are hard to follow and know exactly what they are supposed to be (maybe more descriptive types would help). Something that can be done about this would be splitting the configuration of the client and its initialization from its actual runtime.
> 
> Something that would help would be extracting the the logic starting from https://github.com/worldcoin/iris-mpc/blob/fdade49324353622f27c2ad8b02c35d5beea0b4a/iris-mpc/src/client.rs#L292C21-L292C89
> 
> ```
> let party_shares: IrisCodePartyShares = if random.unwrap_or(false) {
> ```
> 
> Into a separate function that gets called on initialization or on the `run` function, that returns all the inputs and outputs needed for the e2e test.
> 
> Another aspect worth thinking about is removing from the E2EClient struct some utility functions. `send_enrollment_request` and `send_enrollment_request` are good candidates, as they are fairly self contained and would require only the aws clients to be passed as parameters. Ideally even, put in a separate helper module.

Let me add some descriptions - lets not bloat this too much right now. Over time we can move other functions to a different place

iris-mpc/src/client.rs
```diff
@@ -80,7 +77,7 @@ pub struct Opt {
     pub max_concurrent_requests: usize,
 
     #[arg(long, env)]
-    pub use_data_from_file: bool,
+    pub use_data_from_file: Option<String>,
```
carlomazzaferro, line 80: rename this to `data_from_file`? Otherwise the `use` implies a boolean

carlomazzaferro (approved):
just one comment, otherwise I'm ok with this for now

iris-mpc/src/client.rs
danielle-tfh: Will do in the next refactor PR for this

PR #1104, opened 2025-03-05
modularization of iris-mpc-hnsw

carlomazzaferro:
Refactoring the `iris-mpc` package into sensible modules. No change to any functionality of logic is present here. 

We also keep the gpu server as is - with only minor modifications (straightforward removals of code that will likely remain as such for both servers (e.g. tracing and common seed setup))

```
├── Cargo.toml
├── bin                      # binaries only, no business logic should be included
│   ├── client
│   │   └── client.rs
│   ├── server
│   │   └── iris_mpc_hawk.rs
│   └── server.rs  
├── build.rs
└── src
    ├── lib.rs
    ├── client                # core client runtime logic
    │   └── mod.rs
    ├── server                # core server runtime logic
    │   ├── mod.rs
    │   └── utils.rs
    └── services
        ├── mod.rs
        ├── aws               # aws interaction logic
        │   ├── clients.rs
        │   ├── mod.rs
        │   ├── s3.rs
        │   └── sns.rs
        ├── init              # initialization logic (healtchecks, tracing, etc.)
        │   └── mod.rs
        ├── processors        # batch processors and iris processors 
        │   ├── batch.rs
        │   └── mod.rs
        └── store             # database interaction logic
            └── mod.rs


```


iris-mpc/bin/server.rs
```diff
@@ -8,9 +8,9 @@ use axum::{response::IntoResponse, routing::get, Router};
 use clap::Parser;
 use eyre::{eyre, Context, Report};
 use futures::{stream::BoxStream, StreamExt};
-use iris_mpc::aws::clients::AwsClients;
-use iris_mpc::aws::sns::{send_error_results_to_sns, send_results_to_sns};
-use iris_mpc::init::{initialize_chacha_seeds, initialize_tracing};
+use iris_mpc::services::aws::clients::AwsClients;
+use iris_mpc::services::aws::sns::{send_error_results_to_sns, send_results_to_sns};
```
danielle-tfh, line 12: We spoke about this in the other PR and this can be a follow up but i think `send_error_results_to_sns`, `send_results_to_sns` should maybe exist in some result file that abstracts all result logic

iris-mpc/src/server/mod.rs
```diff
@@ -1,704 +1,51 @@
...
+use eyre::{eyre, WrapErr};
+use iris_mpc_common::config::{Config, ModeOfCompute};
+use iris_mpc_common::helpers::inmemory_store::InMemoryStore;
+use iris_mpc_common::helpers::key_pair::SharesEncryptionKeyPairs;
+use iris_mpc_common::helpers::shutdown_handler::ShutdownHandler;
+use iris_mpc_common::helpers::smpc_request::{
+    ANONYMIZED_STATISTICS_MESSAGE_TYPE, IDENTITY_DELETION_MESSAGE_TYPE, REAUTH_MESSAGE_TYPE,
+    UNIQUENESS_MESSAGE_TYPE,
 };
+use iris_mpc_common::helpers::smpc_response::{
```
danielle-tfh, line 23: smpc_response may be a better place for those sns messages? thoughts?

carlomazzaferro: That'd be also a good option

iris-mpc/bin/server/iris_mpc_hawk.rs
```diff
@@ -0,0 +1,35 @@
+#![allow(clippy::needless_range_loop)]
```
danielle-tfh, line 1: I am finding the setup a bit confusing here --> server folder but server file not in the folder. Should we move it in?

iris-mpc/src/services/processors/batch.rs
```diff
@@ -0,0 +1,514 @@
+use aws_sdk_s3::Client as S3Client;
```
danielle-tfh, line 1: Nice! I like this abstraction a lot

iris-mpc/src/services/store/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+        "Aurora Loading summary => Loaded {:?} items. Waited for stream: {:?}, Loaded into \
+         memory: {:?}",
+        record_counter - n_loaded_via_s3,
+        time_waiting_for_stream,
+        time_loading_into_memory,
+    );
+}
+
+#[allow(clippy::too_many_arguments)]
+pub async fn load_db(
```
danielle-tfh, line 65: I feel like this folder should be in load rather than store? or at least called store?

iris-mpc/bin/server/iris_mpc_hawk.rs
carlomazzaferro: the idea was to have the `server` folder contain just the binaries for `iris_mpc_hnsw` and `iris_mpc_gpu`. I can make that change if it makes it clearer

iris-mpc/src/services/store/mod.rs
carlomazzaferro: The thinking here is that in the future this module will become something more abstracted, e.g. an `IrisLoader` or something alike. For now, it is just a place to put logically cohesive functionality related to DB operations

PR #1116, opened 2025-03-07
factor out server job processor

carlomazzaferro:


iris-mpc/src/services/processors/mod.rs
```diff
@@ -1,4 +1,6 @@
 pub mod batch;
+pub mod job;
+pub mod message;
```
danielle-tfh, line 3: How about message --> result/result_message?

danielle-tfh (approved):

danielle-tfh (approved):

PR #1119, opened 2025-03-08
Migrate gRPC exponential backoff dependency

bgillesp:
Migrates from `backoff` crate to `backon` crate for an implementation of exponential backoff for gRPC functionality.

The `backoff` crate is unmaintained and subject to security advisory [RUSTSEC-2025-0012](https://rustsec.org/advisories/RUSTSEC-2025-0012), and similar crate `backon` is recommended as an actively maintained alternative.


iliailia (approved):
Should be good, thanks!

PR #1120, opened 2025-03-10
sync queues based on sns sequence number

eaypek-tfh:
## Change
- This PR is part 1/2 of switching from postgres `sync` table based SQS sync to SNS sequence number based SQS sync.
- In this PR, we keep using the `skip_request_ids` based on sync table and we keep populating this table. The aim is to switch back to this old flow if smt goes wrong with the new SQS sync flow. This is also why it's behind feature flag.
- In an additional PR (after 2-3 weeks), we'll completely drop the `sync` table and only rely on the new SQS sync flow.

## Why this change?
- Currently, we only support skipping uniqueness requests but the queues can be desynced on deletion, reauth, or any future request types.
- Instead of updating our logic for each new request type, syncing on SNS sequence number would be fully generic and future-proof.

## Notes & Refs
- From AWS docs ([link](https://docs.aws.amazon.com/sns/latest/dg/fifo-topic-message-ordering.html)): The sequence number is a large, non-consecutive number that Amazon SNS assigns to each message. The length of the sequence number is 128-bits, and continues to increase for each [Message Group](https://docs.aws.amazon.com/sns/latest/dg/fifo-message-grouping.html).
- As an additional follow-up during part-2, we can parse the message body and write related error results to SNS

## Tests
- Tested in stage several times by deleting various number of items from queues randomly, with empty queues, and with already synced queues

### DIFF
--- a/iris-mpc-common/src/helpers/sqs.rs
+++ b/iris-mpc-common/src/helpers/sqs.rs
@@ -139,8 +139,7 @@ pub async fn delete_messages_until_sequence_num(
                 break;
             }
         } else {
-            tracing::info!("No more messages in the queue. Done.");
-            break;
+            tracing::info!("Could not find more messages in the queue. Retrying.");
         }
     }
 
--- a/iris-mpc-common/src/helpers/sqs.rs
+++ b/iris-mpc-common/src/helpers/sqs.rs
@@ -1,0 +1,148 @@
+use crate::config::Config;
+use crate::helpers::smpc_request::SQSMessage;
+use aws_sdk_sqs::Client;
+use eyre::Context;
+
+/// SQS messages contain a sequence number when they originate from SNS and raw message delivery is enabled.
+/// This function reads the top of the requests SQS queue and returns its sequence number.
+pub async fn get_next_sns_seq_num(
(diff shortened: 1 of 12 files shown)


iris-mpc-common/src/helpers/sqs.rs
```diff
@@ -0,0 +1,148 @@
...
+                    .queue_url(&config.requests_queue_url)
+                    .receipt_handle(msg.receipt_handle.clone().unwrap_or_default())
+                    .visibility_timeout(0)
+                    .send()
+                    .await
+                    .context("while changing message visibility in SQS")?;
+                break;
+            }
+        } else {
+            tracing::info!("No more messages in the queue. Done.");
```
dkales, line 142: Similar to the related PR, this branch seems to have invalid state if it is reached, since we have not actually reached the target sequence number, so I feel erroring/waiting should be the correct choice here instead

eaypek-tfh: Thank you, this was a copy left-over. Fixed it by not breaking here 👍

dkales (approved):

PR #1124, opened 2025-03-11
Explicit gRPC networking concurrency handler

iliailia:
This introduces a `GrpcHandle` struct that runs a task loop (see `GrpcHandle::new`) to handle networking requests. This allows to remove all the thread-safe data structures (such as `RwLock`) that store gRPC clients, incoming and outgoing channels in `GrpcNetworking`.

This PR doesn't change the performance of local benchmarks.


iris-mpc-cpu/src/network/grpc.rs
```diff
@@ -29,94 +28,309 @@ use tonic::{
...
+
+        // Loop to handle incoming tasks from job queue
+        tokio::spawn(async move {
+            while let Some(job) = rx.recv().await {
+                let job_result = match job.task {
+                    GrpcTask::Send(task) => grpc
+                        .send(task.value, &task.receiver, &task.session_id)
+                        .map(|_| MessageResult::Empty),
+                    GrpcTask::Receive(task) => grpc
+                        .receive(&task.sender, &task.session_id)
```
naure, line 159: This will block everything until the message is received. Effectively that makes the whole process sequential. For example the sends of other sessions will have to wait in line.

Instead, this must spawn a new task and not await. The `return_channel` goes into the task. Then, in a long, long time (for a CPU), the message will show up, and the background task will move it to the return channel.

There are other approaches possible too but that is the simplest right here.

naure: This does not apply to Send because it is not async.

This does applies to the other operations too, although they are less performance critical, but you could do this same spawn approach for all operations (except Send).

iliailia: Yep, I see the problem now...

The problem is that spawning a new thread will cause related parts of `GrpcNetworking` to be wrapped up in a thread-safe structs, which will cause locking. So to reduce it to the minimum, I'd go with spawning only `receive` as it'll cause wrapping only `UnboundedReceiver` of `MessageQueueStore` into `Mutex`.

iliailia: Done, see the most recent commit.

iris-mpc-cpu/src/network/grpc.rs
```diff
@@ -29,94 +28,334 @@ use tonic::{
...
+            while let Some(job) = rx.recv().await {
+                match job.task {
+                    GrpcTask::Send(task) => {
+                        let job_result = grpc
+                            .send(task.value, &task.receiver, &task.session_id)
+                            .map(|_| MessageResult::Empty);
+                        let _ = job.return_channel.send(job_result);
+                    }
+                    GrpcTask::Receive(task) => {
+                        let mut grpc = grpc.clone();
```
naure, line 167: That is a really expensive clone. This could lookup the message queue first, and clone only an `Arc` to send into the `spawn`.

The thing being cloned could be one of:

- Lookup the session and pass `Arc<MessageQueueStore>`. Need an Arc but no need for Mutex once it is constructed,
- or lookup up the session and sender too and send `Arc<Mutex<UnboundedReceiver<SendRequest>>>`,
- or for the brave, take the lock and send [`OwnedMutexGuard<UnboundedReceiver<SendRequest>>`](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#method.lock_owned)

naure: I think option 2 is the best, WDYT?

iliailia: Yep, the second options works nicely.

iliailia: Done

naure (approved):

PR #1125, opened 2025-03-11
feat: provide environment values for ampc-hnsw deployment

jazzbee:
Provide values for app deployment

### DIFF
--- a/deploy/stage/ampc-hnsw-0-stage/values-ampc-hnsw.yaml
+++ b/deploy/stage/ampc-hnsw-0-stage/values-ampc-hnsw.yaml
@@ -70,3 +70,50 @@ env:
 
   - name: SMPC__IMAGE_NAME
     value: $(IMAGE_NAME)
+
+
+initContainer:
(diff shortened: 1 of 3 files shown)


deploy/stage/ampc-hnsw-0-stage/values-ampc-hnsw.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - name: SMPC__MODE_OF_COMPUTE
+    value: "Cpu"
+
+  - name: SMPC__MODE_OF_DEPLOYMENT
+    value: "ShadowIsolation"
+
+  - name: SMPC__DATABASE__URL
+    valueFrom:
+      secretKeyRef:
+        key: DATABASE_AURORA_HNSW_URL
```
danielle-tfh, line 23: cc @jazzbee we only need this database URL to be passed through - the other one is not needed for staging shadow mode

kkowalski (approved):

PR #1126, opened 2025-03-11
Add configuration for using seperate HNSW cluster for staging deployment

danielle-tfh:
### Notes
* add config for the separate HNSW aurora DB clusters
* for staging isolation deployment, we should use the HNSW aurora DB cluster to separate the GPU and CPU implementations.
   * This will allow us to run testing in CPU that is not dependent on a sync protocol

### DIFF
--- a/iris-mpc-common/src/config/mod.rs
+++ b/iris-mpc-common/src/config/mod.rs
@@ -41,6 +41,9 @@ pub struct Config {
     pub database: Option<DbConfig>,
 
     #[serde(default)]
+    pub cpu_database: Option<DbConfig>,
+
+    #[serde(default)]
     pub aws: Option<AwsConfig>,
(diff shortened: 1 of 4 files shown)


iris-mpc-common/src/config/mod.rs
```diff
@@ -194,20 +197,23 @@ pub struct Config {
 #[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
 pub enum ModeOfCompute {
     /// Computation with standard CPUs (see HNSW graph).
-    CPU,
```
danielle-tfh, line 197: in rust, enums are meant to be camel case and not all in caps.
Same for changes bellow

carlomazzaferro (approved):

PR #1127, opened 2025-03-11
Get actor e2e tests working

philsippl:


danielle-tfh (approved):

.github/workflows/cpu-e2e-tests.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Set Rust as default
+        if: steps.changed-files-yaml.outputs.src_any_changed == 'true'
+        run: rustup default 1.81.0
+
+      - name: Run CPU E2E tests
+        if: steps.changed-files-yaml.outputs.src_any_changed == 'true'
+        run: cargo test --release -p iris-mpc-cpu -- e2e --test-threads=1 --include-ignored
+        env:
+          RUST_LOG: info
+          RUST_BACKTRACE: 1
```
dkales, line 60: This is also still enable in the "rust unit tests" section of the CI pipeline as well, with continue-on-error, so probably can remove it there?

dkales (approved):

philsippl:
hmm gpu still failing, am I missing something @dkales?

dkales:
found one diff @philsippl, lets see if this fixes it (maybe should also just look for a way to have the non_par/par variants of this produce the same DB for same seed

philsippl:
ahh right, that should be it, I think there's is no need for this par version anyways with those smallish dbs

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -561,21 +573,27 @@ impl HawkResult {
...
+        let match_ids = self.match_ids();
         self.connect_plans.0[0]
             .iter()
-            .map(|plan| {
-                plan.as_ref()
-                    .map_or(0, |plan| plan.inserted_vector.to_serial_id())
+            .enumerate()
+            .map(|(idx, plan)| match plan {
+                Some(plan) => plan.inserted_vector.to_serial_id(),
+                None => match_ids[idx][0],
```
naure, line 582: This may panic with out-of-bounds if there are no matches.

naure, line 582: This assumes that there is at least one match, based on the absence of an insertion. This could be more explicit or more defensive with `.first().unwrap_or(0)` as this may not always work (maybe because of "persistence" settings and what not).

naure:
Awesome!

iris-mpc-cpu/src/execution/hawk_main.rs
philsippl: wouldn't then be a plan if there is no match? that was my assumption, maybe wrong.

philsippl: returning just 0 is not safe, we should always either have a match and return the corresponding index or return the newly inserted entry idx.

naure: There is a config `disable persistence` and also requests with `skip persistence`. Not sure how it is all supposed to work with `merged_results` though. But probably the field `is_matches` will be more reliable than `connect_plans`.

https://github.com/worldcoin/iris-mpc/blob/2a1162b7c8b622bf472a4b4069151a27b1bc82d1/iris-mpc-cpu/src/execution/hawk_main.rs#L766

https://github.com/worldcoin/iris-mpc/blob/2a1162b7c8b622bf472a4b4069151a27b1bc82d1/iris-mpc-common/src/job.rs#L59

philsippl: yeah definitely point taken, that we should make this more robust.

PR #1130, opened 2025-03-12
abandon postgres auto increment id in modifications

eaypek-tfh:
## Change
- To sync modifications across nodes, we need the id to be the same in all nodes. However, postgres auto-incremented id is causing us problems. We expect it to increase consecutively but it have gaps although we reset the identity explicitly after each modification sync upon start-up. Attached shows a state of it from stage db
  image
- To fix this, this PR switches into `sns request id` + modification `created_at` timestamp (in microsecond precision). Insertion timestamp is only used to get the last `2 x (batch_size + max_deletions_per_batch)` modifications. Application code is unaware of it. We now group modifications by sns request id which is contained in fanned out (x3 - one per party) SQS message bodies.
  image

## Other considerations
- We could use sns sequence number as well. However, if something in the future forces us to switch into a different SNS or re-create the SNS, then sequence number will probably be reset
- We could still use a sequential id maintained by ourselves, instead of postgres. We could fetch the max id before inserting a modification and pass it while insertion. The advantage of this one is we avoid `created_at`. But we wanted to avoid this burden by just using sns request id. Please comment if it's more desirable especially with HNSW sync domain


iris-mpc-store/migrations/20250313101530_modifications_table_v2.up.sql
```diff
@@ -0,0 +1,16 @@
+-- drop the old modifications table if it exists
+DROP TABLE IF EXISTS modifications;
+
```
dkales, line 3: this means that the rollout of this commit MUST be on a clean shutdown, right? since they cannot use the old modification table some special care needs to be taken when applying this

eaypek-tfh: Currently, the modifications sync upon start-up is disabled already because we had a problem. Same modification was having different `id`s. So, the current table contains garbage due to this postgres auto-increment id being non-consecutive and differing across nodes. So, it's safe to drop the current table.

It's also safe because current modifications db only contains deletions and they are retried periodically.

siajasl: Obvious question but is there a scenario in which the modification history maybe required ?

iris-mpc-store/migrations/20250313101530_modifications_table_v2.up.sql
siajasl, line 1: If the new id column is fixed size text then could use `VARCHAR(size)` rather than TEXT as column type.

iris-mpc-store/migrations/20250313101530_modifications_table_v2.up.sql
```diff
@@ -0,0 +1,16 @@
...
+-- create the new modifications table
+CREATE TABLE IF NOT EXISTS modifications (
+    id TEXT PRIMARY KEY,
+    serial_id BIGINT NOT NULL,
+    request_type TEXT NOT NULL,
+    s3_url TEXT,
+    status TEXT NOT NULL,
+    persisted BOOLEAN NOT NULL DEFAULT FALSE,
+    created_at BIGINT NOT NULL DEFAULT (EXTRACT(EPOCH FROM NOW()) * 1000000)::BIGINT -- microseconds precision
```
naure, line 12: This is still taking a similar risk with database-`NOW()` as before with database-auto-increment.

You could generate your own `created_at` so you can make guarantees (e.g. that it strictly increases between insert).

naure: Or the guarantee can go into the SQL INSERT:

`created_at` is the greatest of either `NOW()`, or `SELECT max(created_at) + 1`.

iris-mpc-store/migrations/20250313101530_modifications_table_v2.up.sql
eaypek-tfh: `created_at` field is fully removed and opted for self-incremented id instead

iris-mpc-store/migrations/20250313101530_modifications_table_v2.up.sql
eaypek-tfh: `created_at` field is fully removed and opted for self-incremented id instead

naure (approved):
LGTM.

naure:
What do you want to happen on `delete_modifications` though?

Here the IDs will reset to 1, right?

eaypek-tfh:
> What do you want to happen on `delete_modifications` though?
> 
> Here the IDs will reset to 1, right?
> 
> It would be cleaner to continue counting and never reuse IDs (like postgres sequence generator was trying to do).

@naure why do you think it'd be reset to 1? Modifications are only deleted from last ids. I tested this already in stage by marking all modifications in progress in all nodes. All the nodes deleted the modifications and the next new modification picked up the old deleted modification id -> thus, having no gap or inconsistencies.

naure:
Got it. So that depends on the sync logic of `server_main` that should never make the table empty.

PR #1132, opened 2025-03-12
Add eval_distance with batches of queries

iliailia:
This PR changes only `VectorStore` API and its `Aby3Store` implementation. The actual batching of queries in `is_match_batch.rs` should be implemented in a follow-up PR depending on the decision made around [this comment](https://github.com/worldcoin/iris-mpc/pull/1117#discussion_r1989898788).

### DIFF
--- a/iris-mpc-cpu/src/execution/hawk_main/is_match_batch.rs
+++ b/iris-mpc-cpu/src/execution/hawk_main/is_match_batch.rs
@@ -79,20 +79,20 @@ async fn per_session(
 
     let mut out = Vec::with_capacity(tasks.len());
     for (query, vectors) in tasks {
-        let matches = per_query(query, &vectors, &mut session).await;
+        let matches = per_query(&[query], &vectors, &mut session).await;
         out.push(matches);
     }
     out
 }
 
 async fn per_query(
-    query: QueryRef,
+    queries: &[QueryRef],
     vector_ids: &[VectorId],
     session: &mut HawkSession,
 ) -> MapEdges<bool> {
     let distances = session
         .aby3_store
-        .eval_distance_batch(&query, vector_ids)
+        .eval_distance_batch(queries, vector_ids)
         .await;
 
     let is_matches = session.aby3_store.is_match_batch(&distances).await;
--- a/iris-mpc-cpu/src/execution/hawk_main/is_match_batch.rs
+++ b/iris-mpc-cpu/src/execution/hawk_main/is_match_batch.rs
@@ -79,20 +79,20 @@ async fn per_session(
 
     let mut out = Vec::with_capacity(tasks.len());
     for (query, vectors) in tasks {
-        let matches = per_query(&[query], &vectors, &mut session).await;
+        let matches = per_query(query, &vectors, &mut session).await;
         out.push(matches);
     }
     out
 }
 
 async fn per_query(
-    queries: &[QueryRef],
+    query: QueryRef,
     vector_ids: &[VectorId],
     session: &mut HawkSession,
 ) -> MapEdges<bool> {
     let distances = session
         .aby3_store
-        .eval_distance_batch(queries, vector_ids)
+        .eval_distance_batch(&[query], vector_ids)
         .await;
 
     let is_matches = session.aby3_store.is_match_batch(&distances).await;
--- a/iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
+++ b/iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
@@ -285,19 +285,24 @@ impl VectorStore for Aby3Store {
         self.lift_distances(dist).await.unwrap()[0].clone()
     }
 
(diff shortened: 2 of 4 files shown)


iris-mpc-cpu/src/execution/hawk_main/is_match_batch.rs
```diff
@@ -79,20 +79,20 @@ async fn per_session(
...
     let mut out = Vec::with_capacity(tasks.len());
     for (query, vectors) in tasks {
-        let matches = per_query(query, &vectors, &mut session).await;
+        let matches = per_query(&[query], &vectors, &mut session).await;
         out.push(matches);
     }
     out
 }
 
 async fn per_query(
```
dkales, line 88: I guess this name is then kind of misleading now

iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
```diff
@@ -285,19 +285,24 @@ impl VectorStore for Aby3Store {
...
-        let pairs = vectors
+        let pairs = queries
             .iter()
-            .map(|vector| (&query.processed_query, &**vector))
+            .flat_map(|q| {
+                vectors
+                    .iter()
+                    .map(|vector| (&q.processed_query, &**vector))
+                    .collect::<Vec<_>>()
+            })
```
dkales, line 305: This flattens it here, but does not unflatten into per query results. Is this fine to leave up to the caller or is the internal unflattening a nicer interface?

iliailia: At the moment, `eval_distance_batch` is usually followed by `less_than_batch` or `is_match_batch` that accept flattened vectors. And the current design in `hawk_main` does unflattening.

iris-mpc-cpu/src/execution/hawk_main/is_match_batch.rs
iliailia: True, but I'd leave this to a follow-up PR.

naure:
Just a quick look.

Btw what is not used by the HNSW searcher does not have to be changes in the VectorStore trait, we can also add Aby3 methods.

It is unclear atm if the distance evaluations is supposed to be a batch of pairs, or a Cartesian products (all queries against all be vectors).

In calculate_is_match I don't know whether we should increase or decrease the batch size size (versus parallelism).

iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
naure: This does look incorrect or at least surprising to me too. This does a flatten / Cartesian product, since it is only called with one element in this PR.

I think this should be documented and have a different name than just "_batch".

naure: Then we can use the function in follow up PRs.

iris-mpc-cpu/src/hnsw/vector_store.rs
```diff
@@ -63,12 +63,14 @@ pub trait VectorStore: Clone + Debug {
     /// Override for more efficient batch distance evaluations.
     async fn eval_distance_batch(
         &mut self,
-        query: &Self::QueryRef,
+        queries: &[Self::QueryRef],
         vectors: &[Self::VectorRef],
     ) -> Vec<Self::DistanceRef> {
         let mut results = Vec::with_capacity(vectors.len());
```
naure, line 69:
```suggestion
        let mut results = Vec::with_capacity(queries.len() * vectors.len());
```

iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
naure: Ok that's what the PR description said :)

iris-mpc-cpu/src/execution/hawk_main/is_match_batch.rs
naure: Better move the new `&[query]` inside of `per_query` then.

iliailia: Done

iris-mpc-cpu/src/hnsw/vector_store.rs
iliailia: Done

dkales (approved):

iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
iliailia: What action do you want me to do here? For clarity, I can unflatten the result right here. And in a later PR, we can change it if necessary.

naure: Nothing, I just misunderstood your intention before.

naure (approved):

PR #1139, opened 2025-03-13
Update HNSW host names

danielle-tfh:
### Notes
* use correct ordering for HNSW nodes


wojciechsromek (approved):

naure:
Are those 0.0.0.0 necessary somewhere?

There was at least one fix about this: https://github.com/worldcoin/iris-mpc/pull/1038/files

PR #1149, opened 2025-03-14
conifigure disable deletion / reauths from hawk, remove circuit break…

carlomazzaferro:
…er message

### DIFF
--- a/iris-mpc/src/services/processors/batch.rs
+++ b/iris-mpc/src/services/processors/batch.rs
@@ -19,9 +19,8 @@ use iris_mpc_common::helpers::aws::{
 use iris_mpc_common::helpers::key_pair::SharesEncryptionKeyPairs;
 use iris_mpc_common::helpers::shutdown_handler::ShutdownHandler;
 use iris_mpc_common::helpers::smpc_request::{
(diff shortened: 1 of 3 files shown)


iris-mpc/src/services/processors/batch.rs
```diff
@@ -93,33 +92,13 @@ pub async fn receive_batch(
...
-                            .map_err(ReceiveRequestError::FailedToDeleteFromSQS)?;
-                        if let Some(batch_size) = circuit_breaker_request.batch_size {
-                            // Updating the batch size to ensure we process the messages in the next
-                            // loop
-                            *CURRENT_BATCH_SIZE.lock().unwrap() =
-                                batch_size.clamp(1, max_batch_size);
-                            tracing::info!(
-                                "Updating batch size to {} due to circuit breaker message",
-                                batch_size
+                    IDENTITY_DELETION_MESSAGE_TYPE => {
```
danielle-tfh, line 95: In this case, our deletion queues will probs fill up very heavily.

Wonder if we should delete but not add?

carlomazzaferro: Sounds fair to me 👍

danielle-tfh (approved):

PR #1158, opened 2025-03-15
aurel/grpc-pool: GRPC connection pool

naure:
AFAICT the `PartyNodeClient` does not use any sort of connection pool. It is only one connection so the whole networking pipeline is single-threaded. Not sure what the performance impact is if any though.

This is an attempt at using a connection pool using an existing facility in `tonic`. It is just a different way to create the same kind of `Channel` than the default `PartyNodeClient::connect` does. I would expect then sessions to be distributed over the connections.

The `test_hawk_main` is stuck atm. Maybe it would be simpler to create our own fixed pool of `Vec<PartyNodeClient>` instead.

### DIFF
--- a/.github/workflows/build-all-targets.yml
+++ b/.github/workflows/build-all-targets.yml
@@ -3,9 +3,9 @@ name: Build All Targets
 on:
   push:
     paths-ignore:
-      - 'adr/**'
-      - 'deploy/**'
-      - '.github/**'
+      - "adr/**"
+      - "deploy/**"
+      - ".github/**"
 
 concurrency:
   group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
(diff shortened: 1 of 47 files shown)


iliailia (commented):
This makes sense to me. It'd be great to do some tests and run benchmarks (`cargo bench hnsw`) before merging.

dkales:
Just to add some of our insights, we also had a single client being a bottleneck in another project, and there we added a Vec<Client> based pool, since we knew the maximum number of parallel sessions we would need and could easily do this.
I would be interested if the pooling that is builtin with balance manages to achieve the same performance.

naure:
Now simplified using `Vec<Client>`.

iliailia:
Thanks! Could you run `cargo bench hnsw` to see whether it affected the local networking performance?

naure:
> Thanks! Could you run `cargo bench hnsw` to see whether it affected the local networking performance?

This benchmark does not have this sort of parallelism.

iliailia:
> > Thanks! Could you run `cargo bench hnsw` to see whether it affected the local networking performance?
> 
> This benchmark does not have this sort of parallelism.

Right, sorry for bothering. Approved!

iliailia (approved):

naure:
This sounds tricky to benchmark accurately but we will try at some point in order to tune the parallelism parameters.

PR #1159, opened 2025-03-17
Implement parallel MPC execution for HNSW graph insertion

bgillesp:
Adds support for parallel execution of binary insertion of a new node into the neighbors' neighbor lists.

A new `BinarySearch` struct is implemented which keeps track of ongoing state of a binary search procedure and provides the next array index which needs to be compared against the query.  This struct is used in `HnswSearcher::insert_prepare` to execute binary insertion in parallel using one call to `less_than_batch` per binary search step, or about 10 calls in total at realistic parameter sizes.


siajasl (approved):
Nice looking code.  Perfect level of comments.

iliailia:
Oh, my bad. I think I really misunderstood what you've done here. I see from experiments that rounds got reduced by a factor of 350 and bandwidth by a factor of 30. Amazing!

iliailia (approved):

bgillesp:
> I see from experiments that insertion rounds got reduced by a factor of 350 and bandwidth by a factor of 30.

Yes, the insertion procedure is logically very shallow because all of the "neighbors' neighborhoods" can be updated independently with the new node, so total depth is just the maximum of binary search over the neighborhood sizes, or about 10 less-than ops.  Definitely some good low-hanging fruit to gather.

PR #1162, opened 2025-03-18
Make Session unclonable

iliailia:
This is the first step to make sessions able to store message queues coming from gRPC interactions. This also makes explicit that sessions should be operating in a single thread.


iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -36,6 +36,15 @@ pub struct GraphMem<V: VectorStore> {
...
 
+impl<V: VectorStore> Clone for GraphMem<V> {
+    fn clone(&self) -> Self {
+        GraphMem {
+            entry_point: self.entry_point.clone(),
+            layers: self.layers.clone(),
+        }
+    }
+}
+
```
naure, line 47: Good find. That's a limitation of derive(Clone).

naure (approved):
Nice extra type safety.

PR #1167, opened 2025-03-19
Aurel/intra batch

naure:
Match the requests with each other within a batch.

In this PR I am trying a new style:
- First a schedule of tasks is generated. This deals with the logistics of rotations/sessions/etc. It is a single function `schedule` that produces pure numbers so it stays fairly readable and testable.
    - Although there’s still some logistics in `per_session` that I could move to `schedule`.
- Meanwhile, the high-level function `intra_batch_is_match` deals exclusively with `tokio / Arc / Mutex` boilerplate.

I could potentially port the other similar algorithms (`search_both_eyes` and `calculate_missing_is_match`) to this `schedule` style.


iris-mpc-cpu/src/execution/hawk_main/intra_batch.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    let mut match_lists = vec![Vec::new(); n_requests];
+
+    // Find pairs with left AND right match.
+    for ((i_request, j_request), [left, right]) in join {
+        if left && right {
+            // The pair of matching requests track each other's index.
+            match_lists[i_request].push(j_request);
+            match_lists[j_request].push(i_request);
```
dkales, line 121: If i recall correctly, this is not the current behavior in the batch matching in GPU. I think atm if we have x in the batch twice, [x, a,b,c,d,e,f,x], then we report a batch match for the second x but not for the first one, since it is fine for the first one to be unique in that case. This is the case, right @philsippl?

naure: That makes sense. See the latest changes around `earlier_request` and around `HawkResult::new`.

naure: What do we do if there is a chain of matches though? Say request `J` is a duplicate of request `I`, but `I` will also not be inserted (for any reason), then should `J` be inserted 🙃 ?

naure:
@philsippl Can you explain what into the different fields of `ServerJobResult` in case there is a duplicate in a batch? Does it count as a "match" in these fields:

```
merged_results
matches
match_ids
```

naure:
Now it also returns the serial IDs of the intra-batch matches.

Say request `J` is a duplicate of request `I`. Normally request `I` will be inserted and it will have a new serial ID. We return this fresh ID of `I` in the `match_ids` and `merged_results` of `J`.

Note that there is still the edge case explained above. But at this point I need to refactor the `HawkResult` logic which I’d rather do in another PR.

iris-mpc-cpu/src/execution/hawk_main/intra_batch.rs
dkales: I guess atm this is handled by the assumption that if J is not inserted due to a reason then I will also not be inserted due to the same reason. (I guess in theory one could craft iris codes such that J matches DB, I does not, but J matches I. This is probably an edge case and the default behavior of not inserting I in this case is also fine?)

naure: Yes that sounds sensible.

iliailia (approved):
LGTM

iris-mpc-cpu/src/execution/hawk_main/intra_batch.rs
philsippl: yes, it's as @dkales describes.

PR #1168, opened 2025-03-19
Support mirroring of iris codes

philsippl:
Preparation for Mirror PAD.


iris-mpc-common/src/galois_engine.rs
```diff
@@ -364,7 +414,7 @@ pub mod degree4 {
         let store_iris_shares = code_share.clone();
         let store_mask_shares = mask_share.clone();
 
-        // With rotations for in-memory database.
+        // With rotations for in-memory database. // TODO: is this dead code?
```
dkales, line 417: It is needed for the batch cross-check IIRC.

dkales (approved):

iris-mpc-common/src/galois_engine.rs
philsippl: ah right!

philsippl:
@leonanos8 you can use the function for masks, just do it before bringing them into the "trimmed" form (so the query).

PR #1169, opened 2025-03-19
Experiments to reduce GPU computation work

dkales:
Based on an idea by Philipp, we can only match a subset of the DB based on the results of the other eye. This PR is work in progress to implement this.

### DIFF
--- a/iris-mpc-gpu/src/dot/distance_comparator.rs
+++ b/iris-mpc-gpu/src/dot/distance_comparator.rs
@@ -1,8 +1,8 @@
 use super::ROTATIONS;
 use crate::{
     helpers::{
-        device_manager::DeviceManager, launch_config_from_elements_and_threads,
-        DEFAULT_LAUNCH_CONFIG_THREADS,
+        device_manager::DeviceManager, htod_on_stream_sync,
+        launch_config_from_elements_and_threads, DEFAULT_LAUNCH_CONFIG_THREADS,
     },
     threshold_ring::protocol::{ChunkShare, ChunkShareView},
 };
@@ -13,6 +13,7 @@ use cudarc::{
     },
     nvrtc::compile_ptx,
 };
+use rand::seq::index;
 use std::{cmp::min, sync::Arc};
 
 const PTX_SRC: &str = include_str!("kernel.cu");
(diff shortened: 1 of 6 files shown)


dkales:
This is working now, tested also with 2 GPUs per party on runpod.
A few notes:
* If the number of partial matches + Or rule indices is larger than the max_chunk_size, we fall back to doing  full match for now.
We could instead just to multiple partial matches, should also not be too bad to implement.
* The copying of the subset of db indices to the GPU now syncs for each device, this could probably be improved, but our htod_on_stream method syncs after each copy due to it taking a slice and needing that for general safety.
* The tests previously checked the returned partial results, this is no longer possible due to the right side not producing them fully anymore. Should the partial results of the right half be returned at all?
* The bucket stats for the right side now get collected much less often, since it only collects the matches in the subset of the DB. The generic distribution should probably still be fine, but one should maybe investigate the effects of only doing stats on the pre-filtered DB...

iris-mpc-gpu/src/dot/distance_comparator.rs
```diff
@@ -324,6 +406,101 @@ impl DistanceComparator {
...
+                .collect::<Vec<_>>();
+            ids.sort();
+            matches.push(ids);
+        }
+
+        tracing::info!(
+            "Unique Partial matches len: {:?}",
+            matches.iter().map(|m| m.len()).collect_vec()
+        );
+        tracing::info!("Unique Partial matches: {:?}", matches);
```
philsippl, line 499: probably want to remove this, could be large?

philsippl:
@worldcoin/proof-of-personhood could someone do a staging e2e test of this branch when you have a chance?

leonanos8:
> @worldcoin/proof-of-personhood could someone do a staging e2e test of this branch when you have a chance?

@philsippl do you think this is a pre-requisite for this test to happen? ->https://linear.app/worldcoin/issue/INFRA-4277/introduce-p4-instances-into-iris-mpc-gpu-in-staging

dkales:
@philsippl pls take a quick look at the last commit, I think this needs to be this way in case there are too many matches collected, because we ATM cannot distinguish between exactly CHUNK_LEN matches and more than CHUNK_LEN matches since it gets truncated. This check removes the ability to do the partial one if there are exactly CHUNK_LEN matches, but at least will not produce any wrong results.

dkales:
Also any comments on my points above?

philsippl:
> @philsippl pls take a quick look at the last commit, I think this needs to be this way in case there are too many matches collected, because we ATM cannot distinguish between exactly CHUNK_LEN matches and more than CHUNK_LEN matches since it gets truncated. This check removes the ability to do the partial one if there are exactly CHUNK_LEN matches, but at least will not produce any wrong results.

good catch, this seems fine for now until we actually iterate over multiple chunks. Alternatively we could also return the actual counter from `get_partial_results`, but this seems not worth it.

philsippl:
> This is working now, tested also with 2 GPUs per party on runpod. A few notes:
> 
> * If the number of partial matches + Or rule indices is larger than the max_chunk_size, we fall back to doing  full match for now.
>   We could instead just to multiple partial matches, should also not be too bad to implement.
> * The copying of the subset of db indices to the GPU now syncs for each device, this could probably be improved, but our htod_on_stream method syncs after each copy due to it taking a slice and needing that for general safety.
> * The tests previously checked the returned partial results, this is no longer possible due to the right side not producing them fully anymore. Should the partial results of the right half be returned at all?
> * The bucket stats for the right side now get collected much less often, since it only collects the matches in the subset of the DB. The generic distribution should probably still be fine, but one should maybe investigate the effects of only doing stats on the pre-filtered DB...

re those points:
- I think we should iterate over it in multiple chunks, but let's do that as a follow-up (we also need to look into streaming)
- htod_on_stream makes it currently impossible to misuse, but we could have a more "unsafe" version which makes the caller ensure that the slice is valid and we instead sync after starting the `open_index_mapping_kernels` kernel on all devs? On the otherhand the runtime of those kernels should be tiny compared to everything else.
- I would leave them in the reply, in case we want to ever implement something like switching which side we do for the full lookup vs. the partial one (for the sake of bucket statistics)
- Yeah my only idea would be doing something like the above in case we really need / want it, but would just leave it as is.

philsippl:
> > @worldcoin/proof-of-personhood could someone do a staging e2e test of this branch when you have a chance?
> 
> @philsippl do you think this is a pre-requisite for this test to happen? ->https://linear.app/worldcoin/issue/INFRA-4277/introduce-p4-instances-into-iris-mpc-gpu-in-staging

@leonanos8 no, this can go in parallel, not a blocker.

leonanos8:
> @leonanos8 no, this can go in parallel, not a blocker.

@philsippl Ok ok, I will perform some end to end tests in stage with the current stage infra and will let you know

leonanos8:
@philsippl @dkales proceeded with the following tests:

- `stage env`
Deployed an image that contains this change in stage env. After the new image got deployed in all the iris-mpc stage clusters, I performed several simulated enrollments using `orb-tools`.
All were successfull

- `stage e2e isolated environment`
Used our testing framework for e2e and ran the tests with ~200 enrollments. Execution was successfull: https://github.com/worldcoin/orb-context-environment/actions/runs/14082025464/job/39436989423

It seems that the change is working as expected.

@philsippl do you have any other scenarios in mind to test this? Maybe perform orb verification with some test orb in the office that points to stage mutliple times?

dkales:
@leonanos8 Thank you for the staging tests! I think what would be nice if there are some performance numbers that we can get from the staging deployment. Theoretically this should reduce the runtime of a batchquery by about 50%, depending on the database size. IDK if this is easily possible (comparing the staging performance to a deployment from main).

leonanos8:
> @leonanos8 Thank you for the staging tests! I think what would be nice if there are some performance numbers that we can get from the staging deployment. Theoretically this should reduce the runtime of a batchquery by about 50%, depending on the database size. IDK if this is easily possible (comparing the staging performance to a deployment from main). Also the effects would only show up with a pretty large database (> 32k at least, sizes around 1M would be great)

@dkales thanks for your suggestion. As discussed internally we could try and perform the following test:
- Populate `iris-mpc` stage databases with ~1m records of iris sample/random data
- Run a batch of enrollments with nodes running code from main (not containing the improvement)
- Run another batch of enrollemts with nodes running code that the improvement
Check the batch time for both cases and validate the ~50% improvement in batch processing time

I will push so we can have these results.
I will post results in this Linear Ticket: https://linear.app/worldcoin/issue/POP-2364/run-batch-time-performance-tests-in-stage

philsippl:
@worldcoin/proof-of-personhood what's the plan to release this?

PR #1181, opened 2025-03-21
Add version id to iris-mpc store

danielle-tfh:
### Notes
* add versioning to irises
* the trigger checks if the values actual values change - so for re-running modifications the version would not update
* we introduce a new type of iris db version - this way GPU and CPU can use what they need (it also stops GPU from streaming the version into memory)
* remove unused methods `insert_or_update_left_iris` and `insert_or_update_right_iris`


eaypek-tfh (approved):
LGTM! 🚀

naure (approved):

PR #1187, opened 2025-03-23
[POP-2328] Enrich BatchQuery with mirrored shares and masks for mirror attack detection

leonanos8:
### Description
This PR is the first step of full-face mirror attack detection. We are enriching the `BatchQuery` struct to also include the vertically mirrored iris shares and masks.

More Feature context in this design doc: https://www.notion.so/worldcoin/Solution-Design-Defending-against-full-face-mirror-attack-in-the-backend-side-1b98614bdf8c807e9e4bd5ee949bfb4c#1ba8614bdf8c805f89e7ea23218f7b62

Follow-up PR will use the mirrored shares and masks in the actor.rs code to perform computation for both normal and mirrored case

### Implementation Details:
- We only create and keep in memory the 31 rotations of the Langrange interpolated mirrored shares and masks, to be used by the actor while comparing for uniqueness
- Notes around the refactor of `decode_iris_message_shares()`:
Before this PR, the above function was also hiding the trimming of the mask. This is breaking the `Single Responsibility Principle`. When reading the code it was not obvious that the trimming was part of the decode process. In this PR, I have to mirror the masks before trimming them (ref: https://github.com/worldcoin/iris-mpc/pull/1168#issuecomment-2743774548), so I decided to refactor this function to stick to just decoding the data and have an extra function that does the trimming.
Benefits of This Approach:
      - Improved Readability: The function now clearly does only one thing - decoding base64 shares
      - Better Separation of Concerns: Decoding and trimming are now separate operations
      - More Flexibility: Having access to the untrimmed mask allows for other operations before trimming (like the mirroring that we have to do in this case)
      - Self-Documenting: The code now makes it explicit that trimming is happening

### DIFF
--- a/iris-mpc/bin/server.rs
+++ b/iris-mpc/bin/server.rs
@@ -76,26 +76,32 @@ type GaloisShares = (
     Vec<GaloisRingTrimmedMaskCodeShare>,
     Vec<GaloisRingIrisCodeShare>,
     Vec<GaloisRingTrimmedMaskCodeShare>,
+    Vec<GaloisRingIrisCodeShare>,
+    Vec<GaloisRingTrimmedMaskCodeShare>,
 );
 type ParseSharesTaskResult = Result<(GaloisShares, GaloisShares), Report>;
 
 fn decode_iris_message_shares(
     code_share: String,
     mask_share: String,
-) -> eyre::Result<(GaloisRingIrisCodeShare, GaloisRingTrimmedMaskCodeShare)> {
+) -> eyre::Result<(GaloisRingIrisCodeShare, GaloisRingIrisCodeShare)> {
     let iris_share = GaloisRingIrisCodeShare::from_base64(&code_share)
         .context("Failed to base64 parse iris code")?;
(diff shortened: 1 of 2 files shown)


iris-mpc/bin/server.rs
```diff
@@ -76,26 +76,32 @@ type GaloisShares = (
...
+    let mask_share = GaloisRingIrisCodeShare::from_base64(&mask_share)
+        .context("Failed to base64 parse iris mask")?;
 
     Ok((iris_share, mask_share))
 }
 
+fn trim_mask(mask: GaloisRingIrisCodeShare) -> GaloisRingTrimmedMaskCodeShare {
+    mask.into()
+}
+
```
philsippl, line 99: you like this better than the `.into()` in place?

philsippl (approved):

iris-mpc/bin/server.rs
leonanos8: I wrote it like this to make it obvious from the first read that we are actually trimming the mask. I believe that it makes the `get_iris_parse_task()` function a little bit more self-documented

PR #1193, opened 2025-03-24
[POP-2332] Implement compare_multiple_threshold function to compute anonymized statistics

dkales:
This adds a function to aggregate matched distance results that are collected beforehand into buckets based on parameterized threshold values.
In another followup PR this will be integrated together with collection of the match pairs + execution of this function once enough matches are collected + output of the result


iris-mpc-cpu/src/protocol/binary.rs
```diff
@@ -413,6 +413,138 @@ pub(crate) async fn bit_inject_ot_2round(
     Ok(res)
 }
 
+/// This function splits a bit shared additively over F_2 into three shares that need to be combined using the arithmetic xor function
+/// This is a completely local function, `session` is only used to determine the current party id.
+pub(crate) fn split_for_arithmetic_xor(
```
iliailia, line 418: You might want to reuse `a2b_pre` here.

iris-mpc-cpu/src/protocol/binary.rs
```diff
@@ -413,6 +413,138 @@ pub(crate) async fn bit_inject_ot_2round(
...
+        input1.into_iter(),
+        input2.into_iter()
+    )
+    .map(|(a, b, inp1, inp2)| inp1 + inp2 - Share::new(a, b) * RingElement(2u32))
+    .collect_vec();
+    Ok(VecShare::new_vec(res))
+}
+
+/// Bit injects a vector of bits
+pub(crate) async fn bit_inject_arithmetic_xor(
```
iliailia, line 537: Why not to use `bit_inject_ot_2round`?

iris-mpc-cpu/src/protocol/binary.rs
dkales: I saw that but that: 1) would do the party matching for each share instead of once, and 2) this converts bit shares directly to u32 shares instead.

iris-mpc-cpu/src/protocol/binary.rs
dkales: Yeah mostly the types. I also opted for the much simpler arithmetic xor version of bit inject here, since the performance of this is not very critical. Arithmetic Xor bit inject has 2 RE of communication per party vs the 4/3 RE per party (unbalanced) for the 3-OT variant

iliailia: Check this out https://github.com/worldcoin/iris-mpc/pull/1202/commits/7247278b4fefde6fd0ad1961fedf473fde3fcc47

dkales:
@iliailia Have now used your methods, pls take a look. (Putting generic ring open in the binary mod feels a bit wrong, should this be somewhere else)

iris-mpc-cpu/src/protocol/binary.rs
```diff
@@ -814,3 +814,30 @@ pub async fn open_bin(session: &mut Session, shares: &[Share<Bit>]) -> eyre::Res
         .map(|(s, c)| Ok((s.a ^ s.b ^ c).convert()))
         .collect::<eyre::Result<Vec<_>>>()
 }
+
+#[instrument(level = "trace", target = "searcher::network", skip_all)]
+pub async fn open_ring<T: IntRing2k + NetworkInt>(
```
iliailia, line 819: You might want to move this to `ops.rs`

iliailia (approved):
Nice, thank you!

PR #1198, opened 2025-03-26
chore(POP-2297): Allow using multiple databases in different deployment modes

wojciechsromek:
Work done:
- Added support for separate CPU and GPU databases
- Refactored database connection logic into a new PostgresClient struct
- Modified transaction handling to support separate database connections
- Moved database migrations to a common location (root of the main workspace)


iris-mpc-common/src/postgres/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+        let connect_sql = sql_switch_schema(schema_name)?;
+
+        let pool = PgPoolOptions::new()
+            .max_connections(MAX_CONNECTIONS)
+            .after_connect(move |conn, _meta| {
+                // Switch to the given schema in every connection.
+                let connect_sql = connect_sql.clone();
+                Box::pin(async move {
+                    conn.execute(connect_sql.as_ref()).await.inspect_err(|e| {
+                        eprintln!("error in after_connect: {:?}", e);
```
eaypek-tfh, line 43:
```suggestion
                        tracing::error!("error in after_connect: {:?}", e);
```

iris-mpc/bin/server.rs
```diff
@@ -893,8 +894,22 @@ async fn server_main(config: Config) -> eyre::Result<()> {
...
+        config.app_name, config.environment, config.party_id
+    );
+    let db_config = config
+        .database
+        .as_ref()
+        .ok_or(eyre!("Missing database config"))?;
+
+    tracing::info!(
+        "Creating new iris storage from: {:?} with schema {}",
+        db_config,
```
eaypek-tfh, line 908: was worried about this but it seems we have a custom fmt::Debug impl which masks out db url 👍

iris-mpc/bin/server.rs
```diff
@@ -2032,8 +2047,8 @@ async fn process_identity_deletions(
             "Started processing deletion request",
         );
 
-        // overwrite postgres db with dummy values.
-        // note that both serial_id and postgres db are 1-indexed.
+        // overwrite postgres.rs db with dummy values.
+        // note that both serial_id and postgres.rs db are 1-indexed.
```
eaypek-tfh, line 2051: these seem like rename left-overs, let's update all to remove `.rs` 👍

eaypek-tfh (approved):
It looks clean to me! 🚀 

Btw, this deserves a test in stage to see if migrations and transactions are still functioning correctly 👍

iris-mpc/bin/server.rs
wojciechsromek: yes, thank you

iris-mpc-common/src/postgres/mod.rs
dkales, line 1: Can/should this not live in `iris-mpc-store`?

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
```diff
@@ -32,13 +33,23 @@ pub struct GraphPg<V: VectorStore> {
     phantom: PhantomData<V>,
 }
 
+static MIGRATOR: Migrator = sqlx::migrate!("./../migrations");
```
dkales, line 36: Should the migrations not be split up between the Graph and normal store? ATM this applies all migrations to the DB?

iris-mpc/src/server/mod.rs
```diff
@@ -811,3 +810,92 @@ pub async fn server_main(config: Config) -> eyre::Result<()> {
...
+        config.app_name, config.environment, config.party_id
+    );
+
+    // Always need the "GPU" database config
+    let gpu_db_config = config
+        .database
+        .as_ref()
+        .ok_or(eyre!("Missing GPU database config"))?;
+
+    let gpu_postgres_client = PostgresClient::new(&gpu_db_config.url, &schema_name).await?;
```
dkales, line 826: This is a bit annoying for the isolation mode below, since you still need a valid url here to connect to. Maybe we should move this into the respective two branches below instead?

iris-mpc/src/server/mod.rs
```diff
@@ -811,3 +810,92 @@ pub async fn server_main(config: Config) -> eyre::Result<()> {
...
+                "Creating new graph store from: {:?} in mode {:?}",
+                cpu_db_config,
+                config.mode_of_deployment
+            );
+            let graph_store = GraphStore::new(&cpu_postgres_client).await?;
+
+            Ok((store, graph_store))
+        }
+
+        // All other modes: we only need/use the GPU DB
```
dkales, line 882: The name GPU DB is not really representative here, right? Even in the standard deployment we do not have a "GPU DB" here, this would still be the DB for the CPU deployment, no? The GPU binary never calls this method

I think ATM out of the 3 cases only 2 are ever happening:
* SharesDB and GraphDB are the same DB => Isolation mode (DB_URL points to its own DB), Standard Mode (DB_URL points to production DB)
* SharesDB and GraphDB are different DBs => Shadow Mode (DB_URL points to production DB with readonly access, CPU_DB_URL points to its own DB)

iris-mpc/src/server/mod.rs
```diff
@@ -811,3 +810,92 @@ pub async fn server_main(config: Config) -> eyre::Result<()> {
...
+        }
+
+        ModeOfDeployment::ShadowReadOnly => {
+            // Store -> GPU
+            tracing::info!(
+                "Creating new iris store from: {:?} in mode {:?}",
+                gpu_db_config,
+                config.mode_of_deployment
+            );
+            let store = Store::new(&gpu_postgres_client).await?;
```
dkales, line 863: can we somehow programmatically check the rights of the DB user as a safeguard?

iris-mpc-common/src/postgres/mod.rs
wojciechsromek: From my perspective a database client is not a part of the store, but a required dependency. Also, the client is very general, hence I put it into the common crate.

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
wojciechsromek: I did not manage to separate them, as far as I have managed to learn, `sqlx` is not able to keep separate sets of migrations, it assumes that there is only one set of them. This way what will happen is, both DBs are going to be created with the same structure (in `ShadowReadOnly`).

iris-mpc/src/server/mod.rs
wojciechsromek: Correct me if I am wrong, but I believe that we never run the service without this url. I agree that naming can be better, will change.

iris-mpc/src/server/mod.rs
wojciechsromek: We could try to verify if this is read-only, will think about approach.

wojciechsromek:
After conversation with Daniel I will keep this PR in place until consult with Carlo / Danielle.

iris-mpc/src/server/mod.rs
wojciechsromek: I ended up implementing this in a slightly different way, by disabling persistence on the `iris-mpc` db in the ShadowReadOnly. I also provided the environment setup because otherwise it would be impossible to test reliably.

carlomazzaferro (approved):

eaypek-tfh (approved):

PR #1202, opened 2025-03-28
Generic network value creation

iliailia:
This allows to send and receive generic network values, which in turn allows to compute SMPC functions over generic arithmetic types implementing the `IntRing2k` trait.

### DIFF
--- a/iris-mpc-cpu/src/protocol/binary.rs
+++ b/iris-mpc-cpu/src/protocol/binary.rs
@@ -76,57 +76,65 @@ pub(crate) fn transposed_pack_xor<T: IntRing2k>(
     res
 }
 
-pub(crate) async fn and_many_send(
+pub(crate) async fn and_many_send<T: IntRing2k>(
     session: &mut Session,
-    a: SliceShare<'_, u64>,
-    b: SliceShare<'_, u64>,
-) -> Result<Vec<RingElement<u64>>, Error>
+    a: SliceShare<'_, T>,
+    b: SliceShare<'_, T>,
+) -> Result<Vec<RingElement<T>>, Error>
 where
-    Standard: Distribution<u64>,
+    Standard: Distribution<T>,
 {
+    if ![16,32,64].contains(&T::K) {
+        return Err(eyre!("Invalid bit size in and_many_send"));
+    }
     if a.len() != b.len() {
         return Err(eyre!("InvalidSize in and_many_send"));
     }
     let mut shares_a = Vec::with_capacity(a.len());
     for (a_, b_) in a.iter().zip(b.iter()) {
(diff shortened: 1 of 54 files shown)


iris-mpc-cpu/src/protocol/binary.rs
```diff
@@ -702,7 +722,7 @@ pub async fn extract_msb_u32<const K: usize>(
     x_: VecShare<u32>,
 ) -> Result<VecShare<u64>, Error> {
     let x = x_.transpose_pack_u64_with_len::<K>();
-    extract_msb::<K>(session, x).await
+    extract_msb::<u64>(session, x).await
```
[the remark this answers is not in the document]
iliailia: I want to use the generics here to compute on batches of u32 and u16, because we have quite a lot of them in a typical HNSW execution.

iliailia:
> LGTM, left a few comments. Also, the prefixed length encoding for networking is u32, which limits messages to 4GB. Is this enough in general?

Yep, the maximal message we're sending in a typical HNSW execution is about 12 KB. So u16 is actually a better fit.

iris-mpc-cpu/src/protocol/binary.rs
```diff
@@ -305,14 +299,17 @@ where
...
         let m0_and_m1 = NetworkValue::vec_from_network(reply_m0_and_m1).unwrap();
-        assert!(
-            m0_and_m1.len() == 2,
-            "Deserialized vec in bit inject is wrong length"
-        );
-        let (m0, m1) = m0_and_m1.into_iter().collect_tuple().unwrap();
-
-        let m0 = T::into_vec(m0);
-        let m1 = T::into_vec(m1);
+        if m0_and_m1.len() == 2 {
```
dkales: this seems wrong way round at first glance?

iliailia: Yep, I already fixed it.

iliailia:
> > LGTM, left a few comments. Also, the prefixed length encoding for networking is u32, which limits messages to 4GB. Is this enough in general?
> 
> Yep, the maximal message we're sending in a typical HNSW execution is about 12 KB. So u16 is actually a better fit.

Actually, this is not correct due to the recent changes in the HNSW algorithm. The maximal message I saw in experiments is about 500 KB, which is way less than 4 GB, but its serialized length doesn't feet u16.

dkales (approved):

iris-mpc-cpu/src/protocol/binary.rs
dkales: my bad, apparently had review open on outdated version

dkales:
Will then rebase the threshold_many PR on this once merged

PR #1205, opened 2025-03-31
Unify on 1-based serial ID

naure:
The concept of 0-based / 1-based IDs is starting to make things quite complicated (e.g. vector ID https://github.com/worldcoin/iris-mpc/pull/1151).

This PR isolates all the +1 -1 conversions at the actor interface (`InMemoryStore` and `ServerJobResult`).

From there, it will be possible to go further and remove the concept of 0/1 conversions at all in the GPU code.

### DIFF
--- a/iris-mpc-store/src/lib.rs
+++ b/iris-mpc-store/src/lib.rs
@@ -48,10 +48,10 @@ pub enum StoredIris {
 
 impl StoredIris {
     /// Returns the `id` from either variant.
-    pub fn index(&self) -> usize {
+    pub fn serial_id(&self) -> usize {
         match self {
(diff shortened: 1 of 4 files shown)


iris-mpc-store/src/lib.rs
```diff
@@ -67,8 +67,8 @@ pub struct DbStoredIris {
 }
 
 impl DbStoredIris {
-    /// The index which is contiguous and starts from 0.
-    pub fn index(&self) -> usize {
+    /// The index which is contiguous and starts from 1.
+    pub fn serial_id(&self) -> usize {
```
naure, line 71: _This is only a rename - this was 1-based all along._

PR #1207, opened 2025-04-01
Update batch size from queue size

carlomazzaferro:


.test.env
```diff
@@ -7,7 +7,7 @@ SMPC__MODE_OF_DEPLOYMENT=STANDARD
 SMPC__DATABASE__MIGRATE=true
 SMPC__DATABASE__CREATE=true
 SMPC__DATABASE__LOAD_PARALLELISM=8
-SMPC__MAX_BATCH_SIZE=64
+SMPC__MAX_BATCH_SIZE=1
```
carlomazzaferro, line 10: Changing this to enable local testing with the e2e tests

eaypek-tfh (commented):
In general looks good to me 🙌 

My worry is that every time `batch-sync-state` API is called, we are polling from SQS to see the first message. This could cause some messages to be put into DLQ after some attempts. To mitigate, maybe we can store the last sns_sequence_number received while polling and return it as part of sync with the other nodes.

The other thing is it seems we are only waiting for 10 seconds and if no messages are received, then we try to get a single msg. Wdyt about polling for indefinitely until we either have no new messages for `batch_polling_timeout_secs` or we filled the batch already?

Also side note, we can use `wait_time_seconds` on sqs `receive_message` call instead of handling tokio cases.

Would be interested in @dkales' thoughts

iris-mpc-common/src/helpers/sqs.rs
```diff
@@ -43,7 +43,7 @@ pub async fn get_next_sns_seq_num(
         }
     } else {
         tracing::info!("Timeout while waiting for a message in the queue. Queue is clean.");
-        Ok(None)
+        Ok(Some(0))
```
dkales, line 46: a bit concerned if 0 is the best value to signal this, can this not be 0 on new queues at least?

iris-mpc/src/services/processors/batch.rs
```diff
@@ -137,13 +139,110 @@ impl<'a> BatchProcessor<'a> {
...
 
-        // Poll until we have enough messages
-        // temporary hack for staging to only process 1 message at a time
-        // this helps with the correctness test
-        while self.msg_counter < 1 {
+        // Try to fill the batch with a timeout
+        match tokio::time::timeout(
+            polling_timeout,
+            self.poll_until_batch_size(max_batch_size, queue_url),
+        )
```
dkales, line 151: I think this could also lead to some edge cases where the get_message call is cancelled due to this timeout, and messages are already in flight. This means that subsequent calls to get_message actually block/return nothing, since the in-flight message is not deleted and we have to wait for its visibility timeout to elapse before the system is ready again.
We encountered this in document-mpc

iris-mpc-common/src/helpers/sqs.rs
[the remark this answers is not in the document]
carlomazzaferro: Good call, fixed the comment. It is indeed disabled

carlomazzaferro:
> In general looks good to me 🙌
> 
> My worry is that every time `batch-sync-state` API is called, we are polling from SQS to see the first message. This could cause some messages to be put into DLQ after some attempts. To mitigate, maybe we can store the last sns_sequence_number received while polling and return it as part of sync with the other nodes.
> 

Addressed this by storing the sequence number locally every time we poll a new message from SQS, good call!

> The other thing is it seems we are only waiting for 10 seconds and if no messages are received, then we try to get a single msg. Wdyt about polling for indefinitely until we either have no new messages for `batch_polling_timeout_secs` or we filled the batch already?
> 

addressed this as well, removed the call to poll for the 1 message

> Also side note, we can use `wait_time_seconds` on sqs `receive_message` call instead of handling tokio cases.

Addressed this!

carlomazzaferro:
> If using wait_time_seconds, this will also need the additional timeout config for the SQS client. (see recent PR in document-mpc)

will incorporate

carlomazzaferro:
> If using wait_time_seconds, this will also need the additional timeout config for the SQS client. (see recent PR in document-mpc)

I've added this: https://github.com/worldcoin/iris-mpc/pull/1207/files#diff-40791a8ed45d85e98a0fd4171631f22916dc8ffd3f13a24d28cda9e4bea58d88R6-R23

iris-mpc-common/src/helpers/batch_sync.rs
```diff
@@ -0,0 +1,87 @@
...
+    sqs_client: &aws_sdk_sqs::Client,
+) -> Result<BatchSyncState> {
+    let next_sns_sequence_num = match get_next_sns_seq_num(config, &sqs_client.clone()).await? {
+        Some(seq) => {
+            println!("fetching next_sns_sequence_num: {}", seq);
+            seq
+        }
+        None => {
+            println!("fetching next_sns_sequence_num: None (queue was empty)");
+            0
```
eaypek-tfh, line 45: shall we remove these? probably debug logs

eaypek-tfh (approved):

PR #1210, opened 2025-04-01
Implement binary to initialize test databases from plaintext

bgillesp:
Implements a new binary to:
- [x] Read serialized synthetic iris codes (plaintext) into memory
- [x] Construct HNSW graphs of plaintext iris codes in memory (per iris side)
- [x] Persist the graphs to Postgres at various checkpoints
- [x] Locally generate Galois ring secret shares for plaintext iris codes
- [x] Persist the secret shared iris codes to Postgres

The binary is designed to be run in several iterations to support checkpointing of databases at various stages of generation.  To accomplish this, the program checks for existing iris code entries in the `irises` table of Party 0, and if any are detected it loads the HNSW graph from Party 0's database, and loads corresponding plaintext iris codes from an initial part of the input NDJSON file.

To support consistent graph construction from a fixed seed, the state for the HNSW graph construction PRNGs is saved to a file after the binary finishes executing, and is loaded from file if any existing irises are found in the database on startup.  To accomplish this the binary uses `ChaCha8Rng` structs for the HNSW graph construction, since the `AesRng` structs previously used don't support serialization.  To compare these constructions deterministically to testing of the graph genesis binary, we will need to ensure that that binary is using the same RNG construction.

Graph construction runs in single separate threads per iris side.


iris-mpc-cpu/bin/init_test_dbs.rs
```diff
@@ -0,0 +1,139 @@
...
+    let ef_search = args.ef_search;
+    let layer_probability = args.layer_probability;
+
+    let mut params = HnswParams::new(ef_constr, ef_search, M);
+    if let Some(q) = layer_probability {
+        params.layer_probability = q
+    }
+    let searcher = HnswSearcher { params };
+
+    let mut rng = AesRng::seed_from_u64(42_u64);
```
iliailia, line 62: How about to have an argument whether the seed is random or fixed?

iris-mpc-cpu/bin/init_test_dbs.rs
```diff
@@ -0,0 +1,495 @@
...
+
+    /// Extends iris shares table by inserting irises following the current
+    /// maximum serial id.
+    async fn persist_vector_shares(
+        &self,
+        shares: Vec<(GaloisRingSharedIris, GaloisRingSharedIris)>,
+    ) -> Result<(), Box<dyn Error>> {
+        let mut tx = self.store.tx().await?;
+
+        let last_serial_id = self.store.get_max_serial_id().await.unwrap_or(0);
```
dkales, line 450: Here it might help to have an assert against the expected serial id of this batch, since they kind of implicitly got one assigned from the plainstore/graph? I think the current setup works out such that this is as expected, but it should not hurt to have this here to uphold this invariant?

bgillesp: Yes, the behavior here definitely relies on the fact that everything is being processed in a structured/sequential way.  I updated this function call to return a tuple giving the range of serial ids assigned to the shares, which are then double-checked against the expected values in `main`.

PR #1213, opened 2025-04-02
On-demand distances

naure:
This PR removes the need for storing distances in edges. This saves a lot of memory and database space.

- [x] API for batched distance evaluations.
- [x] Compute neighbor distances during insertion search.
- [x] Optimize batched distance evaluations.
- [x] API `get_links` without distances.
- [x] Remove distances from everywhere.

### DIFF
--- a/iris-mpc-cpu/src/hnsw/graph/neighborhood.rs
+++ b/iris-mpc-cpu/src/hnsw/graph/neighborhood.rs
@@ -10,13 +10,43 @@ use crate::hnsw::{
     },
     VectorStore,
 };
-use itertools::Itertools;
 use serde::{Deserialize, Serialize};
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
 use tracing::{debug, instrument};
 
 /// A sorted list of edge IDs (without distances).
-pub type SortedEdgeIds<Vector> = SortedNeighborhood<Vector, ()>;
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct SortedEdgeIds<V>(pub Vec<V>);
+
+impl<V> SortedEdgeIds<V> {
+    pub fn from_ascending_vec(edges: Vec<V>) -> Self {
+        SortedEdgeIds(edges)
+    }
+
+    pub fn trim_to_k_nearest(&mut self, k: usize) {
+        self.0.truncate(k);
+    }
+}
+
+impl<V> Default for SortedEdgeIds<V> {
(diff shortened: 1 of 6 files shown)


iris-mpc-cpu/src/hnsw/graph/neighborhood.rs
```diff
@@ -10,10 +10,14 @@ use crate::hnsw::{
     },
     VectorStore,
 };
+use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use std::ops::Deref;
 use tracing::{debug, instrument};
 
+/// A sorted list of edge IDs (without distances).
+pub type SortedEdgeIds<Vector> = SortedNeighborhood<Vector, ()>;
```
bgillesp, line 19: This is a clever solution, but I'm wondering if just using a type alias of `Vec<Vector>` would better represent the available functionality?  Most of the non-trivial code for `SortedNeighborhood` relies on comparisons using cached distances, so pretty much everything else in the `impl` looks like it's just a wrapper around `Vec` function calls.

iris-mpc-cpu/src/hnsw/graph/neighborhood.rs
```diff
@@ -96,6 +100,16 @@ impl<Vector: Clone, Distance: Clone> SortedNeighborhood<Vector, Distance> {
         }
     }
 
+    pub fn edge_ids(&self) -> SortedEdgeIds<Vector> {
```
bgillesp, line 103: This call is identical to the next function `vectors_cloned` except the result is wrapped in a `SortedEdgeIds`.  Either use `vectors_cloned` in this call, or if you do change to a pure `Vec` definition for `SortedEdgeIds` then just get rid of it.

bgillesp (approved):
Great update, and really great observation on the relative lack of utility of storing cached distances with the graph.  Only minor nit is with the type choice for `SortedEdgeIds` -- see comments below.  In the current implementation this should save about 2/3 of the graph memory usage if I'm not mistaken, so really awesome improvement!

iris-mpc-cpu/src/hnsw/graph/neighborhood.rs
naure: I turned it into a Vec wrapper.

iris-mpc-cpu/src/hnsw/graph/neighborhood.rs
naure: Done.

PR #1214, opened 2025-04-02
refactor batch processor into something more sensible

carlomazzaferro:
Preparation for the work needed to poll sqs messages without using the current mechanics of `CURRENT_BATCH_SIZE` + batch size coming `UniquenessRequest`.

Next step is implementing the mechanic by which we take batches as follows:

- Poll from SQS until a timeout is reached
- Once timeout is reached, communicate to other parties your latest sequence number
- If your sequence number < max(other parties sequence numbers) continue polling messages until the max is reached
- Process batch

WIP for this is happening here: https://github.com/worldcoin/iris-mpc/pull/1207


wojciechsromek (approved):
It improves readability a lot, nice one

dkales (approved):
also LGTM

PR #1217, opened 2025-04-03
Sync on common config elements during startup

dkales:
This PR computes a hash of the common config elements and syncs on it during startup to detect basic config mismatches.

TODOs:
- [ ] Agree on all parts of config which are part of sync
- [x] ATM just a basic hash is sent around, should we just send around a common config struct to print better errors?


iris-mpc-common/src/config/mod.rs
```diff
@@ -211,6 +212,121 @@ pub struct Config {
...
+    pub fn common_config_hash(&self) -> String {
+        let mut hasher = sha2::Sha256::default();
+
+        // This is destructured here intentionally to cause a compile error if
+        // any of the fields are added to the struct without being considered if they should be in the common config hash or not.
+        let Config {
+            environment,
+            party_id: _,           // party id is different for each server
+            requests_queue_url: _, // requests queue url is different for each server
+            results_topic_arn,     // TODO: do we want the results topic ARN in the hash?
```
carlomazzaferro, line 225: Would say yes

iris-mpc-common/src/config/mod.rs
```diff
@@ -211,6 +212,121 @@ pub struct Config {
...
+            kms_key_arns: _,       // kms key arns are different for each server
+            service: _,
+            database: _,     // database is different for each server
+            cpu_database: _, // cpu database is different for each server
+            aws: _,          // aws is different for each server
+            processing_timeout_secs,
+            startup_sync_timeout_secs,
+            public_key_base_url: _, // TODO: include in common config hash?
```
carlomazzaferro, line 233: also yes

iris-mpc-common/src/config/mod.rs
```diff
@@ -211,6 +212,121 @@ pub struct Config {
...
+            shares_bucket_name: _,  // TODO: should this be the same for all servers?
```
carlomazzaferro, line 234: yes, same for all

iris-mpc-common/src/config/mod.rs
```diff
@@ -211,6 +212,121 @@ pub struct Config {
...
+            init_db_size,
+            max_db_size,
+            max_batch_size,
+            heartbeat_interval_secs,
+            heartbeat_initial_retries,
+            fake_db_size,
+            return_partial_results,
+            disable_persistence,
+            enable_debug_timing: _,
+            node_hostnames, // TODO: is this the same for all servers?
```
carlomazzaferro, line 245: yes, all the same

iris-mpc-common/src/config/mod.rs
```diff
@@ -211,6 +212,121 @@ pub struct Config {
...
+            luc_lookback_records,
+            luc_serial_ids_from_smpc_request,
+            match_distances_buffer_size,
+            match_distances_buffer_size_extra_percent,
+            n_buckets,
+            enable_sending_anonymized_stats_message,
+            enable_reauth,
+            hawk_request_parallelism,
+            hawk_connection_parallelism,
+            hawk_server_healthcheck_port: _, // different for each server
```
carlomazzaferro, line 269: Only locally, but I get your point

iris-mpc-common/src/config/mod.rs
carlomazzaferro: Except... When running locally. Let's leave it out for now

carlomazzaferro: See: https://github.com/worldcoin/iris-mpc/blob/36b5b4732ab93b767de254eb2b5b2858ca1f0035/.test.hawk0.env#L3

We'd have to change this, otherwise locally this will fail. Let's leave it out for now.

naure: Yes, this 0.0.0.0 is not necessary.

dkales:
Cleaned this up a bit, and removed some things from the common sync that might have a good reason to be different (ports & urls due to proxies/LB/etc).

BTW the CPU integration tests cut of the logs for the docker-compose up step, so the reason for the error (the new startup sync failed) was not apparent in the previous runs. IDK if this can be fixed or if we would need to not use the action.

naure:
It seems easier and more useful to send the whole config indeed.

eaypek-tfh (approved):

PR #1219, opened 2025-04-03
Handle reset check requests

eaypek-tfh:
## Change
- This PR handles `reset_check` messages which are the first communication round to complete a Reset action.
- When received, we conduct a 1-to-N query using AND rule. Note that we skip the OR rule on last N serial_ids too. 
- We skip writing anything into memory or database.

## Side Changes
- Refactor `ExpectedResult` creation in tests to use a builder pattern.
- Move reauth modification logic within feature flag check to not have any action if it's disabled but we still get a request.

### DIFF
--- a/iris-mpc/bin/server.rs
+++ b/iris-mpc/bin/server.rs
@@ -25,12 +25,13 @@ use iris_mpc_common::{
         smpc_request::{
             decrypt_iris_share, get_iris_data_by_party_id, validate_iris_share,
             CircuitBreakerRequest, IdentityDeletionRequest, ReAuthRequest, ReceiveRequestError,
-            SQSMessage, UniquenessRequest, ANONYMIZED_STATISTICS_MESSAGE_TYPE,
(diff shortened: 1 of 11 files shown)


iris-mpc/bin/server.rs
```diff
@@ -470,6 +473,64 @@ async fn receive_batch(
...
+                            batch_query
+                                .request_types
+                                .push(RESET_CHECK_MESSAGE_TYPE.to_string());
+                            batch_query.metadata.push(batch_metadata);
+                            batch_query.sns_message_ids.push(sns_message_id);
+
+                            // skip_persistence is only used for uniqueness requests
+                            batch_query.skip_persistence.push(false);
+
+                            // We don't need OR rule for reset check
```
wojciechsromek, line 514: Maybe state that we NEED to use and rule here?

wojciechsromek (approved):

philsippl (approved):

PR #1221, opened 2025-04-03
Aurel/state check

naure:
Continuing the spirit of validating the state that is supposed to stay in sync across nodes.

- [x] Introduce a hasher for sets (ignoring order).
- [x] Hash the iris store.
- [x] Hash the graph, with efficient mutations.
- [x] Compare notes with the other parties on startup.
- [x] Compare again after each batch.

Benchmark (laptop): `350ns` to hash the largest graph node; expecting a few seconds total on startup.

### DIFF
--- a/iris-mpc-cpu/src/execution/hawk_main/state_check.rs
+++ b/iris-mpc-cpu/src/execution/hawk_main/state_check.rs
@@ -1,4 +1,5 @@
 use eyre::{eyre, Result};
+use itertools::{chain, Itertools};
 use serde::{Deserialize, Serialize};
 use siphasher::sip::SipHasher13;
 use std::hash::{Hash, Hasher};
@@ -11,11 +12,11 @@ pub struct SetHash {
 }
 
 impl SetHash {
-    pub fn add(&mut self, value: &impl Hash) {
+    pub fn add_unordered(&mut self, value: impl Hash) {
         let mut hasher = SipHasher13::default();
         value.hash(&mut hasher);
-
-        self.hash ^= hasher.finish();
+        let h = hasher.finish();
+        self.hash = self.hash.wrapping_add(h);
     }
 
     pub fn digest(&self) -> u64 {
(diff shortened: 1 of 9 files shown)


iris-mpc-cpu/src/execution/hawk_main/state_check.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub fn digest(&self) -> u64 {
+        self.accumulator
+    }
+
+    fn hash(value: impl Hash) -> u64 {
+        let mut hasher = SipHasher13::default();
+        value.hash(&mut hasher);
+        hasher.finish()
+    }
+}
```
dkales, line 32: This could be very easy to misuse IMO, since it just takes all impl Hash, which do not enforce the correct usage on call sites. It is fine RN since the interface is actually pretty small, but could it make sense to have it be something like
```rust
struct SetHash<T: Hash> {
  accumulator: u64,
  phantom: PhantomData<T>
 }
 impl<T: Hash> SetHash<T> {
   pub fn add_unordered(&mut self, value: T {
        self.accumulator = self.accumulator.wrapping_add(Self::hash(value));
    }
    ...
 }
 ```
 
 And restrict a single instance of the SetHash to be for one type only?
 
 Also can probably just take references here instead to avoid moving the hashed value in here.

File: iris-mpc-cpu/src/execution/hawk_main/state_check.rs [hunk 2]
```diff
@@ -0,0 +1,112 @@
...
+
+impl SetHash {
+    pub fn add_unordered(&mut self, value: impl Hash) {
+        self.accumulator = self.accumulator.wrapping_add(Self::hash(value));
+    }
+
+    pub fn remove(&mut self, value: impl Hash) {
+        self.accumulator = self.accumulator.wrapping_sub(Self::hash(value));
+    }
```
dkales, line 21: Think the accumulation here is fine for the use-case but should keep in mind that this method + the data types here allow a pretty easy ~2^32 cost collision attack.

dkales (approved):

iris-mpc-cpu/src/execution/hawk_main/state_check.rs
naure: This can be more type safe, although I do use it for fields of different types (graph entry point and layers), and writing the T can get verbose (enums, lifetimes, etc).

It does work by reference or tuples of references.

iris-mpc-cpu/src/execution/hawk_main/state_check.rs
naure: Definitely not cryptographic hash, it is a checksum.

naure: Renamed to `checksum`.

PR #1224, opened 2025-04-07
Handle reset update requests

eaypek-tfh:
## Change
- This is the second part on reset logic. The first one was https://github.com/worldcoin/iris-mpc/pull/1219
- In this one, we replace the iris code shares for successful reset requests sent from signup-service.
- We don't increment batch counter for this type of message and treat it like deletion messages. The reason is because we do not need a comparison, we just update without checks.
- Since this is a new modification on db and memory, we leverage current modifications table and sync mechanism to sync and replay reset_update changes.

### DIFF
--- a/iris-mpc-gpu/tests/e2e.rs
+++ b/iris-mpc-gpu/tests/e2e.rs
@@ -21,6 +21,7 @@ mod e2e_test {
     const MATCH_DISTANCES_BUFFER_SIZE: usize = 1 << 7;
     const MATCH_DISTANCES_BUFFER_SIZE_EXTRA_PERCENT: usize = 100;
     const MAX_DELETIONS_PER_BATCH: usize = 10;
+    const MAX_RESET_UPDATES_PER_BATCH: usize = 10;
 
     fn install_tracing() {
         tracing_subscriber::registry()
@@ -206,6 +207,7 @@ mod e2e_test {
                 NUM_BATCHES,
                 MAX_BATCH_SIZE,
                 MAX_DELETIONS_PER_BATCH,
+                MAX_RESET_UPDATES_PER_BATCH,
                 [&mut handle0, &mut handle1, &mut handle2],
             )
             .await?;
--- a/iris-mpc-gpu/tests/e2e.rs
+++ b/iris-mpc-gpu/tests/e2e.rs
@@ -8,14 +8,13 @@ mod e2e_test {
         test::{load_test_db, TestCaseGenerator},
     };
     use iris_mpc_gpu::{helpers::device_manager::DeviceManager, server::ServerActor};
+    use rand::random;
     use std::{env, sync::Arc};
     use tokio::sync::oneshot;
     use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
     const DB_SIZE: usize = 8 * 1000;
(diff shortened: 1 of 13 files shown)


iris-mpc-gpu/tests/e2e.rs
```diff
@@ -16,7 +16,7 @@ mod e2e_test {
     const DB_BUFFER: usize = 8 * 1000;
     const DB_RNG_SEED: u64 = 0xdeadbeef;
     const INTERNAL_RNG_SEED: u64 = 0xdeadbeef;
-    const NUM_BATCHES: usize = 30;
+    const NUM_BATCHES: usize = 40;
```
eaypek-tfh, line 19: I needed to bump it a bit so that `ResetUpdate` and `MatchAfterResetUpdate` could be included. With previous random seeders and batch size, they were not included.

dkales: I think maybe we should let this be a bit more random in the sense we do pick a truly random u64 as a seed for the testing, however, we print/log it for reproducibility still?

eaypek-tfh: On the advantage side, it covers different combinations in batches but on the other side, we lose reproducibility. I don't have a strong opinion. Wdyt @philsippl?

dkales: My idea would be something like
```rust
let seed = if let Some(seed) = env::var!("SEED") {
  seed.parse()
 } else {
   thread_rng().gen()
 };
 tracing::info!("Seed for this test run is: {}, seed);
 ```
 
 To have both, one could also pass the seed to the CI action if needed to run it there again?

File: iris-mpc/bin/server.rs [hunk 2]
```diff
@@ -2020,13 +2175,16 @@ async fn server_main(config: Config) -> eyre::Result<()> {
 
             metrics::histogram!("receive_batch_duration").record(now.elapsed().as_secs_f64());
 
-            process_identity_deletions(
+            // Persist deletions and updates to postgres db before the actor processes the batch.
+            // This way, there is no need to pass shares back from actor to the server
```
philsippl, line 2179: @eaypek-tfh don't understand this comment or can't remember the details, can you explain more?

philsippl (approved):
lgtm!

iris-mpc/bin/server.rs
eaypek-tfh: Of course. When implementing deletions, we decided not to count them as part of the batch because there's no need for 1-to-N comparison for those requests. So, we send them as a sub-query to the actor along with batch so that we can update in-memory db as well.

This comment points that we treat reset-updates just like deletions because they also don't need any comparison. For deletions, we opted to persist into postgres here before actor persists into memory. The advantage is that we don't need to pass reset_update shares back from actor to server as part of `ServerJobResult` just for postgres persistence.

But now that I think more, this can cause a problem with modifications because the modification row will still be `in_progress` while the irises row will be updated already. So, we can either mark modification as `completed` upon writing into irises here, or we can write deletions and updates together with other types [here](https://github.com/worldcoin/iris-mpc/blob/1589977876d02a8316998862d056c066f9c9d796/iris-mpc/bin/server.rs#L1764-L1791) (enrollment, reauth, etc)

PR #1229, opened 2025-04-10
calculate bucket statistics for HNSW

carlomazzaferro:
integrates the work from @dkales into the actor to calculate bucket statistics for HNSW.

### DIFF
--- a/iris-mpc-cpu/src/execution/hawk_main.rs
+++ b/iris-mpc-cpu/src/execution/hawk_main.rs
@@ -32,6 +32,7 @@ use iris_mpc_common::{
 };
 use itertools::{izip, Itertools};
 use rand::{thread_rng, Rng, SeedableRng};
+use std::slice::Iter;
 use std::{
     collections::HashMap,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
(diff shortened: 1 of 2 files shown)


iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -454,6 +533,49 @@ impl HawkActor {
         Ok(connect_plan)
     }
 
+    async fn calculate_threshold_a(n_buckets: usize) -> Vec<u32> {
+        let thresholds: Vec<f64> = (0..n_buckets)
+            .map(|i| i as f64 / (n_buckets * 2) as f64)
+            .collect();
+
```
dkales, line 540: This goes from [0,0.5), in GPU we go from [0,0.375] if I am not mistaken

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -454,6 +533,49 @@ impl HawkActor {
...
+            let bucket_result_shares = compare_threshold_buckets(
+                &mut session.aby3_store.session,
+                translated_thresholds.as_slice(),
+                distances.as_slice(),
+            )
+            .await?;
+
+            let bucket_results =
+                open_ring(&mut session.aby3_store.session, &bucket_result_shares).await?;
+            buckets.push(bucket_results)
```
dkales, line 565: revealing many sub bucket-caclulations here seems to defeat the purpose of aggregating this for privacy a bit.
We could/should instead collect the distances in a large global vec and run a single bucket comparison on it IMO

dkales:
In general, this seems like we calculate many small histograms and reveal them for each query, I think that this goes against the spirit of this, and we should set aside a match distance buffer internally that the queries add the still secret-shared dot products to, and once it is sufficiently filled, we do the bucket calculation once.

iris-mpc-cpu/src/execution/hawk_main.rs
carlomazzaferro: makes sense, will try to adapt to this!

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -362,6 +370,86 @@ impl HawkActor {
...
+
+        let buckets = open_ring(&mut session.aby3_store.session, &bucket_result_shares).await?;
+        Ok(buckets)
+    }
+
+    pub async fn fill_anonymized_statistics_buckets(
+        &mut self,
+        session: &mut HawkSession,
+    ) -> Result<()> {
+        if self.distances_cache[LEFT].len() > self.args.match_distances_buffer_size {
```
philsippl, line 423: At some point we should probably rename this config option, because it's a bit confusing with this implementation (since the buffer is not hard capped)

philsippl (approved):
looks good for me

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -733,6 +821,29 @@ impl HawkHandle {
...
+                let mut right_session = hawk_session_right.write().await;
+
+                hawk_actor
+                    .fill_anonymized_statistics_buckets(&mut left_session)
+                    .await?;
+                hawk_actor
+                    .fill_anonymized_statistics_buckets(&mut right_session)
+                    .await?;
+
                 let match_result = {
```
naure, line 847: This block should be wrapped in a function to release the locks.

PR #1240, opened 2025-04-14
aurel/optional-vectors: Support missing vectors

naure:


bgillesp (approved):
Looks good!  Taking advantage of the naturally occurring neighborhood updates to remove deleted graph references is a really nice idea.

PR #1243, opened 2025-04-15
Update e2e tests with mirror checks not being persisted

danielle-tfh:
### Notes
* ensure no mirror checks are flagged for non-mirror tests

### DIFF
--- a/iris-mpc-common/src/test.rs
+++ b/iris-mpc-common/src/test.rs
@@ -1126,13 +1126,14 @@ impl TestCaseGenerator {
     ) {
         tracing::info!(
             "Checking result for request_id: {}, idx: {}, was_match: {}, matched_batch_req_ids: \
-             {:?}, was_reauth_success: {}, was_skip_persistence_match: {}",
+             {:?}, was_reauth_success: {}, was_skip_persistence_match: {}, full_face_mirror_attack: {}",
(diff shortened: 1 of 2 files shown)


iris-mpc-common/src/test.rs
```diff
@@ -1090,19 +1089,19 @@ impl TestCaseGenerator {
...
                 // Flip right bits to above threshold - (right) does not match
                 for i in 0..(THRESHOLD_ABSOLUTE as i32 + variation) as usize {
-                    code_right.code.flip_bit(i);
+                    code_right.code.flip_bit(i * 2);
                 }
             } else {
                 // Flip left bits to above threshold - (left) does not match
                 for i in 0..(THRESHOLD_ABSOLUTE as i32 + variation) as usize {
-                    code_left.code.flip_bit(i);
+                    code_left.code.flip_bit(i * 2);
```
dkales, line 1097: Why these *2? The location of the flipped bits should not matter right?

dkales (approved):

iris-mpc-common/src/test.rs
danielle-tfh: Yeah - they do not matter. I was just running some tests to see if it made a difference to re-auths being tagged as mirror attacks

PR #1244, opened 2025-04-15
Reduce processing time of mirror attacks

danielle-tfh:
### Notes
* remove processing reset update requests, LUC and deletion for mirror checks - this will improve performance and ensure these requests are not processed twice.
* LUC is not needed for mirror attacks since half mirror attacks are already checked on the orb

### DIFF
--- a/iris-mpc-gpu/src/server/actor.rs
+++ b/iris-mpc-gpu/src/server/actor.rs
@@ -666,7 +666,9 @@ impl ServerActor {
             "Reset update batch sizes mismatch"
         );
 
-        if !batch.or_rule_indices.is_empty() || batch.luc_lookback_records > 0 {
+        if (!batch.or_rule_indices.is_empty() || batch.luc_lookback_records > 0)
+            && orientation == Orientation::Normal
+        {
             assert!(
                 (batch.or_rule_indices.len() == batch_size) || (batch.luc_lookback_records > 0)
             );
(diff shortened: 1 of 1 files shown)


iris-mpc-gpu/src/server/actor.rs
```diff
@@ -717,7 +719,7 @@ impl ServerActor {
         ///////////////////////////////////////////////////////////////////
         // PERFORM DELETIONS (IF ANY)
         ///////////////////////////////////////////////////////////////////
-        if !batch.deletion_requests_indices.is_empty() {
+        if !batch.deletion_requests_indices.is_empty() && orientation == Orientation::Normal {
```
dkales, line 722: This is a problem, since if we do not process the deletion requests in the mirror flow already, the two flows have different views of the DB. (Mirror flow has unmodified DB, Normal ones has the DB with deletions). I do not know if that can present an issue in practice though?

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -760,7 +762,7 @@ impl ServerActor {
         ///////////////////////////////////////////////////////////////////
         // PERFORM RESET UPDATES (IF ANY)
         ///////////////////////////////////////////////////////////////////
-        if !batch.reset_update_request_ids.is_empty() {
+        if !batch.reset_update_request_ids.is_empty() && orientation == Orientation::Normal {
```
dkales, line 765: same for resets I guess?

dkales (approved):

PR #1246, opened 2025-04-16
aurel/shared-rng: Setup a common RNG state with fresh seeds

naure:
Previously, the shared RNG was seeded with constant seeds, so the sequence of decisions (e.g. HNSW layer selection) were the same after restarts.

Now, seed randomly at each start.

### DIFF
--- a/iris-mpc-cpu/src/protocol/ops.rs
+++ b/iris-mpc-cpu/src/protocol/ops.rs
@@ -54,21 +54,20 @@ pub async fn setup_replicated_prf(session: &mut NetworkSession, my_seed: PrfSeed
 }
 
 /// Setup an RNG common between all parties, for use in stochastic algorithms (e.g. HNSW layer selection).
-///
-/// Security: honest-but-curious, i.e. a malicious party can influence the RNG outcomes.
 pub async fn setup_shared_rng(session: &mut NetworkSession, my_seed: PrfSeed) -> Result<AesRng> {
     let my_msg = NetworkValue::PrfKey(my_seed).to_network();
-
-    // send my_seed to the other parties
-    session.send_prev(my_msg.clone()).await?;
-    session.send_next(my_msg).await?;
 
     let decode = |msg| match NetworkValue::from_network(msg) {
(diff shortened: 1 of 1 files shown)


iris-mpc-cpu/src/protocol/ops.rs
```diff
@@ -50,6 +53,28 @@ pub async fn setup_replicated_prf(session: &mut NetworkSession, my_seed: PrfSeed
...
+/// Setup an RNG common between all parties, for use in stochastic algorithms (e.g. HNSW layer selection).
+///
+/// Security: honest-but-curious, i.e. a malicious party can influence the RNG outcomes.
+pub async fn setup_shared_rng(session: &mut NetworkSession, my_seed: PrfSeed) -> Result<AesRng> {
+    let my_msg = NetworkValue::PrfKey(my_seed).to_network();
+
+    // send my_seed to the other parties
+    session.send_prev(my_msg.clone()).await?;
+    session.send_next(my_msg).await?;
+
```
bgillesp, line 65: This approach is great under our semi-honest assumption.  Could we tweak the implementation slightly to make this component robust even in the absence of this assumption by interleaving the send and receive calls?  So it would be something like

```rust
session.send_next(my_msg.clone()).await?;
let prev_seed = decode(session.receive_prev().await)?;

session.send_prev(my_msg).await?;
let next_seed = decode(session.receive_next().await)?;
```

This way a dishonest adversary can only see one of the two other seeds before submitting their seed to other parties (the "next" party won't send their seed back until they've received a seed already), so it makes it more challenging to manipulate the resulting shared seed.

bgillesp (approved):
Great approach for this, thanks!  I had just one small suggestion for a tweak in the `setup_shared_rng` protocol in the comments, if it seems like a good idea.

iris-mpc-cpu/src/protocol/ops.rs
naure: I wanted it in one round initially, but yes we might as well go for the two rounds approach. Done.

PR #1247, opened 2025-04-16
aurel/parallel-setup: Create sessions in parallel

naure:
_Follow-up to #1246._

Create sessions in parallel to reduce startup time for many sessions.

This also opens up the possibility of creating new sessions for each batch.


bgillesp (approved):
Looks good to go!  The refactored structure of `new_sessions`/`create_session` is really clean. :+1:

PR #1249, opened 2025-04-16
Implement seeded hnsw randomness for tests

bgillesp:
This PR adds a configuration parameter for an explicit base seed to be used for HNSW graph generation.  This seed is used to derive individual RNGs associated with `HawkSession` structs, constructed by hashing this base seed, the `StoreId`, and the `SessionId` into a final `u64` seed.  (Note this construction needs to produce statistical randomness, but does not need to be cryptographically secure, so a `u64` seed suffices.)  If no explicit base seed is specified in the server `Config` (the `hnsw_prng_seed` parameter is set to `None`) then the default behavior is used which instead chooses a random seed per session agreed upon by the server parties.  (See #1246.)


siajasl (commented):
Code changes make sense.  Am now testing locally.

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -482,6 +498,13 @@ impl HawkActor {
     }
 }
 
+pub fn session_seeded_rng(base_seed: u64, store_id: StoreId, session_id: SessionId) -> ChaCha8Rng {
```
naure, line 501: Why not use the same type `AesRng`?

naure:
I merged #1246 and changed the base to `main`.

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -485,6 +498,13 @@ impl HawkActor {
...
+    let mut hasher = DefaultHasher::new();
```
naure, line 502: There is no specification of what this is. Better use explicitly `siphasher::sip::SipHasher13`.

siajasl (approved):
After testing it looks good to me.

iris-mpc-cpu/src/execution/hawk_main.rs
bgillesp: Reason is that `AesRng` doesn't support serialization (by design I believe, since it's meant to be used for secure applications), while I needed to be able to recreate the full RNG state between runs of a binary for reproducibility in some testing scenarios.

siajasl (commented):
LGTM

PR #1251, opened 2025-04-17
aurel/refactor-stats: Fix and simplify anon stats

naure:
- Fix double call to `fill_anonymized_statistics_buckets` (it is called twice _and_ it used to process both eyes inside).
- Fix: look only at the bottom layer.
- Move code to its own function (generally trying to simplify the main loop).
- Clarify that a single networking session is relevant here (not hawk session with left/right stores).

### DIFF
--- a/iris-mpc-cpu/src/execution/hawk_main.rs
+++ b/iris-mpc-cpu/src/execution/hawk_main.rs
@@ -35,7 +35,6 @@ use iris_mpc_common::{
 };
 use itertools::{izip, Itertools};
 use rand::{thread_rng, Rng};
-use std::slice::Iter;
 use std::{
     collections::HashMap,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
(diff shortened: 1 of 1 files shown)


iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -402,19 +399,31 @@ impl HawkActor {
...
+        session: &HawkSessionRef,
+        search_results: &BothEyes<VecRequests<VecRots<InsertPlan>>>,
+    ) -> Result<()> {
+        let mut session = session.write().await;
+
+        for side in [LEFT, RIGHT] {
+            self.cache_distances(side, &search_results[side]);
+
+            self.fill_anonymized_statistics_buckets(&mut session.aby3_store.session, side)
+                .await?;
```
bgillesp, line 413: In theory the `LEFT` and `RIGHT` calls here could be run in parallel in different sessions I think, but it may not be worth the hassle since they're relatively shallow operations in terms of round complexity.

bgillesp (approved):
Solid refactor, thanks!

iris-mpc-cpu/src/execution/hawk_main.rs
naure: Right. And the whole thing could also run in parallel to other steps. This will require a little more refactor to not capture `&mut HawkActor`.

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -402,19 +399,31 @@ impl HawkActor {
...
             })
             .collect::<Vec<_>>()
     }
 
-    pub async fn cache_distances(
-        &mut self,
-        side: usize,
-        search_results: Iter<'_, VecRots<InsertPlan>>,
-    ) -> Result<()> {
+    fn cache_distances(&mut self, side: usize, search_results: &[VecRots<InsertPlan>]) {
```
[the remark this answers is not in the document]
naure: Good catch! The higher layers are not nearest neighbors so their distance is meaningless here. Done https://github.com/worldcoin/iris-mpc/pull/1251/commits/f28e1f03d41f550e12e011f60adea9bde2037878

PR #1254, opened 2025-04-18
aurel/state_check_type: Use NetworkValue for state checks

naure:


bgillesp (approved):
Looks good!  Definitely seems like a good idea to use a consistent communication interface for messages on the gRPC pipes.

PR #1260, opened 2025-04-22
Load the second half of the database on demand for GPU version

dkales:
This is a first draft of functionality to load the second half (meaning left/right eye) of the DB (which is only searched based on some subset due to OR indices and partial results of first half of DB) on demand.
I have implemented a first version of the loader for the GPU E2E tests, this is actually most of the changes since we did not keep track of DB changes in these tests beforehand. As a side effect the Actor test setup now essentially has an expected view of the 3 node DBs for the parties after the test, in case this is useful.

TODOs:
- [ ] Implement OnDemandLoader for real DB (might need interface changes for the trait since that is most likely async and the actor impl is not)
- [ ] Test performance implications for real DB (This is something I cannot really do nicely outside of the staging setup I guess?)

Some other thoughts:
- The current test cases do not check that resets/reauths actually modified the database in the intended way, should this be added somehow?


iris-mpc-gpu/src/server/actor.rs
```diff
@@ -2912,41 +3032,60 @@ fn write_db_at_index(
...
-                streams[device_index].stream,
-            );
+                helpers::dtoh_at_offset(
+                    db.code_gr.limb_1[device_index],
+                    dst_index * code_length,
+                    *query.limb_1[device_index].device_ptr(),
+                    code_length * 15 + src_index * code_length * ROTATIONS,
+                    code_length,
+                    streams[device_index].stream,
+                );
```
philsippl, line 3052: why duplicated?

iris-mpc-gpu/src/dot/share_db.rs
```diff
@@ -571,6 +571,87 @@ impl ShareDB {
         }
     }
 
+    pub fn load_host_db_subset_into_chunk_buffers(
+        &self,
+        db: &[Vec<Vec<u16>>],
+        buffers: &DBChunkBuffers,
+        streams: &[CudaStream],
+    ) {
+        // TODO: check if we should do all of them concurrently instead of one by one
```
philsippl, line 580: yeah I think moving the synchronize out of the loop should be fine.
We might also want to do the preprocessing of the slices in parallel before, but I think this is fine as a start. Let's add some timings here to see how long preprocessing and copying takes?

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1902,11 +1927,117 @@ impl ServerActor {
...
+            tracing::info!("{} DB indices: {:?},", self.party_id, db_indices);
+
+            // fetch them from the DB
+            let now = Instant::now();
+            // since we are in a tokio::spawn_blocking context, we can use the tokio runtime handle
+            let iris_codes = tokio::runtime::Handle::current().block_on(async {
+                on_demand_loader
+                    .load_records(eye_db, &db_indices)
+                    .await
+                    .expect("Can load records from on-demand loader")
```
philsippl, line 1950: note: theoretically we could start streaming in those as we find them in the previous search. But this would also require how fetch those partial matches. Currently we fetch them after having processed all chunks, we could do that instead for every chunk and stream those already. Same for the `OR` indices, they are known at the start already.

iris-mpc-gpu/src/server/actor.rs
dkales: copy-paste error it seems

iris-mpc-gpu/src/dot/share_db.rs
dkales: Atm we cannot do so since the copied data is only allocated inside the loop and will become invalid. Will move them outside and sync after.

iris-mpc-gpu/src/dot/share_db.rs
```diff
@@ -571,6 +571,87 @@ impl ShareDB {
...
+
+            let a0_host = db[idx]
+                .iter()
+                .flat_map(|record| record.iter().map(|&x| ((x as i8) as i32 - 128) as i8))
+                .collect::<Vec<_>>();
+
+            let a1_host = db[idx]
+                .iter()
+                .flat_map(|record| record.iter().map(|&x: &u16| ((x >> 8) as i32 - 128) as i8))
+                .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
dkales: This mapping could be done and kept around at a caching layer as well?

dkales:
Have now addressed most of the comments here, should be good for staging now @philsippl

philsippl:
looking great! @danielle-tfh could you give this a try on staging?

PR #1266, opened 2025-04-24
Refactor server_main into smaller function calls

bgillesp:
This PR restructures the primary HNSW `server_main` function by pulling out logical components of the server functionality into individual functional calls.  This restructuring is in preparation for implementing the "genesis" indexer, which will be used to initially construct an HNSW graph from pre-existing secret shared iris shares, and which will share much (but not all) of the functionality of `server_main`.


siajasl (approved):
LGTM

PR #1275, opened 2025-04-25
aurel/pipeline-receive-batch

naure:
Receive the next batch in parallel to processing the current batch.

With `channel(1)`, when the main takes the ready batch (`batch_stream.recv()`), a `permit` is given to start receiving the next batch in the background.


philsippl (approved):
looks good! let's also bring this into the "gpu" server?

naure:
> looks good! let's also bring this into the "gpu" server?

Yes, will do.

PR #1279, opened 2025-04-25
Update init-test-dbs binary to improve usability

bgillesp:
Two changes are implemented to the `init-test-dbs` binary:
- Transactions committing secret-shared irises to Postgres have batch size reduced from 1,000 to 100
- Secret sharing, a relatively lightweight operation, is executed at the start of the binary rather than after the quite long graph construction process

The Postgres commit batch size is now controlled by the `SECRET_SHARING_PG_TX_SIZE` const in the binary, so can be further adjusted as needed.

### DIFF
--- a/iris-mpc-cpu/bin/init_test_dbs.rs
+++ b/iris-mpc-cpu/bin/init_test_dbs.rs
@@ -34,8 +34,17 @@ use tracing::{info, warn};
 
 use eyre::Result;
 
+/// Number of MPC parties.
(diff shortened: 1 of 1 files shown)


sdwoodbury (approved):

siajasl (approved):
Looks good.  However I feel that the main function is beginning to require a degree of rationalization ... i.e. factoring out into a set of functions much like what is being done for the main binary.  But that is not within scope for this PR.

PR #1283, opened 2025-04-27
revert dockerfile.base changes

eaypek-tfh:
## Change
- For prod release, let's remove the Dockerfile.base changes as we do not know about their performance yet


asurak (approved):
LGTM

PR #1285, opened 2025-04-28
aurel/fix-or_rule

naure:
There is this difference between GPU and CPU server code:
  [iris-mpc/bin/server.rs#L365](https://github.com/worldcoin/iris-mpc/blob/24780da09f777948998978446a8dd28b5365f303/iris-mpc/bin/server.rs#L365)
  [iris-mpc/src/services/processors/batch.rs#L564](https://github.com/worldcoin/iris-mpc/blob/24780da09f777948998978446a8dd28b5365f303/iris-mpc/src/services/processors/batch.rs#L564)

It looks like a bug fix that got lost somehow ([#1107](https://github.com/worldcoin/iris-mpc/pull/1107/files)). Re-applying it here.

---

There may be a second issue around this. The function `receive_batch` looks like this:

```rust
match request_type {
    UNIQUENESS => {
        if luc_enabled {
            if luc_serial_ids_from_smpc_request {
                if … uniqueness_request.or_rule_serial_ids {
                    batch_query.or_rule_indices.push( … );
                } else {
                    batch_query.or_rule_indices.push(vec![]);
                }
            }
        }
    }

    REAUTH | RESET => {
        batch_query.or_rule_indices.push( … );
    }
}
```

There are paths without `push` so `or_rule_indices` becomes misaligned with the requests.

---

Potentially a third issue in function `generate_luc_records`:

```rust
    // If there are no OR rules, return only the lookback records
    if or_rule_indices.is_empty() {
        return lookback_records;
    }

    // Otherwise, merge them into each inner vector of or_rule_indices
    for (idx, (or_ids, luc_ids)) in
        izip!(or_rule_indices.iter_mut(), lookback_records.iter()).enumerate()
    {
        if skip_lookback_requests.contains(&idx) {
            continue;
        }
        // Add the lookback IDs
        or_ids.extend_from_slice(luc_ids);
```

But the path `return lookback_records` does not honor `skip_lookback_requests`, which meant to skip reauth and reset requests, and instead it includes the lookback records also for reauth and reset requests. It may be that this case never happens but that is hard to review.

### DIFF
--- a/iris-mpc-gpu/src/server/actor.rs
+++ b/iris-mpc-gpu/src/server/actor.rs
@@ -3032,7 +3032,7 @@ pub fn generate_luc_records(
 
     // If there are no OR rules, return only the lookback records
     if or_rule_indices.is_empty() {
-        return lookback_records;
+        or_rule_indices = vec![vec![]; batch_size];
     }
 
     // Otherwise, merge them into each inner vector of or_rule_indices


iris-mpc-gpu/src/server/actor.rs
```diff
@@ -3032,7 +3032,7 @@ pub fn generate_luc_records(
 
     // If there are no OR rules, return only the lookback records
     if or_rule_indices.is_empty() {
-        return lookback_records;
+        or_rule_indices = vec![vec![]; batch_size];
```
dkales, line 3035: why this change? I guess the final result is the same but it does more work now?

dkales (approved):

iris-mpc-gpu/src/server/actor.rs
dkales: I saw your explanation for the third case above, nvm.

PR #1295, opened 2025-05-01
aurel/luc: Match with LUC entries and apply the OR rule

naure:
- [x] Find the spatial LUC IDs from requests `or_rule_indices`.
- [x] Find the temporal LUC IDs from `luc_lookback_records`.
    - Does not apply to REAUTH and RESET.
- [x] Send the LUC IDs into the existing pipeline where the actual comparisons happen.
    - Called Step 1: `missing_vector_ids -> calculate_missing_is_match` 
- [x] Matching logic on LUC results with the OR rule.
    - Called Step 2: `luc_results -> is_match`
- [x] Introduce `VectorId::any_version`.
    - This allows passing both versioned and unversioned IDs through the same pipeline.
    - Export aliases `SerialId` and `VersionId` while we’re there.
- [x] Lookup irises by serial ID for when the version is not known. Move the version from HashMap key to value.
- [x] Refactor to forward `batch: BatchQuery` through `HawkRequest` and `HawkResult` where it is needed. Move `filter_for_insertion`.


iris-mpc-cpu/src/execution/hawk_main/matching.rs
```diff
@@ -93,17 +98,31 @@ impl Step1 {
...
+            luc_ids: Vec::new(),
         }
     }
 
     fn missing_vector_ids(&self, side: usize) -> VecEdges<VectorId> {
         let other_side = 1 - side;
-        self.anti_join[other_side].clone()
+        let anti_join = &self.anti_join[other_side];
+
+        chain!(anti_join, &self.luc_ids).cloned().collect_vec()
```
dkales, line 109: This could use a `unique()` to reduce work a bit in some scenarios, although the correctness should still work without it, right?

dkales (approved):

iris-mpc-cpu/src/execution/hawk_main/matching.rs
naure: Good point, I added that.

This required aligning to a single kind of "versioned" ID. Then I could also remove the suspicious idea of ANY_VERSION too.

PR #1299, opened 2025-05-01
use `eyre::bail!` macro and `eyre::Result` consistently

einar-inversed3:
`bail!(<args>)` is shorthand [1] for `return Err(eyre!(<args>))`, so let's use it.

[1]: https://docs.rs/eyre/latest/eyre/macro.bail.html#:~:text=This%20macro%20is,Err(eyre!(%3Cargs%3E))


siajasl (approved):
LGTM.  However I feel that @bgillesp should also review as it touches upon code written by TFH team.

einar-inversed3:
@bgillesp what do you think about this PR?

bgillesp:
Definitely in favor of using more concise syntax when available!  I don't see any issues with this change in terms of touching existing code -- as you say, it should be equivalent.

If we're already tidying up usage of `eyre`, would you also be able to take a pass through the codebase and replace all uses of inline qualifications of `eyre::Result` with `use eyre::Result` and just `Result` in the code?  I've been thinking about making a quick PR for this myself, and it seems like this could be a reasonable place to include the change.  (Or feel free to make a separate PR if you want to keep the scope tidy here.)

einar-inversed3:
Sure, I will continue on this branch.

einar-inversed3:
`Result<T>` is now used consistently as well.

PR #1303, opened 2025-05-02
fix: introduce separate bucket recv buffer

philsippl:


danielle-tfh (approved):

iris-mpc-gpu/src/threshold_ring/protocol.rs
```diff
@@ -286,6 +287,9 @@ impl Buffers {
         let ot_m1 = Some(Self::allocate_single_buffer(chunk_size * 128, devices));
         let ot_wc = Some(Self::allocate_single_buffer(chunk_size * 128, devices));
 
+        let buckets_recv_buffer =
+            n_buckets.map(|n_buckets| Self::allocate_buffer(n_buckets * 32, devices));
+
```
dkales, line 292: this does not need to be `* 32`, right?

dkales (commented):
Think one buffer len is too large, also, why do we need this, in case the buckets are too large for the buffer we used in the meantime? Or was there any parallel processing issue going on with the concrete buffer?

iris-mpc-gpu/src/threshold_ring/protocol.rs
philsippl: ah right, copy paste :)

dkales (approved):

PR #1304, opened 2025-05-02
Chore: simplify PlaintextStore data types and interface

bgillesp:
This PR removes the `PlaintextIris` and `PlaintextPoint` wrapper structs previously used to represent iris codes in `PlaintextStore`, and simplifies the query type.  `PlaintextIris::QueryRef` is now `Arc<IrisCode>`, allowing function calls to pass the query iris code directly, rather than first inserting the code into the store's data storage with a "persisted" flag set to false.  Additionally, the `PlaintextStore::create_graph` function now requires an explicit `HnswParams` struct rather than resorting to a default value.

### DIFF
--- a/iris-mpc-cpu/bin/generate_benchmark_data.rs
+++ b/iris-mpc-cpu/bin/generate_benchmark_data.rs
@@ -1,7 +1,7 @@
 use aes_prng::AesRng;
 use iris_mpc_cpu::{
     hawkers::plaintext_store::PlaintextStore,
-    hnsw::HnswParams,
+    hnsw::{HnswParams, HnswSearcher},
     py_bindings::{io::write_bin, plaintext_store::to_ndjson_file},
 };
 use rand::SeedableRng;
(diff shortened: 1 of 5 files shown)


iris-mpc-cpu/bin/generate_benchmark_data.rs
```diff
@@ -13,13 +14,18 @@ async fn main() -> Result<(), Box<dyn Error>> {
     std::fs::create_dir_all(format!("{crate_root}/data"))?;
     let mut rng = AesRng::seed_from_u64(0_u64);
     println!("Generating plaintext store with 100_000 irises");
-    let mut store = PlaintextStore::create_random_store(&mut rng, 100_000).await?;
+    let mut store = PlaintextStore::new_random(&mut rng, 100_000).await;
     println!("Writing store to file");
     to_ndjson_file(&store, &format!("{crate_root}/data/store.ndjson"))?;
 
+    let searcher = HnswSearcher {
+        params: HnswParams::new(320, 256, 256),
```
iliailia, line 22: Let's make arguments for the HNSW parameters. If none given, stick to the default ones.

bgillesp: Sure, can do -- I'm in process of collapsing a few dependent PRs right now, so I'll make a new small PR for it later today once everything is merged.

iliailia (approved):
Neat! Thanks!

PR #1308, opened 2025-05-06
Danielle/add separate genesis folder

danielle-tfh:
### Notes
* add seperate package for genesis
* this ensures that genesis changes do not mean changing the main binary, keeps the main HNSW service clean.
* we also move common used  methods to common  package - this will also be helpful for future sync mechanisms like v3 codes and other recovery protocols


siajasl (approved):
LGTM.

PR #1328, opened 2025-05-08
[POP-2447] Introduce anonymized statistics for mirror orientated computation

leonanos8:
### Description
- Replicate the process of collecting and communicating anonymized matching statistics for the mirror orientated computation
- Duplicate the match distances buffers. Keep buffers for normal and mirror orientation
- Introduce env var to control wether we send the mirror anon stats
- Remove the mirror flag from kernel code. Now we want to compute anon stats for both orientations. **actor** is responsible to pass the correct orientation distance buffers to the kernel code

~~Still wip:~~

- [x] Add check for mirror anon stats in the e2e tests

### DIFF
--- a/iris-mpc-gpu/src/server/actor.rs
+++ b/iris-mpc-gpu/src/server/actor.rs
@@ -150,6 +150,7 @@ pub struct ServerActor {
     mask_chunk_buffers: Vec<DBChunkBuffers>,
     phase1_events: Vec<Vec<CUevent>>,
     phase2_events: Vec<Vec<CUevent>>,
+    // Normal orientation buffers
     match_distances_buffer_codes_left: Vec<ChunkShare<u16>>,
     match_distances_buffer_codes_right: Vec<ChunkShare<u16>>,
     match_distances_buffer_masks_left: Vec<ChunkShare<u16>>,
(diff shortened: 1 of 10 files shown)


iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1591,7 +1652,7 @@ impl ServerActor {
         Ok(result)
     }
 
-    fn try_calculate_bucket_stats(&mut self, eye_db: Eye) {
+    fn try_calculate_bucket_stats(&mut self, eye_db: Eye, is_mirror: bool) {
```
dkales, line 1655: can we reuse the Orientation enum here for more clarity at call sites?

dkales (commented):
LGTM, minor comment

dkales (approved):

iris-mpc-gpu/src/server/actor.rs
leonanos8: Good point, changed here -> [a18599b](https://github.com/worldcoin/iris-mpc/pull/1328/commits/a18599bf2000947021b6f5269690aea291f67d85)

Increases readability as well, thanks for pointing out

PR #1329, opened 2025-05-08
HNSW Genesis :: Fix regression

siajasl:
The outer loop of the HNSW Genesis Indexer utilizes a BatchGenerator to push into the inner loop batches of Irises for indexation.  This PR:

- Fixes an issues with fetching a batch of Iris data from the database;
- Adds functional tests over the fetcher and the generator itself.
- Improves logging.
- Adds random Iris serial identifiers for exclusion in genesis indexation testing.


iris-mpc-cpu/bin/generate_serial_ids_for_deletion.rs
```diff
@@ -0,0 +1,55 @@
+use clap::Parser;
```
danielle-tfh, line 1: Since this is just for testing - should we rather make it a bash script? would not want to polute the code? Also this way we can remove genesis more easily once we no migrate to HNSW if its more confined?

siajasl: That's fine with me.

danielle-tfh (approved):

bgillesp (approved):
Looking good to me, thanks!  I wasn't quite able to hunt down, where exactly was the issue you fixed, related to SQL interpolation?

PR #1332, opened 2025-05-08
aurel/filter-intra: Do not include intra-batch results in match_ids

naure:
This PR is about whether intra-batch matches are included in a given field.

AFAICT the GPU implementation:
- returns `match_ids: match_ids_filtered`, where `filtered` means no-intra-batch.
- returns the partial match lists without filtering. I’m not sure whether there was intra-batch matches in it in the first place though.

This includes also https://github.com/worldcoin/iris-mpc/pull/1333


iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -843,27 +845,32 @@ impl HawkResult {
         let match_ids = self.select_indices(Filter {
             eyes: Both,
             orient: Only(Normal),
+            intra_batch: false,
         });
 
         let (partial_match_ids_left, partial_match_counters_left) = self.select(Filter {
             eyes: Only(Left),
             orient: Only(Normal),
+            intra_batch: true,
```
dkales, line 854: AFAIK, the batch_matches are not put into the partial match ids/counters:
https://github.com/worldcoin/iris-mpc/blob/e947d73284643df2c9c32d5e6717103326771697/iris-mpc-gpu/src/dot/kernel.cu#L258-L294 has no reference to these passed in compared to the above `merge*` functions.

dkales: This of course also follows for right and mirrored variants of partial matches.

dkales (changes requested):

iris-mpc-cpu/src/execution/hawk_main.rs
naure: Thanks for the info, done!

dkales (approved):

PR #1333, opened 2025-05-08
aurel/fix-intra-logic: Emulate the behavior of inserting entries one-by-one

naure:


iris-mpc-cpu/src/execution/hawk_main/matching.rs
```diff
@@ -179,8 +179,30 @@ impl BatchStep2 {
...
+                .select(Filter {
+                    eyes: Both,
+                    orient: Both,
+                    intra_batch: true,
+                })
+                .any(|id| match id {
+                    MatchId::Search(_) => true,
+                    MatchId::Luc(_) => true,
+                    MatchId::IntraBatch(request_i) => {
+                        let other_is_unique = !is_matches.get(request_i).unwrap_or(&true);
```
dkales, line 199: I find this a bit confusing because you need to think about if the ! or the unwrap or is more strongly binding.
Maybe spelling out all 3 possibilities is easier to follow?:

iliailia (approved):

iris-mpc-cpu/src/execution/hawk_main/matching.rs
dkales:
```suggestion
                        match is_matches.get(request_i) {
                            Some(true) => false, // the request we matched with was before us in the batch, and was a match, so we are not blocked by this intra-batch match
                            Some(false) => true, // the request we matched with was before us in the batch, and was unique, so we are blocked by this intra-batch match
                            None => false, // The request we matched with is after us in the batch, so we are not blocked by it.
                        }
```

dkales (approved):

iris-mpc-cpu/src/execution/hawk_main/matching.rs
naure: Thanks!

PR #1287, opened 2025-04-29
hawk genesis indexer: `fetch_height_of_indexed`

einar-inversed3:

PR #1334, opened 2025-05-08
Endpoint `height-of-graph-genesis-indexation`

einar-inversed3:
Features

1. ~~Persist latest processed iris index to disk.~~
2. Endpoint `/height-of-graph-genesis-indexation`.
3. ~~Consensus check among nodes.~~

supercedes #1287


iris-mpc-common/src/helpers/fetch_index.rs
```diff
@@ -0,0 +1,22 @@
+use eyre::{ensure, Result};
+use std::path::Path;
+use tokio::fs;
+
+/// The name of the file that stores the previous iris index as a string.
+pub const PREV_IRIS_INDEX_FILE: &str = "prev_iris_index.txt";
+
+pub type IrisSerialId = u64;
+
+pub async fn fetch_height_of_indexed() -> IrisSerialId {
```
danielle-tfh, line 10: @bgillesp can we rather create this in a database rather than on disk? for multiple deployments this would not be saved.

bgillesp: Yes, I'll take a look at setting this up -- I guess it probably makes sense to make a new table with genesis-specific intermediate state?

iris-mpc-common/src/server_coordination.rs
```diff
@@ -184,6 +200,72 @@ pub async fn wait_for_others_unready(config: &Config) -> Result<()> {
     Ok(())
 }
 
+/// Assumption This function assumes that each of the nodes have finished indexing the irises before it is called.
+pub async fn check_consensus_on_iris_height(config: &Config) -> Result<()> {
```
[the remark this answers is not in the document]
bgillesp: Agreed, sync state would probably be a good place to do this check, especially since this logically is data that is meant to be coordinated just at startup.  Might be easy to just make the `my_state` input to `start_coordination_server` slightly generic so that the genesis server can provide data specific to that setting?

iris-mpc-common/src/helpers/fetch_index.rs
bgillesp: Just pushed a commit which implements a new `persistent_state` table which provides a generic key-value store to keep track of any data we want.  Will integrate this with the height implementation tomorrow morning.

iris-mpc-store/src/lib.rs
```diff
@@ -566,6 +567,88 @@ WHERE id = $1;
...
+        } else {
+            Ok(None)
+        }
+    }
+
+    /// Set the entry in the `persistent_state` table for primary key `(domain, key)`.
+    ///
+    /// If an entry already exists for this primary key, this overwrites the existing
+    /// value with the value specified here.
+    pub async fn update_persistent_state<T: Serialize>(
```
einar-inversed3, line 604:
```suggestion
    pub async fn upsert_persistent_state<T: Serialize>(
```
@bgillesp which name do you prefer?

bgillesp: Good point that `update` implies an existing entry is being modified.  `upsert` is quite correct in terms of functionality but I fear it's a bit less legible -- could we just go with something simple like `set_persistent_state`?

iris-mpc-common/src/server_coordination.rs
```diff
@@ -105,7 +106,32 @@ pub async fn start_coordination_server(
...
+                )
+                .route(
+                    "/height-of-graph-genesis-indexation",
+                    get({
+                        let my_state = Arc::clone(&my_state);
+                        let height: IrisSerialId = (my_state.genesis_config.as_ref().map(
+                            |GenesisConfig {
+                                 max_indexation_height: _,
+                                 last_indexation_height,
+                             }| *last_indexation_height,
```
einar-inversed3, line 122: @siajasl is it `max_` or `last_`?

siajasl: It is `last_`.  The process needs to know to which height itself and the other nodes have indexed.

iris-mpc-cpu/src/genesis/utils/fetcher.rs
siajasl, line 1: With the `set_height_of_indexed` function, the fetcher component is now also mutating the state of the system.  Therefore we could rename the fetcher component to something like state_accessor.  This comment doesn't block any merge.

siajasl (approved):
LGTM.  Let's get it merged and into testing.

iris-mpc-common/src/server_coordination.rs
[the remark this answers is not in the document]
einar-inversed3: Removed.

iris-mpc-store/src/lib.rs
einar-inversed3: Yes. Done.

iris-mpc-common/src/helpers/fetch_index.rs
einar-inversed3: Thanks!

iris-mpc-common/src/server_coordination.rs
```diff
@@ -105,7 +106,32 @@ pub async fn start_coordination_server(
                 )
                 .route(
                     "/startup-sync",
-                    get(move || async move { serde_json::to_string(&my_state).unwrap() }),
+                    get({
```
danielle-tfh, line 109: https://github.com/worldcoin/iris-mpc/commit/d6e026bf61955caf64cbc734415721f61602a641 I think this should be handled here. Mind pulling the latest commit so we can remove this

einar-inversed3: Done.

danielle-tfh: Hey @einar-inversed3 I think we can remove this entirely from the code itself - It is not needed since we use the sync state.

danielle-tfh: This will also affect the HSNW startup process so lets rather remove it

PR #1337, opened 2025-05-09
aurel/reauth: Support reauth requests

naure:
- [x] Send the `reauth_target` given in requests through the `matching` pipeline. That will include this vector in the comparison work (`missing_vector_ids -> calculate_missing_is_match`), and the result will come back to `step2`.
- [x] Extend the uniqueness logic (`is_match`) to the new `Decision` type that covers the reauth cases.
- [x] Return `successful_reauths`.
- [x] Update the irises in the iris stores (as opposed to previously append-only).
- [x] Updating the graph is the same process as insertion. The difference is in the `vector_id / version_id`.
- [ ] Purging old versions from the graph is left for another PR.


iliailia (approved):
LGTM!

PR #1340, opened 2025-05-09
skip cpu tests

carlomazzaferro:


bgillesp (approved):
Works for now, thanks!  Will try to help hunt down what's going on with this test hanging.

PR #1342, opened 2025-05-12
add mirrored shares to batch query

carlomazzaferro:
It seems like for the mirrored case we miss adding the mirrored shares to the `BatchQuery`. While we implement the logic to parse them in `from HawkRequest`, we do not add them which leads to the error described here: https://tfh.slack.com/archives/C08FBFZSTHP/p1746809553497219?thread_ts=1746807504.843829&cid=C08FBFZSTHP


dkales (approved):

dkales (approved):
LGTM, seems this is an issue due to the separation of the GPU and CPU version of the receive_batch functionality, since the GPU version had added this, while the CPU one did not have it.

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -1328,7 +1328,13 @@ mod tests {
     fn receive_batch_shares(shares: Vec<GaloisRingSharedIris>) -> [IrisQueryBatchEntries; 4] {
         let mut out = [(); 4].map(|_| IrisQueryBatchEntries::default());
         for share in shares {
-            let one = preprocess_iris_message_shares(share.code, share.mask).unwrap();
+            let one = preprocess_iris_message_shares(
+                share.code.clone(),
+                share.mask.clone(),
+                share.code,
+                share.mask,
+            )
```
dkales, line 1336: I think in the test it does not matter much, but maybe we should also do this in a "correct" way here and address the TODO below already?

carlomazzaferro: Good call, yes will incorporate this

iris-mpc-cpu/src/protocol/shared_iris.rs
```diff
@@ -50,26 +50,52 @@ impl GaloisRingSharedIris {
...
-            GaloisRingSharedIris {
-                code: code_shares[1].clone(),
-                mask: GaloisRingTrimmedMaskCodeShare::from(&mask_shares[1]),
-            },
-            GaloisRingSharedIris {
-                code: code_shares[2].clone(),
-                mask: GaloisRingTrimmedMaskCodeShare::from(&mask_shares[2]),
-            },
-        ]
+        let code_shares_mirrored = code_shares
```
carlomazzaferro, line 61: @dkales I don't particularly love this change, as it requires changes to a lot of caller code. Do you think there's a way around it, or a simple refactor to add a more descriptive type as the output of this function makes more sense?

dkales: I guess @naure should have the say here on if this is needed/useful in the test.

naure: Alternatively, this function could stay unchanged, and move the conversion to mirrored version in a separate function.

naure (approved):

iris-mpc-cpu/src/protocol/shared_iris.rs
carlomazzaferro: @naure I have adopted your suggestion, it is indeed quite cleaner!

bgillesp (commented):
A bit behind on the merge schedule for this, but looks good!  Thanks again for hunting down the issue.

PR #1345, opened 2025-05-12
feat: multiplex sessions over a gRPC stream

sdwoodbury:
This change makes it easy to batch gRPC messages together. This is reduces networking overhead. 

## Session vs Stream
- The networking code now has to distinguish between a gRPC stream and a session. Previously, each stream was associated with a single session and the streams were identified by the `SessionId`. Now, when a gRPC stream is created, it is given a `StreamId` and will be associated with the following range of session ids: [ `stream_parallelism` * `StreamId`, `stream_parallelism` * (`StreamId` + 1) ). 
- A `StreamId` type was created to help prevent confusion. The type is identical to the `SessionId` type. 

## Benchmarks
- `benches/networking.rs` was created to benchmark varying number jobs, request parallelism, stream parallelism, and connection parallelism. Each job consists of a `batch_signed_lift_vec()` and a `cross_compare()`, which is equivalent to an atomic operation in the MPC protocol. 
- Some code is shared with the `hnsw` benchmark and the `bench_utils.rs` file was created to contain the shared code. 
- The benchmark can be run with an optional delay before each message is sent over the network; this would simulate latency of a wide area network. The `networking_benchmark` feature will enable this delay. ex: `cargo bench --bench networking --features networking_benchmark`. 

## Configuration parameters
- a parameter for stream parallelism was added to the various configuration structs. the default is 8. Previously, `connection_parallelism` gRPC connections were opened and sessions were assigned to them round-robin. Now, after the gRPC connections are opened, streams are assigned round robin to the gRPC connections but `stream_parallelism` sessions are added to a stream before opening a new stream. 
- `iris-mpc-cpu/src/execution/local.rs` was modified so that the gRPC connection parallelism, stream parallelism, and request parallelism could be specified by the benchmarks/unit tests. The default for each parameter is still 1. This modification was needed for the `networking` benchmark. 

## Protocol buffer changes
- The server now uses a `SendRequests` message, which is a vec of `SendRequest`. 
- A 4 byte session id has been added to the `SendRequest` message. The Rust code was updated to treat `SessionId` as a `u32`. 

## Session Multiplexing
- `grpc/networking/mod.rs` handles inbound sessions in the `start_message_stream()` function. The gRPC callback: `GrpcHandle::start_message_stream()` creates a single pair of channels for the stream. Then, for each session, another pair of channels is created. The receiving side is inserted into an `inbound_sessions` `HashMap` of `<session_id, OutStream>`. The sending side is inserted into a `inbound_forwarder` `HashMap` of `<session_id, InStream>`. A task is spawned to read from the stream channel and forward the messages to the appropriate session channels. The `GrpcSession` struct is used the same way as before - it is now created with a channel from `inbound_sessions`. 
- `grpc/networking/stream_manager.rs` handles outbound sessions. The `StreamManager::add_session()` function handles the process of creating a new gRPC stream and returning the `OutStreams` to the caller. 

## logging
some redundant bits of information were removed to help me read the logs faster. 
- Some logs were removed because an event was reported at multiple levels of the call stack and was redundant. 
- logging of Identity and SessionId was updated to exclude "player" and "session" so that the log would show "Identity(x)" instead of "player Identity(x)" and "SessionId(x)" instead of "session SessionId(x)". This required the assertions in the grpc `comms_fail` unit test to be updated, but the unit test functionality was unchanged.


iris-mpc-cpu/src/proto/party_node.proto
```diff
@@ -3,11 +3,16 @@ syntax = "proto3";
 package party_node;
 
 service PartyNode {
-    rpc StartMessageStream (stream SendRequest) returns (SendResponse);
+    rpc StartMessageStream (stream SendRequests) returns (SendResponse);
 }
 
 message SendRequest {
-    bytes data = 1;
+    uint32 session_id = 1;
```
iliailia, line 10: Btw, do you have an idea how this increased the total throughput?

sdwoodbury: i just ran a test in a branch where i used a histogram to record the number of messages received at once due to stream-multiplexing. 

with connection_parallelism and stream_parallelism set to 8, i'm seeing an average batch size of 6 messages, with a standard deviation of 2 (max possible is 8).  at the same time, i also saw 578,644  rounds of communication, sending a total of 51,197,421 bytes at the application level. 

if the networking headers add 40 bytes per message and 489,000 fewer messages are sent, that's approximately 18.7 MiB saved out of 70.9 MiB total (application data + 40 bytes per message)  - a ~26% reduction in bandwidth. 

i'm unsure about the actual throughput improvement because I don't know when the system becomes saturated. 

I do have some data from a contrived scenario where each job had only 17 rounds of communication and network latency was 30ms. 
- The parameters were: request parallelism - 65536, connection_parallelism - 8, and stream_parallelism - 8. 
- the results are [here](https://www.notion.so/inversed/Potential-gRPC-Optimizations-1e6a3b540bfe80d2b249f83f3469305c)
-  I saw about a 1.65x throughput improvement.

iliailia (approved):
Thanks, neat and clean implementation!

PR #1346, opened 2025-05-12
feat: use binary serialiation for the graph db

sdwoodbury:
The tables `hawk_graph_links` and `hawk_graph_entry` were using a `jsonb` type for `links` and `entry_point`. This PR changes that format to `bytea` and adds custom serialization and deserialization for those columns. 

The `VectorId` struct now requires 6 bytes instead of  21 when serialized. The `SortedEdgeIds` struct now requires 8 bytes instead of 2 (when empty). 

## WARNING
the migration will wipe the graph database. before running this migration, backup the database using the `graph-mem-cli` tool from the `iris-mpc-cpu` crate. After running the migration, restore the database using the `graph-mem-cli` tool. 

## Changes
- created the file `iris-mpc-common/src/serialization.rs` and added two traits: `ReadPacked` and `WritePacked`
- implemented these traits for `VectorId`, `SortedEdgeIds`, and `EntryPoint`. A convenience function was added to `SortedEdgeIds`, and `EntryPoint` to serialize the structs - eliminating the need for the user to create a vector and pass it to `write_packed()`. 
- changed the `GraphOps` implementation to require that `V::VectorRef` implement `ReadPacked` and `WritePacked`
- added a migration script. wasn't sure what the last 6 digits should be. assumed the start of the file name was YYYMMDD.


iris-mpc-common/src/serialization.rs
```diff
@@ -0,0 +1,13 @@
...
+use eyre::Result;
+use std::io::{Read, Write};
+
+pub trait WritePacked {
+    fn write_packed<W: Write>(&self, writer: &mut W) -> Result<()>;
+}
+
+pub trait ReadPacked: Sized {
+    fn read_packed<R: Read>(reader: &mut R) -> Result<Self>;
+}
```
bgillesp, line 13: General question -- why did you decide to roll a custom serialization implementation rather than just using a library like `bincode`?  Are there some advantages we can get because of the restricted context?

sdwoodbury: I just looked into bincode and it appears to be just as efficient as this custom serialization. I will switch to bincode.

sdwoodbury: i made the changes and double checked the serialized sizes. it's ready for review again.  the higher ranked trait bound:  `for<'a> Deserialize<'a>` looks strange to me but i think it's needed.

PR #1348, opened 2025-05-13
only consider one matching rotation for bucket collection

philsippl:


philsippl:
ah right, once again this issue

philsippl:
I think keeping around a mapping of index → matching min rotation shouldn't be too bad, since we operate on chunks of the db at this part?
But still some work, let's wait first what comes out of the analysis if this is an actual issue (to have multiple matching rotations).

leonanos8:
We should close this one right?

PR #1349, opened 2025-05-13
[Genesis] Add syncing of genesis config across nodes

danielle-tfh:
### Notes
* add check of syncing for genesis (re-use the already existing sync state)
* add checking for valid parameters

### DIFF
--- a/iris-mpc-common/src/helpers/sync.rs
+++ b/iris-mpc-common/src/helpers/sync.rs
@@ -149,15 +149,7 @@ impl SyncResult {
             return Err(eyre::eyre!("Genesis config is None for local state"));
         }
         for state in &self.all_states {
-            if state.genesis_config != genesis_config {
-                return Err(eyre::eyre!(
-                    "Inconsistent genesis config!\n
-                have: {:?}\n
-                got: {:?}",
-                    genesis_config,
-                    state.genesis_config
-                ));
-            }
+            ensure!(state.genesis_config == genesis_config, "Inconsistent genesis config");
         }
         Ok(())
     }
--- a/iris-mpc-common/src/helpers/sync.rs
+++ b/iris-mpc-common/src/helpers/sync.rs
@@ -919,7 +919,7 @@ mod tests {
     }
 
     #[test]
-    fn test_check_genesis_config_all_none() {
+    fn test_check_genesis_config_one_none() {
         let genesis_config = Some(GenesisConfig {
             max_indexation_height: 100,
             last_indexation_height: 50,
@@ -956,7 +956,7 @@ mod tests {
     }
 
     #[test]
-    fn test_check_genesis_config_one_none() {
+    fn test_check_genesis_config_all_none() {
         let states = vec![
             SyncState {
                 db_len: 10,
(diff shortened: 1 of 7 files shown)


iris-mpc-common/src/helpers/sync.rs
```diff
@@ -12,6 +18,8 @@ pub struct SyncState {
     pub modifications: Vec<Modification>,
     pub next_sns_sequence_num: Option<u128>,
     pub common_config: CommonConfig,
+    // Genesis Sync State Config
+    pub genesis_config: Option<GenesisConfig>,
```
einar-inversed3, line 22:
```suggestion
    pub genesis_sync_state_config: Option<GenesisConfig>,
```

iris-mpc-common/src/helpers/sync.rs
```diff
@@ -134,6 +142,26 @@ impl SyncResult {
...
+        for state in &self.all_states {
+            if state.genesis_config != genesis_config {
+                return Err(eyre::eyre!(
+                    "Inconsistent genesis config!\n
+                have: {:?}\n
+                got: {:?}",
+                    genesis_config,
+                    state.genesis_config
+                ));
+            }
```
einar-inversed3, line 160:
```suggestion
            ensure!(state.genesis_config == genesis_config, "Inconsistent genesis config");
```

iris-mpc-upgrade-hawk/bin/iris_mpc_hawk_genesis.rs
```diff
@@ -22,6 +22,13 @@ async fn main() -> Result<()> {
     let args = Args::parse();
     let max_indexation_height = args.max_indexation_height;
 
+    // How can i throw an error here if the max_indexation_height is not set?
+    if max_indexation_height.is_none() {
+        eprintln!("Error: --max-height argument is required.");
+        return Err(eyre::eyre!("--max-height argument is required."));
```
einar-inversed3, line 28:
```suggestion
        bail!("--max-height argument is required.");
```

iris-mpc-common/src/helpers/sync.rs
```diff
@@ -807,13 +838,156 @@ mod tests {
...
+                genesis_config: genesis_config_1.clone(),
+            },
+        ];
+
+        let sync_result = SyncResult::new(states[0].clone(), states);
+        assert!(sync_result.check_genesis_config().is_err());
+    }
+
+    #[test]
+    fn test_check_genesis_config_all_none() {
```
carlomazzaferro, line 922: A but misleading name here, `all_none` but only one of them is `None`. Is this intentional? seems swapped with the next test case

iris-mpc-upgrade-hawk/bin/iris_mpc_hawk_genesis.rs
```diff
@@ -22,6 +22,12 @@ async fn main() -> Result<()> {
     let args = Args::parse();
     let max_indexation_height = args.max_indexation_height;
 
+    if max_indexation_height.is_none() {
+        eprintln!("Error: --max-height argument is required.");
+        bail!("--max-height argument is required.");
+    }
+    let max_indexation_height = max_indexation_height.unwrap();
```
carlomazzaferro, line 29: Would it be wise to let this be None, and if so, index until the highest serial id? Would that make sense?

danielle-tfh: I think not - since the nodes could have different serial ids based on when they start and where the MPCv2 deployment is at.

I would also rather specify this for snapshots as well

carlomazzaferro (approved):

PR #1356, opened 2025-05-15
Implement genesis graph insertion logic

bgillesp:
PR implements handling of processing of batches of iris codes in the genesis indexer, for insertion of iris codes into the in-memory graph stores.  Main changes:
- Implements in-memory graph search and insertion for genesis indexer batches, using fixed parallelism specified by `args.request_parallelism` parameter
- Adds simplified `HawkActor` search function
- Refactors `HawkActor` insert functionality, moved to separate `insert` submodule
- Refactors `Aby3Store` insertion functions to provide a more uniform interface

One last part of the pipeline still needs to be implemented: persisting the results to the database.  We should be able to directly use `HawkMutation::persist` for this with the data returned by the batch processing, but we still need to get the batch processing output back to the main genesis server execution loop to bring it in contact with the database connection.


siajasl (approved):
LGTM.  I suspect a small degree of refactoring is worthwhile once full end to end workflow is implemented.

naure:
Very nice!

PR #1364, opened 2025-05-16
Improve usage of last indexed height

danielle-tfh:
### Notes
* follow ups from https://github.com/worldcoin/iris-mpc/pull/1334/files
* remove unused endpoint
* retrieve the height only once and use it for syncing state across nodes


bgillesp (approved):
Thanks for the fixes Danielle!  This all looks aligned with what we're aiming for.

PR #1365, opened 2025-05-17
Decouple standard and genesis sync state

bgillesp:
Sync state has previously been shared between standard and genesis binaries, even though the initial state being synchronized is somewhat distinct between the modes.  This PR decouples the sync state into two separate types in independent submodules.  Additionally, the list of iris code serial ids excluded from indexing is added to the genesis sync state.


siajasl (approved):
LGTM

PR #1366, opened 2025-05-17
[POP-2506] Alternate full scan side per batch

philsippl:
FYI: all done by codex from my phone

## Summary
- alternate the full scan side after every batch

## Testing
- `cargo fmt --all` *(fails: could not download file)*
- `cargo test --workspace -- --test-threads=1` *(fails: could not download file)*

### DIFF
--- a/iris-mpc-gpu/src/server/actor.rs
+++ b/iris-mpc-gpu/src/server/actor.rs
@@ -965,7 +965,10 @@ impl ServerActor {
         ///////////////////////////////////////////////////////////////////
         tracing::info!("Fetching partial {} results", self.full_scan_side);
         let mut partial_matches_side1 = self.distance_comparator.get_partial_results(
-            &self.db_match_list_left,
+            match self.full_scan_side {
+                Eye::Left => &self.db_match_list_left,
+                Eye::Right => &self.db_match_list_right,
+            },
             &self.current_db_sizes,
             &self.streams[0],
         );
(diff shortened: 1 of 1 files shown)


eaypek-tfh (approved):

eaypek-tfh:
PR from the phone? Sounds awesome! 😄 

Shall we ask our friend to enrich the result message with full batch side so that the AI team can have more insight when they analyze stuff?

dkales (approved):
LGTM

leonanos8:
@eaypek-tfh To my understanding AI team only care about the anon stats, and with this change we are going to produce stats for both full scan sides.

> the result message with full batch side so that the AI team can have more insight when they analyze stuff?

Do they also take a look in the results?

dkales:
@leonanos8 Nice catch on the leftover hard-coded side value!

eaypek-tfh:
> @eaypek-tfh To my understanding AI team only care about the anon stats, and with this change we are going to produce stats for both full scan sides.
> 
> > the result message with full batch side so that the AI team can have more insight when they analyze stuff?
> 
> Do they also take a look in the results?

synced in private, this info could be useful to them. we can check with them

leonanos8:
@eaypek-tfh I will follow up with the team. Would not block this PR  tho, could be added as a follow-up.

@dkales @philsippl I am proceeding with performing a correctness test in stage for this one and will let you know

leonanos8 (approved):
Lets proceed, tested this with bulk enrollments in stage

PR #1367, opened 2025-05-18
Implement database persistence for genesis indexer

bgillesp:
This PR implements database persistence for the genesis indexer, following the pattern used by the main server binary of running a background "results processing" task which receives results passed in by the main server loop.  The PR additionally does some refactoring on adjacent parts of the genesis library functionality.


siajasl (approved):
LGTM

PR #1373, opened 2025-05-19
Add mechanism to change HSNW schema for different testing

danielle-tfh:
### Notes
* remove all Genesis config from the main server config (no need to keep genesis data there)
* add mechanism to alter the HNSW schema for running different tests ie: we can run genesis on a different schema then the  normal HNSW one.
* change app name to its actual name `schema_name`

### DIFF
--- a/iris-mpc-common/src/config/mod.rs
+++ b/iris-mpc-common/src/config/mod.rs
@@ -21,8 +21,8 @@ pub struct Opt {
 #[allow(non_snake_case)]
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Config {
-    #[serde(default = "default_app_name")]
-    pub app_name: String,
+    #[serde(default = "default_schema_name")]
+    pub schema_name: String,
 
     #[serde(default)]
     pub environment: String,
@@ -189,6 +189,9 @@ pub struct Config {
     #[serde(default)]
     pub enable_reset: bool,
 
+    #[serde(default)]
+    pub hnsw_schema_name_suffix: String,
+
     #[serde(default = "default_hawk_request_parallelism")]
     pub hawk_request_parallelism: usize,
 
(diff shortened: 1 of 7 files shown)


iris-mpc-common/src/config/mod.rs
```diff
@@ -247,28 +250,6 @@ pub struct Config {
     pub override_max_batch_size: bool,
 }
 
-impl Config {
```
danielle-tfh, line 250: Removed since only Genesis uses them

iris-mpc/src/server/mod.rs
```diff
@@ -260,29 +272,41 @@ async fn prepare_stores(config: &Config) -> Result<(Store, GraphPg<Aby3Store>),
             Ok((iris_store, graph_store))
         }
 
-        // use the base db for both stores
-        _ => {
+        ModeOfDeployment::Standard => {
```
danielle-tfh, line 275: Updated this to use the Standard one - this should have readWrite on Both

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -342,6 +342,15 @@ async fn get_service_clients(
...
     async fn get_pgres_clients(config: &Config) -> Result<(IrisStore, GraphPg<Aby3Store>), Report> {
+        let iris_schema_name = format!(
+            "{}_{}_{}",
+            config.schema_name, config.environment, config.party_id
+        );
+
+        let hawk_schema_name = format!(
+            "{}{}_{}_{}",
+            config.schema_name, config.hnsw_schema_name_suffix, config.environment, config.party_id
+        );
```
eaypek-tfh, line 353: [nit] I'm seeing schema being formatted in two places. Shall we maybe move this to the common store crate so that we have a single source of truth?

eaypek-tfh (approved):
Nice way to keep things working as is if nothing is passed to the hnsw schema suffix. I liked it! 👍

iris-mpc-upgrade-hawk/src/genesis/mod.rs
danielle-tfh: Yes - these methods can be combined but it needs a greater refactor ie: Both stores should live in common-store otherwise it becomes a circular dependency.

eaypek-tfh: makes sense. let's leave it for now then 👍

PR #1376, opened 2025-05-20
(feature): add test tool for importing and exporting the HNSW graph

sdwoodbury:
This tool was created in anticipation of the HNSW schema change, which will change the format of the graph edges from Json to binary.

There is currently not a good way to perform a migration that involves running custom code. A more powerful tool such as `sqlx_migrator` or `sea-orm` would allow that, but that is out of scope for the schema change PR. 

This PR allows for bypassing the migration step by going directly from `GraphMem` to/from a file, serialized/deserialized using `bincode`. 

## Intended Usage
- first, connect to the desired database and run the `backup-db` command
- next, check out a future version of `iris-mpc-cpu`, which will include a migration that drops the HNSW database tables and re-creates them with columns of the desired format. 
- finally, run the `restore-db` command. this will load the database from the file and persist it in postgres. 

## additional feature
the `compare-to-db` command was added to compare a file (generated with `backup-db`) to the database. 

## testing
Testing was performed using the  `docker-compose.test.yaml` file.

command used to populate the local database
`target/debug/graph-mem-cli --db-url postgresql://postgres:postgres@localhost:5432/SMPC_dev_2 --schema SMPC_dev_2 --file /tmp/db_loader store-random`

command used to test the dump/load functionality
`target/debug/graph-mem-cli --db-url postgresql://postgres:postgres@localhost:5432/SMPC_dev_2 --schema SMPC_dev_2 --file /tmp/db_loader verify-backup`

### DIFF
--- a/iris-mpc-cpu/bin/graph_mem_cli.rs
+++ b/iris-mpc-cpu/bin/graph_mem_cli.rs
@@ -47,7 +47,7 @@ async fn main() -> Result<()> {
         command,
     } = cli;
 
-    let mut db_context = DbContext::new(&db_url, &schema).await;
+    let db_context = DbContext::new(&db_url, &schema).await;
 
     match command {
         Command::DumpDb => {
--- a/iris-mpc-cpu/bin/graph_mem_cli.rs
+++ b/iris-mpc-cpu/bin/graph_mem_cli.rs
@@ -1,0 +1,59 @@
+use clap::{Parser, Subcommand};
+use iris_mpc_cpu::hnsw::graph::test_utils::DbContext;
+use std::path::PathBuf;
+use eyre::Result;
+
+#[derive(Parser)]
+#[command(name = "graph_mem_cli")]
+#[command(about = "Convert from db -> memory -> file or file -> memory -> db", long_about = None)]
+struct Cli {
+    /// Database URL
+    #[arg(long)]
+    db_url: String,
+    /// Database schema
+    #[arg(long)]
+    schema: String,
+    /// File path (input or output depending on mode)
+    #[arg(long)]
+    file: PathBuf,
+    #[command(subcommand)]
+    command: Command,
+}
+
+#[derive(Subcommand)]
+enum Command {
+    /// Load a graph from a database to memory, then write it to a file.
+    DumpDb,
+    /// Load a graph from a file to memory, then write it to a database
+    LoadDb,
+    /// verify that Dump/Load works as expected
+    Test,
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let cli = Cli::parse();
+
+    let Cli {
+        db_url,
+        schema,
+        file,
+        command,
+    } = cli;
+
+    let db_context = DbContext::new(&db_url, &schema).await;
+
+    match command {
+        Command::DumpDb => {
+            db_context.write_graph_to_file(&file).await?;
+        }
+        Command::LoadDb => {
+            db_context.load_graph_from_file(&file).await?;
+        }
+        Command::Test => {
+            db_context.test_load_store(&file).await?;
+        }
+    }
+    println!("Command succeeded");
+    Ok(())
+}
--- a/iris-mpc-cpu/bin/graph_mem_cli.rs
+++ b/iris-mpc-cpu/bin/graph_mem_cli.rs
@@ -26,13 +26,15 @@ struct Cli {
 #[derive(Subcommand)]
 enum Command {
     /// Load a graph from a database to memory, then write it to a file.
-    DumpDb,
+    BackupDb,
     /// (testing only) creates random data, stores it in the database, and then runs dump-db.
-    DumpRandom,
+    StoreRandom,
     /// Load a graph from a file to memory, then write it to a database
-    LoadDb,
-    /// (testing only) verify that Dump/Load works as expected
-    Test,
+    RestoreDb,
+    /// (testing only) verify that Load/Store works as expected
+    VerifyRestore,
+    /// Load a graph from a file and compare it to the graph stored in the database.
+    CompareToDb,
 }
 
 #[tokio::main]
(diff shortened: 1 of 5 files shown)


iris-mpc-cpu/src/hnsw/graph/test_utils.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+impl DbContext {
+    pub async fn new(url: &str, schema: &str) -> Self {
+        let client = PostgresClient::new(url, schema, AccessMode::ReadWrite)
+            .await
+            .unwrap();
+        let store = Store::new(&client).await.unwrap();
+        let graph_pg = GraphPg::new(&client).await.unwrap();
+        Self { store, graph_pg }
```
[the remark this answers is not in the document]
sdwoodbury: i removed most of the unwraps. The `unwrap()` on `new()` is acceptable because this is a test tool and the test can not continue if the connection fails.

iris-mpc-cpu/src/hnsw/graph/test_utils.rs
```diff
@@ -0,0 +1,247 @@
...
+        graph_ops.load_to_mem().await
+    }
+
+    // loads the graph from the database to memory and then writes it to a file
+    pub async fn write_graph_to_file(&self, path: &std::path::Path, dbg: bool) -> Result<()> {
+        let left = self.load_graph_to_mem(StoreId::Left).await?;
+        let right = self.load_graph_to_mem(StoreId::Right).await?;
+        let graph = BothSides { left, right };
+        if dbg {
+            println!("storing graph:");
```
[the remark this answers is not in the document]
sdwoodbury: the rest of the code does not use log or tracing. there is no point to using a logging framework in this case.

iris-mpc-cpu/src/hnsw/graph/test_utils.rs
```diff
@@ -0,0 +1,247 @@
...
+        if dbg {
+            println!("loaded graph:");
+            println!("{:#?}", graph);
+        }
+        self.persist_graph_db(graph.left, StoreId::Left).await?;
+        self.persist_graph_db(graph.right, StoreId::Right).await?;
+        Ok(())
+    }
+
+    pub async fn test_load_store(&self, path: &std::path::Path, dbg: bool) -> Result<()> {
```
bgillesp, line 159: Can you add a quick doc comment for this function?

iris-mpc-cpu/src/hnsw/graph/test_utils.rs
```diff
@@ -0,0 +1,247 @@
...
+            println!("loaded graph:");
+            println!("{:#?}", loaded_graph);
+        }
+        if stored_graph != loaded_graph {
+            return Err(eyre::eyre!("Loaded graph does not match stored graph"));
+        }
+        Ok(())
+    }
+
+    // this function was stolen from other test code.
```
bgillesp, line 184: Can you extend this comment to describe the function?  I know it's for testing only, but since it's exposed to the `graph-mem-cli` binary, probably is worth saying a few words.

iris-mpc-cpu/src/hnsw/graph/test_utils.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub async fn gen_random(&self) -> Result<()> {
+        let rng = &mut AesRng::seed_from_u64(0_u64);
+        let mut vector_store = PlaintextStore::new();
+
+        let vectors = {
+            let mut v = vec![];
+            for raw_query in IrisDB::new_random_rng(10, rng).db {
```
bgillesp, line 191: Can you define a constant for this just above within the function?  Would make it more clear what the parameters are for the construction.

iris-mpc-cpu/src/hnsw/graph/test_utils.rs
```diff
@@ -0,0 +1,247 @@
...
+
+        graph_ops.set_entry_point(ep2.point, ep2.layer).await?;
+
+        let (point3, layer3) = graph_ops.get_entry_point().await?.unwrap();
+        let ep3 = EntryPoint {
+            point: point3,
+            layer: layer3,
+        };
+
+        assert_eq!(ep2, ep3);
```
bgillesp, line 226: In these lines, do we need the extra validation of the get and set methods for entry points?  (I'm assuming this is from the other test code this was sourced from.)  Maybe just keep the `ep2` and associated `set_entry_point` call.

iris-mpc-cpu/src/hnsw/graph/test_utils.rs
```diff
@@ -0,0 +1,247 @@
...
+
+        for i in 1..4 {
```
bgillesp, line 228: Can you add a comment on what this loop is doing?  The `1..4` and `4..7` are a bit mysterious on a first glance, especially since 3 is the magic number for the number of participants in the MPC, though unrelated to what you're doing here.

iris-mpc-cpu/bin/graph_mem_cli.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Subcommand)]
+enum Command {
+    /// Load a graph from a database to memory, then write it to a file.
+    DumpDb,
+    /// (testing only) creates random data, stores it in the database, and then runs dump-db.
+    DumpRandom,
+    /// Load a graph from a file to memory, then write it to a database
+    LoadDb,
+    /// (testing only) verify that Dump/Load works as expected
+    Test,
```
bgillesp, line 35: This subcommand is actually really useful -- we'll be wanting to compare a DB graph against a filesystem graph to check for consistency, where the graphs may be generated in different contexts.  Can you change the name to something like `compare` and give a descriptive doc comment?

bgillesp (approved):
Looks great, thank you!  Requesting one change in the CLI (the `test` subcommand is actually useful for comparing a db graph against a filesystem graph), and just a few small suggestions for clarity and codebase cohesion, but otherwise this looks like exactly what we need.

iris-mpc-cpu/bin/graph_mem_cli.rs
sdwoodbury: per slack i will add an additional command to compare the database against a file.

PR #1377, opened 2025-05-20
Fix/anon stats session

carlomazzaferro:
The main and only change that's relevant reviewing here is the filtering of the matches, so that we keep only the neighbors that actually matched for anonymized statistics reporting: https://github.com/worldcoin/iris-mpc/pull/1377/files#diff-334cb4ae1a3bea043540c0991c569ee4527398a51ee2eb6321ec987fcdebebcdR476-R488

### DIFF
--- a/iris-mpc-cpu/src/execution/hawk_main.rs
+++ b/iris-mpc-cpu/src/execution/hawk_main.rs
@@ -449,14 +449,12 @@ impl HawkActor {
 
     async fn update_anon_stats(
         &mut self,
-        session: &HawkSessionRef,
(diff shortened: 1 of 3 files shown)


iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -1546,3 +1550,6 @@ mod tests_db {
         Ok(())
     }
 }
+
```
carlomazzaferro, line 1553: Ignore, left over and will remove this

philsippl (approved):

naure (approved):

PR #1379, opened 2025-05-20
Add explicit ordering of session checks

danielle-tfh:
### Notes
* if this is done as a try_join messages can be confused between left and right. Therefore we do each one seperately


siajasl (approved):
LGTM

PR #1380, opened 2025-05-20
Update s3 importer to include version id

eaypek-tfh:
## Changes
- We want to use s3 import logic in HNSW too. Therefore, we added the `version_id` to the exporter format already.
- In this PR, we update the export binary parsing logic to take `version_id` into account as well

## Refactors
- There were three different implementations for database loading: GPU, HNSW, Genesis. This PR unifies all of them and moves all the s3 related logic into the iris-mpc-store crate so that everything is isolated there - including the specific s3 client creation.

## Tests
- [x] GPU imports work as before. (In prod, we need to update the s3 directory name when we release this version)
- [x] HNSW works with s3 imports

### DIFF
--- a/iris-mpc-upgrade-hawk/src/genesis/mod.rs
+++ b/iris-mpc-upgrade-hawk/src/genesis/mod.rs
@@ -6,9 +6,7 @@ use aws_sdk_s3::{
 use eyre::{bail, eyre, Report, Result};
 use iris_mpc_common::{
     config::{CommonConfig, Config, ModeOfCompute, ModeOfDeployment},
-    helpers::{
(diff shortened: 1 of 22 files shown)


iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -485,9 +490,16 @@ async fn init_graph_from_stores(
     //       to read into memory
     // -------------------------------------------------------------------
     let store_len = iris_store.count_irises().await?;
-    load_db(&mut iris_loader, iris_store, store_len, parallelism)
```
danielle-tfh, line 488: Nice! Thank you!

danielle-tfh (approved):

carlomazzaferro (approved):
Clean refactor 👍

PR #1381, opened 2025-05-20
aurel/reset-updates: Handle Reset Update requests

naure:
*On top of #1374.*

- [x] Find and parse the Reset Update requests (`batch.reset_update_*`). This is a different list than the other request types.
- [x] Update the iris stores.
- [x] Return a copy of the `reset_update_*` fields.
- [x] Insert into the graph: in-memory and return mutations.


iris-mpc-cpu/src/execution/hawk_main/reset.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+pub async fn search_to_reset(
+    hawk_actor: &mut HawkActor,
+    sessions: &BothEyes<Vec<HawkSessionRef>>,
+    request: &HawkRequest,
+) -> Result<ResetPlan> {
+    // Get the reset updates from the request.
+    let updates = {
+        let store = hawk_actor.iris_store[LEFT].read().await;
```
dkales, line 25: Triggered by quick confusion of only `LEFT` usage here, is there an actual need to have two `iris_store` Arcs for left/right in the `HawkActor`? They point to the same RwLocked IrisStore anyway, right?

dkales (approved):

iris-mpc-cpu/src/execution/hawk_main/reset.rs
naure: This is to obtain the full vector IDs from indices, which should be the same in both stores. The two stores are otherwise used independently.

naure: Added a comment.

PR #1382, opened 2025-05-20
parallel thresholding

carlomazzaferro:


iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -482,37 +482,98 @@ impl HawkActor {
...
+            .iter()
+            .cloned()
+            .zip(slices_data.into_iter())
+            .enumerate()
+            .map(|(i_session, (session_ref, slice))| {
+                let thresholds = translated_thresholds.clone();
+                let tx = tx.clone();
+                async move {
+                    let mut session = session_ref.write().await;
+                    let shares = compare_threshold_buckets(
```
naure, line 525: If `distances.len() < n_sessions` then sometimes `slice.len() = 0` which is not accepted by `compare_threshold_buckets`.

naure (changes requested):

naure:
Excellent! See possible bug above.

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -482,37 +482,98 @@ impl HawkActor {
...
+        let mut session0 = sessions[side][0].write().await;
+        // Open all per-session bucket shares and aggregate into final per-threshold counts
+        let opened = open_ring(&mut session0.aby3_store.session, &bucket_result_shares).await?;
+        let n_buckets = translated_thresholds.len();
+        let mut buckets = vec![0u32; n_buckets];
+        for chunk in opened.chunks(n_buckets) {
+            for (i, &count) in chunk.iter().enumerate() {
+                buckets[i] = buckets[i].saturating_add(count);
+            }
+        }
```
dkales, line 558: I guess this leaks a bit more than strictly needed, as we could add up the shares before opening them.

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -482,37 +482,98 @@ impl HawkActor {
...
+                    .await?;
+                    tx.send((i_session, shares))?;
+                    Ok(())
+                }
+            });
+
+        parallelize(tasks).await?;
+        drop(tx);
+
+        let mut shares_acc = vec![Vec::new(); n_sessions];
```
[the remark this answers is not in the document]
naure: It works with https://docs.rs/futures/latest/futures/future/fn.join_all.html

> … collecting the results into a destination Vec<T> in the same order as they were provided.

iris-mpc-cpu/src/execution/hawk_main.rs
dkales: something like (untested)
```

        let n_buckets = translated_thresholds.len();
        let mut share_buckets = vec![Share::<u32>::default(); n_buckets];
        while let Some((i_session, shares)) = rx.recv().await {
            for (share, bucket) in izip!(shares, share_buckets.iter_mut()) {
                *bucket += share;
            }
        }

        let mut session0 = sessions[side][0].write().await;
        // Open aggregated buckets
        let opened = open_ring(&mut session0.aby3_store.session, &share_buckets).await?;
        Ok(opened)
```

naure: Right. So instead of 

```rust
        let mut bucket_result_shares = Vec::new();
        for shares in shares_acc {
            bucket_result_shares.extend(shares);
        }
```

it could do something like (draft):

```rust
        let mut bucket_result_shares = vec![EMPTY_SHARE; n_bucket];
        for shares in shares_acc {
            for i in 0..n_buckets {
                bucket_result_shares[i] += shares[i];
            }
        }
```

This is similar to this line:

https://github.com/worldcoin/iris-mpc/blob/92083051af3586320f3764082db9b61491abcde2/iris-mpc-cpu/src/protocol/ops.rs#L139

iris-mpc-cpu/src/execution/hawk_main.rs
carlomazzaferro: Thanks, addressed 👍

iris-mpc-cpu/src/execution/hawk_main.rs
carlomazzaferro: Went with @naure 's suggestion here

iris-mpc-cpu/src/execution/hawk_main.rs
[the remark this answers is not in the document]
carlomazzaferro: Thanks, also implemented this - let me know if it looks correct

iris-mpc-common/src/helpers/statistics.rs
```diff
@@ -118,7 +118,18 @@ impl BucketStatistics {
 
             // The difference between buckets[i] and buckets[i - 1], except when i=0
             let previous_count = if i == 0 { 0 } else { buckets_array[i - 1] };
-            let count = buckets_array[i] - previous_count;
+            // Ensure non-decreasing cumulative counts to avoid underflow
+            let count = if buckets_array[i] >= previous_count {
+                buckets_array[i] - previous_count
+            } else {
+                tracing::warn!(
+                    "Non-monotonic cumulative bucket counts at index {}: current {}, previous {}; clamping to zero",
```
carlomazzaferro, line 126: I got randomly `"Panic occurred with message: attempt to subtract with overflow","backtrace" in iris-mpc-common/src/helpers/statistics.rs:121:25`

Added this guard here - but now cannot reproduce, so keeping it as warn so we can introspect later

naure (approved):

PR #1383, opened 2025-05-21
Wait for all batches to be completed before shutting down

danielle-tfh:
### Notes
* wait for all batches to be completed before shutting down genesis


bgillesp (approved):
Looks good, thanks for the fix!

PR #1391, opened 2025-05-21
Update to use batch configuration from CLI

danielle-tfh:
### Notes
* add batch size configuration to CLI input
* remove `db_len` field from `SyncState` struct since the nodes may be processing results and have difference states
* rename `check_genesis_config` to `check_synced_state` in `SyncResult since we check both

### DIFF
--- a/iris-mpc-cpu/src/genesis/state_sync.rs
+++ b/iris-mpc-cpu/src/genesis/state_sync.rs
@@ -8,12 +8,12 @@ pub struct Config {
     pub max_indexation_id: IrisSerialId,
     pub last_indexed_id: IrisSerialId,
     pub excluded_serial_ids: Vec<IrisSerialId>,
+    pub batch_size: usize,
 }
 
 /// Encapsulates a node's synchronization state.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct SyncState {
-    pub db_len: u64,
     pub common_config: CommonConfig,
     pub genesis_config: Config,
 }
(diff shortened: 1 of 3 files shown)


iris-mpc-cpu/src/genesis/state_sync.rs
```diff
@@ -37,30 +37,14 @@ impl SyncResult {
 /// Methods.
 impl SyncResult {
     /// Check if the common part of the config is the same across all nodes.
-    pub fn check_genesis_config(&self) -> Result<()> {
-        let genesis_config = self.my_state.genesis_config.clone();
+    pub fn check_synced_state(&self) -> Result<()> {
```
danielle-tfh, line 40: @siajasl while we work on the same code - lets make sure each of us review the others PR. There were some very specific reasons i did not add DB size  as a sync state(in PR description) and left the default for the max_height as empty (we should be very explicit with how long this runs for and that a bad state can cause issues across nodes)

danielle-tfh:
> This all is looking good -- definitely agree that `batch_size` should be surfaced as a CLI argument since the basic functioning is sensitive to this parameter.
> 
> One question I have is about removing `db_len` from the genesis `Config` struct -- I've been assuming that the eventual production usage of this binary will involve starting with a fixed static snapshot of the existing `irises` table that the binary will chew on, while the production nodes continue to process in the main production databases while saving all queries and results for the duration of the initial graph indexing so that an HNSW server client can catch up from the genesis results to reach the head of current production state.
> 
> In particular, the genesis binary isn't currently set up to handle in-place iris modifications that happen as a result of reauth requests. If using a static database snapshot, then the `db_len` parameter gives a coarse check that the snapshot being used is consistent across nodes. This is probably not an essential check since setting up the snapshots would be a pretty visible step in the rollout, so I don't have a problem with removing it anyway, but it sounds like you maybe have a different idea of how the usage might go? It could be worth making sure we're aligned on what the usage environment will look like.

Thanks for the comment Bryan - good to clear this up early on.

This will run on the live code - creating consistent snapshots would be very tedious from an infrastructure perspective. I was under the impression that for delta, we would use the modifications table https://www.notion.so/worldcoin/Sync-Protocol-Requirements-and-Design-1ae8614bdf8c807ab836fa0356e42714?pvs=4#1b38614bdf8c80e8baf3ff32e53ea2da. We would know what was updated based on the versionID and if the sync protocol had the latest.

The nice thing is that currently, re-auth/updates are not live in production and likely will not be for the next month or longer, we may be able to get away with running the code as is for Genesis and assuming only insertions.

Lets discuss this in the sync tommorow.

PR #1395, opened 2025-05-22
aurel/rotation-trait: Types for with/without rotations

naure:
Optional follow-up to #1381.

A type-safe approach to reuse code with and without rotations.


iris-mpc-cpu/src/execution/hawk_main/rot.rs
```diff
@@ -1,39 +1,66 @@
-use std::ops::Deref;
+use std::{marker::PhantomData, ops::Deref};
 
 use iris_mpc_common::ROTATIONS;
 use itertools::Itertools;
 
+pub trait Rotations: Send + Sync + 'static {
+    /// The number of rotations.
+    fn n_rotations() -> usize;
```
dkales, line 8: This could be an associated const as well, right?

PR #1396, opened 2025-05-22
Improve hnsw graph loading

eaypek-tfh:
## Change
- This PR brings down the HNSW loading (iris shares + graph) from 14-15 minutes to 8-9 minutes in stage
- It loads graph and iris dbs in parallel. It also parallelizes left and right graph loading.
- Further improvements will be handled in follow-ups

### DIFF
--- a/deploy/stage/ampc-hnsw-2-stage/values-ampc-hnsw.yaml
+++ b/deploy/stage/ampc-hnsw-2-stage/values-ampc-hnsw.yaml
@@ -113,9 +113,6 @@ env:
         key: KMS_KEYS
         name: application
 
-  - name: SMPC__PROCESSING_TIMEOUT_SECS
-    value: "240"
-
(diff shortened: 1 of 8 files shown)


deploy/stage/ampc-hnsw-2-stage/values-ampc-hnsw.yaml
```diff
@@ -113,9 +113,6 @@ env:
         key: KMS_KEYS
         name: application
 
-  - name: SMPC__PROCESSING_TIMEOUT_SECS
```
carlomazzaferro, line 116: let's keep this and actually increase it to 1200

eaypek-tfh: this was increased to 1240 in one of the latest PRs but the previous value (this one) was not deleted. ArgoCD was complaining with duplicate values, so, I removed it

PR #1398, opened 2025-05-26
[POP-2478] Store rotation indices for partial matches

leonanos8:
### Notes
- Store the rotations, partial matches and counters in cuda kernel during open results
- ~~Remove duplicate work from~~ Keep logic of storing partial matches in merge db results(current logic)
- Return rotation information in ServerJobResult (actor->server)
- Send rotation information to sns results (server->sns)


iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1565,10 +1553,18 @@ impl ServerActor {
             full_face_mirror_match_ids,
             partial_match_ids_left,
             partial_match_ids_right,
+            partial_match_rotation_indices_left,
```
carlomazzaferro, line 1556: We never call `reset_slice` on any of these new buffers. Is this intended?

iris-mpc-gpu/src/dot/kernel.cu
```diff
@@ -160,20 +176,6 @@ extern "C" __global__ void mergeDbResults(unsigned long long *matchResultsLeft,
...
-                unsigned int queryMatchCounter = atomicAdd(&matchCounterLeft[queryIdx], 1);
-                if (queryMatchCounter < MAX_MATCHES_LEN)
-                    partialResultsLeft[MAX_MATCHES_LEN * queryIdx + queryMatchCounter] = dbIdx;
-            }
-            if (matchRight)
-            {
-                unsigned int queryMatchCounter = atomicAdd(&matchCounterRight[queryIdx], 1);
-                if (queryMatchCounter < MAX_MATCHES_LEN)
-                    partialResultsRight[MAX_MATCHES_LEN * queryIdx + queryMatchCounter] = dbIdx;
-            }
```
dkales, line 175: compared to this code there is now a semantic difference in the old impl: Here, we have a counter per query, such that supermatchers get limited per query and we do not lose any valid matches for later querys.

The new implementation only has a single counter, and checks against query*MAX_MATCHES_LEN, which could lead to some supermatches using up the buffer sizes even for later matchers. I think this should be changed to retain the same functionality.

iris-mpc-gpu/src/server/actor.rs
philsippl: also discussed with @leonanos8 offline. I didn't reset the buffers initially, because we only up to the counter anyways. The counter is reset.

iris-mpc-gpu/src/dot/distance_comparator.rs
```diff
@@ -110,24 +109,19 @@ impl DistanceComparator {
...
-            partial_results_left.push(
-                device
-                    .alloc_zeros(ALL_MATCHES_LEN * query_length / ROTATIONS)
-                    .unwrap(),
-            );
-            partial_results_right.push(
-                device
-                    .alloc_zeros(ALL_MATCHES_LEN * query_length / ROTATIONS)
-                    .unwrap(),
-            );
```
philsippl, line 130: let's keep all the old partial match logic to not change the distribution of partial matches: https://github.com/worldcoin/iris-mpc/pull/1398/files#r2112486847

iris-mpc-gpu/src/server/actor.rs
leonanos8: @philsippl I am keeping the implementation as is - resetting both buffers and the counter for consistency

iris-mpc-gpu/src/dot/kernel.cu
```diff
@@ -83,14 +83,22 @@ extern "C" __global__ void openResults(unsigned long long *result1, unsigned lon
...
             }
 
+            unsigned int matchCounter = atomicAdd(&partialResultsCounter[0], 1);
+            if (matchCounter < MAX_MATCHES_LEN * queryLength) 
+            {
+                partialResultsQueryIndices[matchCounter] = queryIdx / ALL_ROTATIONS;
+                partialResultsDbIndices[matchCounter] = dbIdx + offset;
+                partialResultsRotations[matchCounter] = (queryIdx % ALL_ROTATIONS) - ROTATIONS;  // Convert to signed range [-15, 15]
+            }
+
```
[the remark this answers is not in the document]
leonanos8: Good observation @dkales . I can confirm this, I logged the results in runpod and I noticed exactly this. But as you mention we remap them. ✔️

iris-mpc-gpu/src/dot/distance_comparator.rs
leonanos8: @philsippl restored the old logic of storing and fetching partial matches in this commit: https://github.com/worldcoin/iris-mpc/pull/1398/commits/61fbb4744be1cbb1504fc8378a361e4c57fea0df

iris-mpc-gpu/src/dot/kernel.cu
leonanos8: @dkales @philsippl 
In the last [commit](https://github.com/worldcoin/iris-mpc/pull/1398/commits/61fbb4744be1cbb1504fc8378a361e4c57fea0df), I restored the old counter logic and used the `matchCounterLeft` and `matchCounterRight` for the supermatch check.

Now the PR does kind of 'duplicate' work to also gather matches and rotations during the open_results flow and add them as extra information to the ServerJobResult.

philsippl (approved):

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -1616,10 +1656,6 @@ impl ServerActor {
         for dst in [
             &self.distance_comparator.all_matches,
             &self.distance_comparator.match_counters,
-            &self.distance_comparator.match_counters_left,
-            &self.distance_comparator.match_counters_right,
-            &self.distance_comparator.partial_results_left,
-            &self.distance_comparator.partial_results_right,
```
dkales, line 1622: I think this is a leftover change and should also be reverted? IMO these buffers still need to be reset, otherwise the match counters are all wrong and no partial results will be produced?

leonanos8: fixed here: https://github.com/worldcoin/iris-mpc/pull/1398#:~:text=Unverified-,386029f,-leonanos8%20force%2Dpushed

trying to test this in stage before merging

PR #1399, opened 2025-05-26
Aurel/deletions: Implement deletion requests

naure:
- [x] Erase the irises before the search.
- [x] Return the field `deleted_ids`.
- [x] Increment the version ID in memory. The version ID in the SQL database is incremented by Postgres trigger.


naure:
> Looks good! The assumption here is that dummy irises are never requested, right?

Right it should not match with anything.

PR #1402, opened 2025-05-27
Add snapshotting of the database after each genesis run

danielle-tfh:
### Notes
* perform snapshots at the end of each genesis run


carlomazzaferro (approved):
If it works, it works

PR #1405, opened 2025-05-28
Add modifications table usage for the Delta part of Genesis

danielle-tfh:
### Notes
* https://www.notion.so/worldcoin/Sync-Protocol-Requirements-and-Design-1ae8614bdf8c807ab836fa0356e42714?pvs=4#1ae8614bdf8c8099a634e93fb89f57e1 - adds the usage of the modifications table for updates to delta.

### Process
* we only apply modifications that have occurred since the last time Genesis was run and only include modifications that affect what Genesis has already indexed.

### DIFF
--- a/iris-mpc-cpu/src/genesis/state_accessor.rs
+++ b/iris-mpc-cpu/src/genesis/state_accessor.rs
@@ -1,7 +1,7 @@
 use super::utils::{errors::IndexationError, logger};
 use aws_sdk_s3::Client as S3_Client;
 use eyre::Result;
-use iris_mpc_common::{config::Config, IrisSerialId};
+use iris_mpc_common::{config::Config, helpers::sync::Modification, IrisSerialId};
 use iris_mpc_store::{DbStoredIris, Store};
 use serde::{Deserialize, Serialize};
 use sqlx::{Postgres, Transaction};
@@ -10,12 +10,13 @@ use sqlx::{Postgres, Transaction};
 const COMPONENT: &str = "State-Accessor";
 
 /// Domain for persistent state store entry for last indexed id
-const STATE_DOMAIN_LAST_INDEXED: &str = "genesis";
+const STATE_DOMAIN_GENESIS: &str = "genesis";
 
 /// Key for persistent state store entry for last indexed id
 const STATE_KEY_LAST_INDEXED: &str = "id_of_last_indexed";
+
 /// Key for persistent state store entry for last indexed modification id
-const STATE_KEY_MODIFICATION_ID_LAST_INDEXED: &str = "id_of_last_indexed_modification";
+const STATE_KEY_LAST_INDEXED_MODIFICATION_ID: &str = "id_of_last_indexed_modification";
 
 /// Fetch a batch of iris data for indexation.
 ///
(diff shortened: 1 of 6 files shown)


iris-mpc-store/src/lib.rs
siajasl, line 1: What about non-persisted modifications ?

iris-mpc-upgrade-hawk/src/genesis/mod.rs
siajasl, line 1: My current understanding is that modifications are to be applied after indexation, I.E. indexation is now a 2 phase process.  Therefore I presume that fetching of modifications can wait until after the Iris batches have been processed.  Furthermore we may wish to process modifications in batches rather than pulling potentially a large set into memory ... however ATM I do not fully understand operational strategy and thus this comment maybe somewhat redundant.

iris-mpc-cpu/src/genesis/state_accessor.rs
```diff
@@ -224,12 +250,33 @@ pub async fn set_last_indexed_modification_id(
...
+        tx,
+        STATE_DOMAIN_GENESIS,
+        STATE_KEY_LAST_INDEXED_MODIFICATION_ID,
+    )
+    .await
 }
 
 #[cfg(test)]
-#[cfg(feature = "db_dependent")]
+// #[cfg(feature = "db_dependent")]
```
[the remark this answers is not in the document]
bgillesp: My approach has been to use the `db_dependent` feature flag on unit tests that require the localstack Postgres client to be running, so that in the absence of this setup, some significant portion of the test suite can still run without expected failures.  In theory we could just remove the flag and expect unit tests to be run with the dev localstack present, or alternatively we could replace with a more descriptive flag name like `localstack_tests` which could be enabled by default.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
bgillesp: Danielle's approach of reconciling modifications first was also my understanding of the planned design from Tuesday.  The idea here is that since we're reading the iris shares database into memory from prod for every run of genesis, some underlying iris data may have changed since the last read.  Which iris data this is can be identified by looking at associated modifications table entries, and inserting as new the modified iris codes before continuing the indexing process.

iris-mpc-store/src/lib.rs
```diff
@@ -1364,6 +1420,168 @@ pub mod tests {
...
+        // ID 4: reset_update (should be included when persisted = true)
+        let mod4 = store
+            .insert_modification(
+                103,
+                RESET_UPDATE_MESSAGE_TYPE,
+                Some("http://example.com/reset"),
+            )
+            .await?;
+
+        // ID 4: uniqueness (should NOT be included due to request type)
```
bgillesp, line 1461: Comment probably should be `ID 5`, and later references to IDs also incremented.

iris-mpc-store/src/lib.rs
```diff
@@ -1364,6 +1420,168 @@ pub mod tests {
...
+            request_types[0] == IDENTITY_DELETION_MESSAGE_TYPE,
+            "Should include persisted identity_deletion requests"
+        );
+        assert!(
+            request_types[1] == REAUTH_MESSAGE_TYPE,
+            "Should include persisted reauth requests"
+        );
+        assert!(
+            request_types[2] == RESET_UPDATE_MESSAGE_TYPE,
+            "Should NOT include persisted reset_update requests"
```
bgillesp, line 1521: Should include, rather than should not?

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -221,11 +244,37 @@ async fn exec_main_loop(
...
+                let msg = log_error(format!(
+                    "HawkActor does not support deletion of identities: modification: {:?}",
+                    modification
+                ));
+                bail!(msg);
+            }
+            // TODO: apply modification to the graph
+        }
+        // TODO: set last indexed modification id
+        // set_last_indexed_modification_id(&mut db_tx, _max_modification_id).await?;
```
bgillesp, line 275: This update of the stored "last indexed modification" should be done atomically per modification applied so that the db is in a consistent state after each processed modification entry.

iris-mpc-cpu/src/genesis/state_accessor.rs
```diff
@@ -112,6 +115,40 @@ pub async fn fetch_iris_deletions(
     Ok(s3_object.deleted_serial_ids)
 }
 
+/// Fetch Iris modifications that need to be applied post indexation phase one.
```
bgillesp, line 118: Can we extend the description to specify that the modifications are returned in increasing order of ids?

iris-mpc-store/src/lib.rs
```diff
@@ -465,6 +470,57 @@ WHERE id = $1;
...
+    /// This is for the genesis protocol to fetch modifications that need to be indexed.
+    ///
+    /// # Arguments
+    ///
+    /// * `after_modification_id` - Modification identifier from which to filter result.
+    /// * `serial_id_less_than` - Iris serial identifier to which to filter result.
+    ///
+    /// # Returns
+    ///
+    /// An ordered vector of `Modification` instances.
```
bgillesp, line 483: Also we can specify here that the returned modifications are ordered by increasing modification id.

bgillesp (approved):
Looks great, thanks @danielle-tfh!  Just a few small nits from me -- otherwise we should be good to pick up the graph modifications implementation from here!

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -82,15 +86,31 @@ pub async fn exec_main(
     ));
 
     // Process: set Iris serial identifiers marked for deletion and thus excluded from indexation.
-    let exclusions_all = fetch_iris_deletions(&config, &aws_s3_client).await?;
-    let exclusions = exclusions_all
+    let excluded_serial_ids = fetch_iris_deletions(&config, &aws_s3_client)
+        .await
+        .unwrap()
```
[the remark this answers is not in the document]
danielle-tfh: I will leave this to @siajasl to pick up.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
siajasl: OK misunderstanding upon my part ... makes sense.

PR #1410, opened 2025-05-30
enable LUC in hnsw stage

carlomazzaferro:
1. Enable LUC
2. We actually do care about full face mirror attacks, because they have been implemented. This PR enables sending them to the signup service.

### DIFF
--- a/iris-mpc/src/services/processors/job.rs
+++ b/iris-mpc/src/services/processors/job.rs
@@ -61,7 +61,6 @@ pub async fn process_job_result(
     let now = Instant::now();
 
     let _modifications = modifications;
-    let _full_face_mirror_attack_detected = full_face_mirror_attack_detected;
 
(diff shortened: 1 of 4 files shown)


iris-mpc/src/services/processors/job.rs
```diff
@@ -108,12 +107,15 @@ pub async fn process_job_result(
...
-                None,  // not applicable for hnsw
-                false, // not applicable for hnsw
+                None, // not applicable for hnsw
+                None, // not applicable for hnsw
+                None, // not applicable for hnsw
+                None, // not applicable for hnsw
+                None, // not applicable for hnsw
+                match full_face_mirror_attack_detected.is_empty() {
+                    false => full_face_mirror_attack_detected[i],
+                    true => false,
```
leonanos8, line 117: good catch @carlomazzaferro , I did this intentionally back then because the mirror check was not implemented in HNSW

PR #1411, opened 2025-05-30
Migrate persisted state to the graph store

danielle-tfh:
### Notes
* migrate the persisted state to the graph store as its mainly used for Genesis and graph updates.
* currently, the state is read from the iris pg store but written to the graph store (https://github.com/worldcoin/iris-mpc/blob/5450212b00cb8cdba2fbf4b1eb0061f9ae97dd39/iris-mpc-upgrade-hawk/src/genesis/mod.rs#L645-L647)
* therefore the state has not been kept

This PR migrates all to use the graph store - this also makes sense since its only being used by HNSW related mechanisms


danielle-tfh:
https://linear.app/worldcoin/issue/POP-2542/fix-persistent-store-for-genesis

PR #1413, opened 2025-05-30
HNSW-Genesis: Standardise error propogation

siajasl:
Minor PR in response to feedback regarding standardized error propagation in genesis main.  This resolves the anomoly and pushes the filtering of exclusions into the state-accessor component.

### DIFF
--- a/iris-mpc-cpu/src/genesis/state_accessor.rs
+++ b/iris-mpc-cpu/src/genesis/state_accessor.rs
@@ -124,6 +124,7 @@ pub async fn fetch_iris_deletions(
 }
 
 /// Fetch Iris modifications that need to be applied post indexation phase one.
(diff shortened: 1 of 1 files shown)


iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -291,9 +286,9 @@ async fn exec_main_loop(
                 bail!(msg);
             }
             // TODO: apply modification to the graph
-            // TODO: set last indexed modification id
-            // set_last_indexed_modification_id(&mut db_tx, _max_modification_id).await?;
         }
+        // TODO: set last indexed modification id
```
danielle-tfh, line 290: This was changed based on the comment from Bryan to set after each modification change

danielle-tfh (approved):

PR #1414, opened 2025-05-30
Add dynamic calculation of batch size

danielle-tfh:
### Notes
* process batch rate based on https://www.notion.so/inversed/Parallel-HNSW-Insertions-Analysis-201a3b540bfe808e9803d1bd09bdfa31?pvs=4
* for now use a larger r as default - based on performance this can be lowered


siajasl (approved):
LTGM

PR #1415, opened 2025-05-30
Add static serial ids for local stack

danielle-tfh:
cc @siajasl - make deleted serial ids static


scripts/tools/init-localstack.sh
siajasl, line 1: - What is the % of deletions in PROD ?  We should reflect that I suppose.
- I had generated [a file](https://github.com/siajasl/mpctl/blob/main/resources/misc/deleted_serial_ids.json) with a bunch of deletions at a 1% rate - I dumped it in MPCTL just in case.

siajasl (approved):

scripts/tools/init-localstack.sh
danielle-tfh: I think its ok to not reflect this - this is just for consisted testing

PR #1418, opened 2025-06-01
Siajasl/hawk genesis 2

siajasl:
- Refactoring of main inner loop into two steps:
  - Step 1: Applying modifications
  - Step 2: Indexation
- Refactoring of main function ordering
- Refactoring of snapshotting

### DIFF
--- a/iris-mpc-cpu/src/genesis/state_accessor.rs
+++ b/iris-mpc-cpu/src/genesis/state_accessor.rs
@@ -55,6 +55,7 @@ pub async fn fetch_iris_batch(
 ///
 /// * `config` - Application configuration instance.
 /// * `s3_client` - A configured AWS S3 client instance.
+/// * `max_indexation_id` - Maximum Iris serial identifier to be indexed.
 ///
 /// # Returns
 ///
@@ -63,6 +64,7 @@ pub async fn fetch_iris_batch(
 pub async fn fetch_iris_deletions(
     config: &Config,
     s3_client: &S3_Client,
+    max_indexation_id: IrisSerialId,
 ) -> Result<Vec<IrisSerialId>, IndexationError> {
     // Struct for deserialization.
     #[derive(Serialize, Deserialize, Debug, Clone)]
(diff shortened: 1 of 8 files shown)


iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -240,96 +316,162 @@ pub async fn exec_main(
...
-/// * `modifications` - List of new modifications entries to be processed.
+/// * `ctx` - Execution context information.
 /// * `task_monitor` - Tokio task monitor to coordinate with other threads.
 /// * `shutdown_handler` - Handler coordinating process shutdown.
 /// * `hawk_actor` - Hawk actor managing indexation & search over an HNSW graph.
 /// * `tx_results` - Channel to send job results to DB persistence thread.
 ///
-#[allow(clippy::too_many_arguments)]
-async fn exec_main_loop(
+async fn exec_main_inner(
```
danielle-tfh, line 325: inner feels too broad. Should we keep it as loop? A better idea would be to make one exec_main and have some form of setup function thats called inside.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -773,6 +877,40 @@ fn log_warn(msg: String) -> String {
...
+///
+/// # Arguments
+///
+/// * `config` - Application configuration instance.
+/// * `rds_client` - AWS RDS SDK client.
+/// * `batch_size` - Size of indexation batches.
+/// * `last_indexed_id` - Last Iris serial id to have been indexed.
+/// * `max_indexation_id` - Maximum Iris serial id to which to index.
+///
+async fn set_db_snapshot(
```
danielle-tfh, line 890: These also seem to just be duplicate methods that can rather call the main one directly,

iris-mpc-upgrade-hawk/src/genesis/mod.rs
siajasl: Yeah I will abstract all the setup and then there will be a single exec function with exec_setup, exec_delta, exec_indexation, exec_snapshot sub-functions.

iris-mpc-cpu/src/genesis/state_accessor.rs
```diff
@@ -248,6 +249,70 @@ pub async fn set_last_indexed_modification_id(
...
 
+/// Performs an RDS cluster snapshot.
+///
+/// # Arguments
+///
+/// * `client` - AWS RDS SDK client.
+/// * `db_cluster_endpoint` - RDS cluster endpoint.
+/// * `snapshot_id` - Snapshot identifier.
+///
+pub async fn set_rds_snapshot(
```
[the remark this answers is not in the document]
danielle-tfh: I see but this is `state_accessor`. There's a clear distinction between:

* "What genesis state are we currently tracking?" (this file's job) `state_accessor`
* "Go create of infrastructure related snapshots"

I think it is more than fine for this to stay in the main file - it is not too many lines that warrant a new file/method. It also keeps this file with a clear objective being: "save and get the latest state to know where to run Genesis from" (No kitchen sink file)

I am happy to disagree and commit though. PR approved

danielle-tfh (approved):

PR #1425, opened 2025-06-03
test runner issue with docker setup

orlra:
just a test to see what part of runner is a problem


orlra:
not needed anymore, was just a debug run.

PR #1432, opened 2025-06-04
[POP-2533] Implement processing of modifications in delta

bgillesp:
PR adds support for processing modifications table entries persisted by the production nodes since the last run of the genesis indexer.  The approach is to turn the genesis `JobRequest` and `JobResult` structs into enums with separate variants for indexing of batches and for processing modification table entries.

The current version emulates the main server behavior for reath and reset type transactions, which is to simply re-insert the current newest entry into the graph, without doing any pruning of old versions still existing in the graph.  We will need to modify this behavior later to do deletion of the existing graph nodes and, very rarely, reassignment of the graph entry point.


iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -750,37 +822,64 @@ async fn get_results_thread(
...
+                    ));
+
+                    log_info(format!(
+                        "Job Results :: Persisted to dB: batch-id={batch_id}"
+                    ));
+
+                    // Notify background task responsible for tracking pending batches.
+                    shutdown_handler_bg.decrement_batches_pending_completion();
+                }
+                JobResult::Modification {
```
danielle-tfh, line 855: So it will be changed in memmory but only persisted to the database at the end? That is really nice for not having any graphs in between phases. I wonder if this means we should only change the last persisted index after the graph is persisted in this case.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -344,38 +376,89 @@ async fn exec_setup(
 /// * `ctx` - Execution context information.
 /// * `modifications` - Set of indexation modifications to apply.
 ///
-async fn exec_delta(ctx: &ExecutionContextInfo) -> Result<()> {
+async fn exec_delta(
+    config: &Config,
+    ctx: &ExecutionContextInfo,
+    _imem_iris_stores: &BothEyes<SharedIrisesRef>,
```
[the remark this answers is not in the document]
bgillesp: Not necessary now -- removed.  Was a hold-over from the format of `exec_indexation` it looks like, but isn't needed due to the way we retrieve the modification queries.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
bgillesp: I believe the shutdown handler just maintains an atomic usize counter representing the number of database operations that have been started but not yet completed -- so DB updates are handled atomically on a per-modification or per-indexation-batch basis, but the server waits sometimes for DB updates to finish:
- Before moving from delta protocol to indexing
- Before moving from indexing to running the DB snapshot
- When the binary receives an interrupt signal

The goal is for each atomic transaction to leave the DB in a consistent state, so the "last persisted index" and "last indexed modification id" fields are updated as needed once per committed DB transaction.

PR #1438, opened 2025-06-05
Fix skip persistence on batch

danielle-tfh:
### Notes
* add correct value for skip persistence

### DIFF
--- a/iris-mpc/src/services/processors/batch.rs
+++ b/iris-mpc/src/services/processors/batch.rs
@@ -381,6 +381,8 @@ impl<'a> BatchProcessor<'a> {
         }
 
         self.delete_message(sqs_message).await?;
+        // skip_persistence is only used for uniqueness requests
+        self.batch_query.skip_persistence.push(false);
         self.msg_counter += 1;
         Ok(())
     }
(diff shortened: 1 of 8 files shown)


carlomazzaferro (approved):

iris-mpc/src/services/processors/batch.rs
```diff
@@ -381,6 +381,8 @@ impl<'a> BatchProcessor<'a> {
         }
 
         self.delete_message(sqs_message).await?;
+        // skip_persistence is only used for uniqueness requests
+        self.batch_query.skip_persistence.push(false);
```
naure, line 385: Deletions are not related to this field `skip_persistence` and it should not push a value into it.

The structure `BatchQuery` has multiple independent "tables" inside:
- Uniqueness, reauth, and reset check, with columns: `request_ids`, `request_types`, `metadata`, `or_rule_indices`, and `skip_persistence`.
- Deletions with columns: `deletion_requests_*`.
- Reset updates with columns: `reset_update_*`.

iris-mpc/src/services/processors/batch.rs
```diff
@@ -582,6 +597,8 @@ impl<'a> BatchProcessor<'a> {
                 code_right: right_shares.code,
                 mask_right: right_shares.mask,
             });
+        // skip_persistence is only used for uniqueness requests
+        self.batch_query.skip_persistence.push(false);
```
naure, line 601: Same as above, should not use the field `skip_persistence` in reset update requests.

Note that this function `process_reset_update_request` may have a mistake in it already: it should not do `request_types.push(RESET_UPDATE_MESSAGE_TYPE)`.

naure (dismissed):

iris-mpc/src/services/processors/batch.rs
naure: @carlomazzaferro Can you double-check the reasoning here?

danielle-tfh: @naure you are correct! Thank you for noticing this

PR #1441, opened 2025-06-06
[POP-2563] Genesis: Batch generator unit test hardening

siajasl:
1. Hardens unit tests around batch generation.
2. Introduces a new enum, BatchSize, to correctly represent Static | Dynamic batch sizing.
3. Improves binary logic.


iris-mpc-cpu/src/genesis/batch_generator.rs
```diff
@@ -194,109 +263,335 @@ impl BatchIterator for BatchGenerator {
...
+            _ => match self {
+                BatchSize::Dynamic(r, M) => {
+                    // r: configurable parameter for error rate.
+                    // M: HNSW parameter for nearest neighbors.
+                    // n: current graph size (last_indexed_id).
+                    let N = last_indexed_id;
+
+                    // batch_size: floor(N/(Mr - 1) + 1)
+                    let batch_size =
+                        (N as f64 / (*M as f64 * *r as f64 - 1.0) + 1.0).floor() as usize;
```
bgillesp, line 308: This formula should be fine for the `N=0` case as well if we use integer arithmetic instead of floating point...  Can we instead use something like `(N as usize).div_euclid(*M * *r - 1) + 1` and eliminate the match statement on `last_indexed_id`?

iris-mpc-cpu/src/genesis/batch_generator.rs
```diff
@@ -1,29 +1,56 @@
...
+    ops::RangeInclusive,
+    sync::{Once, OnceLock},
+};
 
 /// Component name for logging purposes.
 const COMPONENT: &str = "Batch-Generator";
 
+/// Synchronization mechanism for batch size initialization.
+static BATCH_SIZE_INIT: Once = Once::new();
+static BATCH_SIZE_VALUE: OnceLock<usize> = OnceLock::new();
```
bgillesp, line 21: Do we need to keep the batch size in a static context?  I view this `BatchGenerator` type as a lightweight component that is utilized in a single-threaded coordinator environment, so probably doesn't need this heavy of machinery.  (Seems unlikely that we would need multithreaded batch origination for the genesis process.)

iris-mpc-cpu/src/genesis/batch_generator.rs
```diff
@@ -117,31 +117,116 @@ impl fmt::Display for BatchGenerator {
...
+    ) -> impl Future<Output = Result<Option<Batch>, IndexationError>> + Send;
+}
+
+/// Policy over batch size calculations.
+#[derive(Debug, Eq, PartialEq)]
+pub enum BatchSize {
+    /// Static batch size.
+    Static(usize),
+    /// Dynamic batch size with size error coefficient & hnsw-m param.
+    Dynamic(usize, usize),
```
bgillesp, line 158: Can we make these struct type rather than tuple type?  I think especially for the `Dynamic` variant it would be helpful to have descriptive field names.

bgillesp (changes requested):
I'm getting a few unit tests failing surrounding the new batch generator changes, so should get those sorted out.  On the general design, my feeling is that the static variable for the batch size may be heavier than we need, since batches for the genesis indexer will only ever be generated by a single top-level coordinator thread.  My preference would be to have `Static` and `Dynamic` batch sizes that produce truly static or dynamic values (i.e. the `Dynamic` variant actually updates the size when producing each new batch based on the current indexation height), and to represent the current binary usage of the `0` batch size value as a `Static` batch size computed using the dynamic formula.  (Perhaps add `new_static`, `new_dynamic`, and `new_static_from_dynamic_formula` constructors for the `BatchSize` type.)  Let me know what you think.

iris-mpc-cpu/src/genesis/batch_generator.rs
siajasl: Done.

PR #1445, opened 2025-06-10
Pop 2530/implement anon stat change in hnsw that do not count all rotations

leonanos8:


iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -538,30 +541,41 @@ impl HawkActor {
...
+                        .push(distance_share);
+                }
+            }
+        }
+
         tracing::info!(
-            "Keeping {} distances for eye {side} out of {} search results. Cache size: {}/{}",
-            distances.clone().count(),
+            "Keeping distances for eye {side} out of {} search results. Cache size: {}/{}",
+            // distances.clone().count(),
```
leonanos8, line 568: lets remove this if not planning to log the amount of distances that we are storing in the cache_distances buffer?

dkales (approved):

PR #1447, opened 2025-06-10
remove properties from batch

carlomazzaferro:
Addressing the comments from a previously merged PR: https://github.com/worldcoin/iris-mpc/pull/1438

The structure BatchQuery has multiple independent "tables" inside:

Uniqueness, reauth, and reset check, with columns: 
 - request_ids
 - request_types 
 - metadata
 - or_rule_indices 
 - skip_persistence
 
Deletions with columns: deletion_requests_*.

Reset updates with columns: reset_update_*.

### DIFF
--- a/iris-mpc/src/services/processors/batch.rs
+++ b/iris-mpc/src/services/processors/batch.rs
@@ -381,8 +381,6 @@ impl<'a> BatchProcessor<'a> {
         }
 
         self.delete_message(sqs_message).await?;
-        // skip_persistence is only used for uniqueness requests
-        self.batch_query.skip_persistence.push(false);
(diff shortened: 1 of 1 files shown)


iris-mpc/src/services/processors/batch.rs
```diff
@@ -381,8 +381,6 @@ impl<'a> BatchProcessor<'a> {
         }
 
         self.delete_message(sqs_message).await?;
-        // skip_persistence is only used for uniqueness requests
-        self.batch_query.skip_persistence.push(false);
         self.msg_counter += 1;
```
danielle-tfh: I think we can also remove the msg_counter

PR #1449, opened 2025-06-12
Store graph modifications

eaypek-tfh:
## Change
- Store modifications upon each request with IN_PROGRESS status
- Correlate modifications to corresponding graph mutations. Introduce `SingleHawkMutation` which contains a `ModificationKey` field and update `HawkMutation` to be a vector of `SingleHawkMutation`.
  - Please note that modifications are ordered and stored by the request order. Each new request gets the next modification id. However, HNSW currently applies graph mutations in a different order where reset update graph mutations are applied after all the other mutations. There is a task to fix that [here](https://linear.app/worldcoin/issue/POP-2588/process-hnsw-reset-update-messages-in-exact-request-order)
- When marking a modification as completed, fetch the corresponding graph mutation by `ModificationKey` and persist it into the `modifications` table

## Next steps:
- Sync and apply modifications in HNSW including graph mutations.


iris-mpc-cpu/src/genesis/hawk_handle.rs
```diff
@@ -167,7 +167,24 @@ impl Handle {
...
         let results: [_; 2] = results_.try_into().unwrap();
 
-        Ok(JobResult::new(request, HawkMutation(results)))
+        // Convert the results into SingleHawkMutation format
+        let [left_plans, right_plans] = results;
+        assert_eq!(left_plans.len(), right_plans.len());
+        let mut mutations = Vec::new();
+
+        for i in 0..left_plans.len() {
+            let left_plan = left_plans.get(i).cloned().flatten();
```
naure, line 176: To clarify, it never makes sense to have only one side. It would be clearer to remove the `Option` as soon as possible rather than propagating them into `SingleHawkMutation`.

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -1274,14 +1316,15 @@ impl HawkHandle {
...
 
-        let mut connect_plans = HawkMutation([vec![], vec![]]);
+        // Store plans for both sides using BothEyes structure
+        let mut plans_both_sides: BothEyes<Vec<Option<ConnectPlan>>> = [Vec::new(), Vec::new()];
 
         // For both eyes.
-        for (side, sessions, search_results, reset_results) in
-            izip!(&STORE_IDS, sessions, search_results, resets.search_results)
+        for (side_idx, (side, sessions, search_results, reset_results)) in
+            izip!(&STORE_IDS, sessions, search_results, resets.search_results).enumerate()
```
naure, line 1327: `side_idx` is the same as `side as usize`, not necessary to add it.

PR #1453, opened 2025-06-12
Add fix for reading graphs with deletions that are not present

danielle-tfh:
### Notes
* With deletions we cannot use the serial ids to load the database since some will be missing
* In this PR we use offset instead


bgillesp (approved):
Yep, that needed to be fixed -- nice find!

PR #1455, opened 2025-06-13
[POP-2586] Select HNSW layers based on identifiers

bgillesp:
This PR revises the method used by the main server executables for choosing random layers for insertion of new HNSW graph entries.  Instead of using an ongoing, stateful PRNG, the new approach uses a globally keyed PRF which takes as input a vector identifier -- either an enrollment request UUID + iris side for new insertions, or a vector id + iris side for insertions of irises with existing vector ids.  This makes the graph construction process less path-dependent, as the insertion layer for nodes no longer depends on ordering of PRNG calls and persistence of PRNG state across execution environments.  The functionality for selecting layers via a PRNG is still available, and is used in some tests, but the main server functionality now uses the PRF-based approach.

### DIFF
--- a/iris-mpc-cpu/src/execution/hawk_main.rs
+++ b/iris-mpc-cpu/src/execution/hawk_main.rs
@@ -2,6 +2,7 @@ use super::player::Identity;
 pub use crate::hawkers::aby3::aby3_store::VectorId;
 use crate::{
     execution::{
+        hawk_main::search::SearchIds,
         local::generate_local_identities,
         player::{Role, RoleAssignment},
         session::{NetworkSession, Session, SessionId},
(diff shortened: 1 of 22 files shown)


iris-mpc-cpu/src/hnsw/searcher.rs
```diff
@@ -232,13 +237,35 @@ impl HnswSearcher {
...
+    ///
+    /// Generates the layer value based on the evaluation of a keyed PRF on a
+    /// hashable input identifier `value`, for instance a vector id or a request id.
+    pub fn select_layer_prf<H: Hash>(&self, prf_key: &[u8; 16], value: &H) -> Result<usize> {
+        // produce `value_hash` from `value` using `SipHasher13`, keyed with `prf_key`
+        let mut hasher = SipHasher13::new_with_key(prf_key);
+        value.hash(&mut hasher);
+        let value_hash: u64 = hasher.finish();
+
+        // initialize `AesRng` with seed `prf_key ^ (value_hash || value_hash)`
```
naure, line 258: Any particular reason behind this algorithm?

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -132,6 +133,7 @@ pub struct HawkActor {
 
     // ---- Shared setup ----
     searcher: Arc<HnswSearcher>,
+    prf_key: Option<Arc<[u8; 16]>>,
```
naure, line 136: There was no need for `OnceLock`, I made it boring with `Option`.

naure (approved):

PR #1459, opened 2025-06-18
Add defining serial id when inserting into plaintext HNSW database

danielle-tfh:
### Notes
* In order to compare graphs created by Genesis (where deletions are skipped), we need a way to skip serial ids for creating benchmark data ie:

```
cargo run --release --bin init-test-dbs -- --source iris-mpc-cpu/data/store-doubled.ndjson --db-url-party1 "postgres://postgres:postgres@localhost:5432/SMPC_dev_0" --db-url-party2 "postgres://postgres:postgres@localhost:5432/SMPC_dev_1" --db-url-party3 "postgres://postgres:postgres@localhost:5432/SMPC_dev_2" --db-schema-party1 "SMPC_dev_0" --db-schema-party2 "SMPC_dev_1" --db-schema-party3 "SMPC_dev_2" -s 1000 --skip-insert-serial-ids 355 356 357 358 359 360 361 362 363 364 365 366 367 368 369 370 371 372 373 374 375 540 541 542 543 544 545 546 547 548 549 550 551 552 553 554 555 556 557 558 559 560 561 562 563 564 565 566 567 568 569 570 571 572 573 574 575
```

In order for that we need to alter the way to the plaintext database uses points ie to not expect that the serial id is the same as the index of the point in the array

### DIFF
--- a/iris-mpc-cpu/src/hawkers/plaintext_store.rs
+++ b/iris-mpc-cpu/src/hawkers/plaintext_store.rs
@@ -53,7 +53,7 @@ impl PlaintextStore {
             .into_iter()
             .enumerate()
             .map(|(idx, iris)| (idx as u32 + 1, iris))
-            .collect::<HashMap<u32, IrisCode>>();
+            .collect::<HashMap<SerialId, IrisCode>>();
         let next_id = points.len() as u32 + 1;
         Self { points, next_id }
     }
@@ -74,13 +74,13 @@ impl PlaintextStore {
         }
 
         // sort in order to ensure deterministic behavior
-        let mut keys: Vec<_> = self.points.keys().cloned().collect();
-        keys.sort();
-        keys.truncate(graph_size);
+        let mut serial_ids: Vec<_> = self.points.keys().cloned().collect();
+        serial_ids.sort();
+        serial_ids.truncate(graph_size);
 
-        for key in keys {
-            let query = self.points[&key].clone();
-            let query_id = VectorId::from_serial_id(key);
+        for serial_id in serial_ids {
+            let query = self.points[&serial_id].clone();
+            let query_id = VectorId::from_serial_id(serial_id);
             let insertion_layer = searcher.select_layer_rng(&mut rng)?;
             let (neighbors, set_ep) = searcher
                 .search_to_insert(self, &graph, &query, insertion_layer)
@@ -91,10 +91,6 @@ impl PlaintextStore {
         }
 
         Ok(graph)
-    }
-
-    pub fn set_next_id(&mut self, next_id: u32) {
-        self.next_id = next_id;
     }
 
     pub fn insert_with_id(&mut self, serial_id: SerialId, query: &IrisCode) -> VectorId {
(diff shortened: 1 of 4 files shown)


iris-mpc-cpu/src/hawkers/plaintext_store.rs
```diff
@@ -17,13 +17,20 @@ use tracing::debug;
...
+    pub serial_id: u32,
+}
 
 /// Vector store which works over plaintext iris codes and distance computations.
 ///
 /// This variant is only suitable for single-threaded operation.
 #[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct PlaintextStore {
-    pub points: Vec<IrisCode>,
+    pub points: HashMap<u32, IrisCode>,
```
bgillesp, line 33: Definitely swapping in a hashmap is the right plan here.  Should we use `iris_mpc_common::vector_id::SerialId` instead of `u32` for the keys?

iris-mpc-cpu/src/hawkers/plaintext_store.rs
```diff
@@ -81,14 +102,17 @@ fn fraction_less_than(dist_1: &(u16, u16), dist_2: &(u16, u16)) -> bool {
 }
 
 impl VectorStore for PlaintextStore {
-    type QueryRef = IrisCode;
+    type QueryRef = IrisCodeWithSerialId;
```
bgillesp, line 105: This isn't my favorite change since the serial id isn't needed for the core operations supported by a `VectorStore` -- i.e. we can do a full HNSW search for a query `IrisCode` without specifying the serial id that it will eventually use (or even without eventually inserting it if we're only interested in the nearest neighbor search results).  Could we consider using an approach similar to what the `Aby3Store` struct uses?

- Maintain a `next_id` variable which tracks the next id available for insertion, always set to the maximum currently stored id, plus one
- For standard insertion (`append`), use this id and increment `next_id`
- Also provide a separate `insert` function which has the serial id as a parameter, and updates `next_id` to be one larger than this if the value is larger than the current value of `next_id`

Then when we need to insert a new node into an HNSW graph with a special serial id, we can do something like what's done in `init_test_dbs.rs` lines 370-383, but using the general `insert` function with a chosen serial id.

bgillesp (commented):
Thanks for working on this!  Only had time for a partial look at the PR so far -- let me know what you think about the alternate suggestion for handling insertion into `PlaintextStore` with a specific serial id.

iris-mpc-cpu/src/hawkers/plaintext_store.rs
bgillesp: Yes, I totally get where you're coming from.  I don't see anything that breaks immediately with this approach, but I will say that my gut feeling about the direction is something like "\*loud sirens + flashing lights\* this cuts against the grain of the underlying abstraction, and is likely to have unintended side-effects down the line for a core component".  Trying to unpack this a bit more, I think some of the intuition is coming from previous versions of `PlaintextStore` which mixed the `QueryRef` abstraction and the inserted `VectorRef` abstraction -- e.g. at the time by preprocessing a query to "pre-insert" it in the store before you could do comparisons.  This arrangement made it much harder to reason about intermediate states during insertion, and in this case made it difficult to "search but not insert".

In the current setting, decoupling the query type from details of how it is eventually to be inserted means that queries can be passed around flexibly and don't need to know anything about the implementation of `PlaintextStore` or synchronization with other queries, pending or happening in parallel.  I think in particular because this is a component that is used widely to represent the "core plaintext functionality" for the HNSW search, I'm more concerned than I usually would be about prioritizing decoupling/flexibility as much as possible.

Not fully ruling out the "id-bundled query" approach, but it does make me nervous.

iris-mpc-cpu/src/py_bindings/plaintext_store.rs
```diff
@@ -41,9 +41,9 @@ pub fn from_ndjson_file(filename: &str, len: Option<usize>) -> io::Result<Plaint
 
     // Iterate over each deserialized object
     let mut vector = PlaintextStore::new();
-    for json_pt in stream {
+    for (idx, json_pt) in stream.into_iter().enumerate() {
         let json_pt = json_pt?;
-        vector.points.push((&json_pt).into());
+        vector.points.insert(idx as u32 + 1, (&json_pt).into());
     }
```
bgillesp, line 48: Set `vector.next_id` to `vector.points.len() + 1` here?

bgillesp (approved):
This is looking great -- thanks so much for the rework!  A couple of small suggestions (including one quick fix for the failing unit tests), but otherwise this seems well structured to me.

iris-mpc-py/src/py_hnsw/pyclasses/plaintext_store.rs
```diff
@@ -14,13 +14,14 @@ impl PyPlaintextStore {
     }
 
     pub fn get(&self, id: u32) -> PyIrisCode {
-        self.0.points[id as usize].clone().into()
+        self.0.points[&id].clone().into()
```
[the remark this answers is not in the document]
danielle-tfh: Nice! Thank you!

PR #1462, opened 2025-06-18
Add iris copying from iris store to HSNW store

danielle-tfh:
### Notes
* draft PR - design still to be confirmed. It makes more sense to copy once then copy after each batch. We can use the highest serial id to ensure we do not copy the same iris twice.
* note: we need the version id to be copied to be inline with the original iris store

### DIFF
--- a/iris-mpc-store/src/lib.rs
+++ b/iris-mpc-store/src/lib.rs
@@ -98,6 +98,16 @@ pub struct StoredIrisRef<'a> {
     pub right_mask: &'a [u16],
 }
 
+#[derive(Clone)]
+pub struct StoredIrisVectorRef<'a> {
+    pub id: i64,
+    pub version_id: i16,
+    pub left_code: &'a [u16],
+    pub left_mask: &'a [u16],
+    pub right_code: &'a [u16],
+    pub right_mask: &'a [u16],
+}
+
 // Convertor: DbStoredIris -> IrisIdentifiers.
 impl From<&DbStoredIris> for VectorId {
     fn from(value: &DbStoredIris) -> Self {
@@ -246,6 +256,39 @@ impl Store {
         );
         query.push_values(codes_and_masks, |mut query, iris| {
             query.push_bind(iris.id);
+            query.push_bind(cast_slice::<u16, u8>(iris.left_code));
+            query.push_bind(cast_slice::<u16, u8>(iris.left_mask));
+            query.push_bind(cast_slice::<u16, u8>(iris.right_code));
+            query.push_bind(cast_slice::<u16, u8>(iris.right_mask));
+        });
+
+        query.push(" RETURNING id");
+
+        let ids = query
+            .build()
+            .fetch_all(tx.deref_mut())
+            .await?
+            .iter()
+            .map(|row| row.get::<i64, _>("id"))
+            .collect::<Vec<_>>();
+
+        Ok(ids)
+    }
+
+    pub async fn insert_copy_iris(
+        &self,
+        tx: &mut Transaction<'_, Postgres>,
+        codes_and_masks: &[StoredIrisVectorRef<'_>],
+    ) -> Result<Vec<i64>> {
+        if codes_and_masks.is_empty() {
+            return Ok(vec![]);
+        }
+        let mut query = sqlx::QueryBuilder::new(
+            "INSERT INTO irises (id, version_id, left_code, left_mask, right_code, right_mask)",
+        );
+        query.push_values(codes_and_masks, |mut query, iris| {
+            query.push_bind(iris.id);
+            query.push_bind(iris.version_id);
             query.push_bind(cast_slice::<u16, u8>(iris.left_code));
             query.push_bind(cast_slice::<u16, u8>(iris.left_mask));
             query.push_bind(cast_slice::<u16, u8>(iris.right_code));
--- a/iris-mpc-upgrade-hawk/src/genesis/mod.rs
+++ b/iris-mpc-upgrade-hawk/src/genesis/mod.rs
@@ -29,10 +29,12 @@ use iris_mpc_cpu::{
         utils, BatchGenerator, BatchIterator, BatchSize, Handle as GenesisHawkHandle,
         IndexationError, JobResult,
     },
-    hawkers::aby3::aby3_store::Aby3Store,
-    hnsw::graph::graph_store::GraphPg,
+    hawkers::aby3::aby3_store::{Aby3Store, SharedIrisesRef},
+    hnsw::{self, graph::graph_store::GraphPg},
 };
-use iris_mpc_store::{loader::load_iris_db, Store as IrisStore};
+use iris_mpc_store::{
+    loader::load_iris_db, Store as IrisStore, StoredIrisRef, StoredIrisVectorRef,
+};
 use std::{
     sync::{atomic::AtomicU64, Arc},
     time::{Duration, Instant},
@@ -155,7 +157,7 @@ pub async fn exec(args: ExecutionArgs, config: Config) -> Result<()> {
     .await?;
     log_info(String::from("Indexation complete."));
 
-    // Phase 3: snapshot.
+    // Phase 3: perform post processing tasks.
     if !args.perform_snapshot {
         log_info(String::from("Snapshot skipped ... as requested."));
     } else {
@@ -196,8 +198,10 @@ async fn exec_setup(
     let mut task_monitor_bg = coordinator::init_task_monitor();
 
     // Set service clients.
-    let ((aws_s3_client, aws_sqs_client, aws_rds_client), (iris_store, graph_store)) =
-        get_service_clients(config).await?;
+    let (
+        (aws_s3_client, aws_sqs_client, aws_rds_client),
+        (iris_store, (hnsw_iris_store, graph_store)),
+    ) = get_service_clients(config).await?;
     log_info(String::from("Service clients instantiated"));
 
     // Set serial identifier of last indexed Iris.
@@ -306,6 +310,9 @@ async fn exec_setup(
         bail!("Shutdown")
         // return Ok(());
     }
+
+    // create copy of iris data
+
     Ok((
         ExecutionContextInfo::new(
             args,
@@ -472,7 +479,6 @@ async fn exec_indexation(
             tx_results.send(result).await?;
             shutdown_handler.increment_batches_pending_completion();
         }
-
         Ok(())
     }
     .await;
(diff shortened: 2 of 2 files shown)


iris-mpc-cpu/src/genesis/batch_generator.rs
```diff
@@ -280,6 +286,25 @@ impl BatchIterator for BatchGenerator {
...
+            let right_store = &imem_iris_stores[RIGHT];
+            let mut codes_and_masks = Vec::new();
+
+            let left_data = left_store.get_vectors(vector_ids.clone().iter()).await;
+            let right_data = right_store.get_vectors(vector_ids.clone().iter()).await;
+            for (i, vector_id) in vector_ids.iter().enumerate() {
+                let left_iris = &left_data[i];
+                let right_iris = &right_data[i];
+
+                codes_and_masks.push(StoredIrisVector {
```
carlomazzaferro, line 299: We return as part of the batch request the iris data to be persisted

iris-mpc-cpu/src/genesis/hawk_job.rs
```diff
@@ -100,25 +109,27 @@ impl JobResult {
...
 
     #[allow(dead_code)]
     pub(crate) fn new_modification_result(
-        modification_id: i64,
+        modification: Modification,
         connect_plans: HawkMutation,
     ) -> Self {
         Self::Modification {
-            modification_id,
+            modification,
```
carlomazzaferro, line 132: Return the full modification, so it can be persisted into hnsw tables

iris-mpc-cpu/src/genesis/batch_generator.rs
```diff
@@ -83,13 +97,6 @@ impl BatchGenerator {
         batch_size: BatchSize,
         exclusions: Vec<IrisSerialId>,
     ) -> Self {
-        assert!(
```
carlomazzaferro, line 86: Incorrect for end_id == start_id, it should simply just do nothing

iris-mpc-cpu/src/genesis/batch_generator.rs
```diff
@@ -207,12 +214,18 @@ impl fmt::Display for BatchSize {
 impl Batch {
     // Returns Iris serial id of batch's last element.
     pub fn id_end(&self) -> IrisSerialId {
-        self.vector_ids.last().map(|id| id.serial_id()).unwrap()
+        self.vector_ids
+            .last()
+            .map(|id| id.serial_id())
+            .unwrap_or_default()
```
carlomazzaferro, line 220: Could be empty in case of a Batch for which `iris_indexation_only` is true, i.e., a batch that only inserts iris codes but does not apply any mutation. This would happen in the case of a batch where all the items are deleted iris codes. We still want to copy those over to the HNSW schema

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -522,8 +536,9 @@ async fn exec_indexation(
             // Coordinator: check background task processing.
             task_monitor_bg.check_tasks();
 
-            // Set next last indexed id.
-            last_indexed_id = batch.id_end();
+            if !batch.iris_indexation_only {
```
carlomazzaferro, line 539: If it is a `batch.iris_indexation_only`, we do not update the `last_indexed_id`

iris-mpc-cpu/src/genesis/batch_generator.rs
```diff
@@ -229,29 +242,83 @@ impl Batch {
...
+        }
 
         // Calculate batch size.
         let batch_size_max = self.batch_size.next_max(last_indexed_id);
 
         // Construct next batch.
+        println!("exclusions={:?}", self.exclusions);
+
         let mut identifiers = Vec::<IrisSerialId>::new();
+        let mut identifiers_for_indexation = Vec::<IrisSerialId>::new();
```
danielle-tfh, line 265: should this be identifiers_for_copying?

iris-mpc-store/src/lib.rs
```diff
@@ -430,6 +496,7 @@ WHERE id = $1;
             RESET_UPDATE_MESSAGE_TYPE,
             REAUTH_MESSAGE_TYPE,
             IDENTITY_DELETION_MESSAGE_TYPE,
+            UNIQUENESS_MESSAGE_TYPE,
```
danielle-tfh, line 499: I think we can remove this? why we would get uniqueness too? This way we can also remove the check for if we get uniqueness modifications

iris-mpc-cpu/src/genesis/hawk_handle.rs
```diff
@@ -246,6 +248,13 @@ impl Handle {
                                     ));
                                     bail!(msg);
                                 },
+                                smpc_request::UNIQUENESS_MESSAGE_TYPE => {
```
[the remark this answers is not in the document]
danielle-tfh: Nice! But we should never get these at all since we do not get them from the iris table

danielle-tfh (commented):
The one thing missing is the overriding of iris data for modifications. ie: a modification means the iris data needs to be altered.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -784,6 +811,7 @@ async fn get_service_clients(
 /// * `shutdown_handler` - Handler coordinating process shutdown.
 ///
 async fn get_results_thread(
+    hnsw_iris_store: IrisStore,
```
danielle-tfh, line 814: This is not the HNSW iris_store right? its the normal GPU iris store?

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -797,29 +825,43 @@ async fn get_results_thread(
                     batch_id,
                     connect_plans,
                     last_serial_id,
+                    iris_data,
+                    iris_indexation_only,
                     ..
                 } => {
                     log_info(format!("Job Results :: Received: batch-id={batch_id}"));
 
                     let mut graph_tx = graph_store.tx().await?;
```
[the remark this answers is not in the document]
carlomazzaferro: I don't think we can, or rather, the point of using the `graph_tx` is that it allows us to represent
A `graph_tx` is:

```
pub struct GraphTx<'a, V> {
    pub tx: Transaction<'a, Postgres>,
    schema_name: String,
    phantom: PhantomData<V>,
}
```

So it enables us to represent both a graph tx, and a generic Postgres one. Since we are interested in applying both a graph, and a plain pg tx in the same db commit, we should use the graph_tx for that

iris-mpc-cpu/src/genesis/batch_generator.rs
```diff
@@ -229,29 +242,83 @@ impl Batch {
...
         while self.range_iter.peek().is_some() && identifiers.len() < batch_size_max {
```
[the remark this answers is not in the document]
danielle-tfh: But it may be better to remove the new storage types and get the irises in the results thread. Lets see what that looks like

iris-mpc-store/src/lib.rs
```diff
@@ -98,6 +98,47 @@ pub struct StoredIrisRef<'a> {
...
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct StoredIrisVector {
+    pub id: i64,
+    pub version_id: i16,
+    pub left_code: Vec<u16>,
+    pub left_mask: Vec<u16>,
+    pub right_code: Vec<u16>,
+    pub right_mask: Vec<u16>,
+}
```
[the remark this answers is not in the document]
danielle-tfh: I agree - lets rather see if we can remove the need for these all together.

iris-mpc-cpu/src/genesis/batch_generator.rs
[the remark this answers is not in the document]
danielle-tfh: Hmm - I do not think we would fall under any constraint with regards to this? @carlomazzaferro any thoughts here?

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -797,39 +825,61 @@ async fn get_results_thread(
...
-                    shutdown_handler_bg.decrement_batches_pending_completion();
+                    // Persist batch of Iris's to the HNSW graph store.
+                    hnsw_iris_store
+                        .insert_copy_iris(
+                            &mut graph_tx.tx,
+                            &iris_data.iter().map(|i| i.as_ref()).collect::<Vec<_>>(),
+                        )
+                        .await?;
+                    if iris_copy_only {
+                        graph_tx.tx.commit().await?;
```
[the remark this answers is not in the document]
danielle-tfh: Good point

iris-mpc-cpu/src/genesis/batch_generator.rs
```diff
@@ -262,14 +253,11 @@ impl BatchGenerator {
...
+        let mut identifiers_for_copying = Vec::<IrisSerialId>::new();
         while self.range_iter.peek().is_some() && identifiers.len() < batch_size_max {
             let next_id = self.range_iter.by_ref().next().unwrap();
-            identifiers_for_indexation.push(next_id);
-            println!(
-                "identifiers_for_indexation={:?}",
-                identifiers_for_indexation
-            );
+            identifiers_for_copying.push(next_id);
+            println!("identifiers_for_copying={:?}", identifiers_for_copying);
```
carlomazzaferro, line 260: Can probably remove the leftover println

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -536,7 +537,8 @@ async fn exec_indexation(
             // Coordinator: check background task processing.
             task_monitor_bg.check_tasks();
 
-            if !batch.iris_copy_only {
+            // copy persisted only
+            if !batch.vector_ids.is_empty() && batch.vector_ids_to_persist.is_empty() {
```
carlomazzaferro, line 541: Can it ever be the case that it evaluates to true? AFAICT, we can only have the opposite, `batch.vector_ids()` empty and `batch.vector_ids_to_persist.is_empty()` not empty, which is the case in which we do iris indexation only

iris-mpc-cpu/src/genesis/batch_generator.rs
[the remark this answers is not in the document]
carlomazzaferro: Okay, had another chat with @danielle-tfh and we understood what you meant - if we have, say, 100 deletions, then we'd need to index 100 + whatever else would come in the batch. I still don't think that would be a problem, as deletions happen quite scattered in time.

At the same time, this also pointed out the fact that we will never have the case in which `identifiers` is empty and `identifiers_for_indexation` isn't, because deletions do not count towards the batch size. This could simplify a bit things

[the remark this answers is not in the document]
bgillesp: My model for how we're limiting things is that the genesis batch size describes the "max number of HNSW insertions being conducted in parallel to manage bandwidth and compute resources", so we should be aiming to run at the full batch size number of non-deleted iris codes in a batch whenever possible.  So the procedure would be to "fill in batch iris codes until `batch_size` non-deleted iris codes are included in the batch, then process" -- meaning that any deleted iris codes are included on top of this maximum batch size.

Extending the batch in this way is almost surely not an issue with the number/distribution of deletions in the actual data, but in principle a long stretch of deleted ids could produce a "batch" with a large number of iris codes for copying (but not insertion).  The only reason this is a problem is that we had some stability issues when running `init-test-dbs` when we were inserting large numbers of iris codes into Postgres in the same transaction (order of thousands), so needed to limit this size.  Practically it shouldn't be necessary, but just noted the edge case since it's a quick conditional to check.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
bgillesp: Do you mean when we do iris copying only?  I've been using "indexation" to mean "insertion into graph" and "copying" to mean "adding iris shares data without insertion into graph".

I think that the invariant we maintain on batches right now is that "vector ids to copy/persist is always non-empty, and vector ids to index/insert into the HNSW graph is a subset of the copy/persist ones".

danielle-tfh: After discussion with @carlomazzaferro we realised this would never be the case that it would be a iris_copy_only. you are right about the definition above Bryan :) Thanks

bgillesp: I think in principle `iris_copy_only` could happen?  Specifically, if the indexation range ends with an interval of one or more deleted elements, and the batch starts in this interval.  But there definitely shouldn't be batches which don't copy any irises.

iris-mpc-cpu/src/genesis/hawk_job.rs
```diff
@@ -100,26 +108,30 @@ impl JobResult {
         batch_id: usize,
         vector_ids: Vec<IrisVectorId>,
         connect_plans: HawkMutation,
+        vector_ids_to_persist: Vec<IrisVectorId>,
     ) -> Self {
         let first_serial_id = vector_ids.first().unwrap().serial_id();
         let last_serial_id = vector_ids.last().unwrap().serial_id();
```
bgillesp, line 114: Should probably be taking these values from `vector_ids_to_persist` since this represents all of the serial ids which are "processed" by this job.

siajasl: Aggree with Bryan's comment.

iris-mpc-cpu/src/genesis/hawk_handle.rs
```diff
@@ -216,42 +218,54 @@ impl Handle {
...
                                         modification,
                                     ));
-                                    bail!(msg);
+                                    Ok((
+                                        vec![],
+                                        IrisVectorId::new(
+                                            serial_id,
+                                            0,
+                                        ),
+                                    ))
```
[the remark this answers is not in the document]
danielle-tfh: Good point - let me update that

PR #1467, opened 2025-06-23
aurel/refactor-batch: Group related "pushes" into methods

naure:
Follow-up to #1464.

Group all the `push` to `BatchQuery` fields into methods. This makes it clear how the fields are related and it avoids forgetting some or pushing at the wrong place.

Exhibit A: There was one too many `metadata` push in `process_reset_update_request`, fixed in this PR.

### DIFF
--- a/iris-mpc-common/src/job.rs
+++ b/iris-mpc-common/src/job.rs
@@ -108,6 +108,49 @@ pub struct BatchQuery {
     pub full_face_mirror_attacks_detection_enabled: bool,
 }
 
+impl BatchQuery {
+    /// Add a Uniqueness, Reauth, or Reset Check request to the batch.
+    pub fn push_matching_request(
+        &mut self,
+        request_id: String,
+        request_type: &str,
+        metadata: BatchMetadata,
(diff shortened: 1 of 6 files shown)


naure:
@eaypek-tfh

There is also the field `sns_message_ids` but it is not used consistently at the moment: all requests in GPU code but only reset updates in CPU code.

```rust
batch_query.sns_message_ids.push(sns_message_id);
```

This should be consolidated to all requests, right?

eaypek-tfh:
> @eaypek-tfh
> 
> There is also the field `sns_message_ids` but it is not used consistently at the moment: all requests in GPU code but only reset updates in CPU code.
> 
> ```rust
> batch_query.sns_message_ids.push(sns_message_id);
> ```
> 
> This should be consolidated to all requests, right?
> 
> 👉 Done in #1468

Yeap, as mentioned there too, we use it to calculate batch hash in GPU and crash when they differ. We'd need same logic in HNSW too 👍

eaypek-tfh (approved):
Thank you! Much cleaner now 💯

PR #1468, opened 2025-06-23
Track SNS ID of all requests

naure:
Follow-up to #1467.

- [x] Refactor sns_message_id into the BatchQuery methods.
- [x] Include the SNS message ID of all requests (CPU code, GPU already did).


eaypek-tfh (approved):
Nice! This was needed 👍 

This is used to calculate batch hash and crash if they differ in GPU. It'd be nice to do so in HNSW too actually

PR #1471, opened 2025-06-24
[POP-2614] Implement parallel quicksort for HNSW search

bgillesp:
This PR implements parallel partially-sorted quicksort to replace the current static sorting network used during HNSW graph traversal.  This algorithm keeps track of partially sorted sub-intervals during the quicksort recursion, and takes advantage of sorted regions to pick high quality central pivot elements.

The current algorithm used to construct ordered graph neighborhoods during HNSW graph traversal uses a static sorting network based on [Batcher's Odd-Even Merge Sort](https://en.wikipedia.org/wiki/Batcher_odd%E2%80%93even_mergesort).  This approach is efficient among static sorting networks, but uses more overall comparisons and has substantially higher sequential depth than efficient dynamic sorting algorithms.  An improvement across the board can be achieved by using parallelized quicksort instead -- expected improvements for inserting new items into a sorted list based on Python simulations are on the order of a 75-80% reduction in both total comparisons and sequential depth.

### DIFF
--- a/iris-mpc-cpu/src/hnsw/graph/neighborhood.rs
+++ b/iris-mpc-cpu/src/hnsw/graph/neighborhood.rs
@@ -5,7 +5,7 @@
 
 use crate::hnsw::{
     sorting::{
-        batcher::partial_batcher_network, binary_search::BinarySearch,
+        batcher::partial_batcher_network, binary_search::BinarySearch, quicksort::apply_quicksort,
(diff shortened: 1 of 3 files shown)


bgillesp:
Context note: force-pushed a new (identical) version of the branch with squashed commits to simplify making all commits verified w/ valid signature.

PR #1474, opened 2025-06-30
Release v0.20.1 to prod with bumped max db size

eaypek-tfh:
## Change
- Bump prod version to v0.20.1
- Bump max db size


leonanos8 (approved):
🚢

PR #1479, opened 2025-07-01
aurel/fix-equal-threshold

naure:


naure:
@iliailia Can you check again?

iliailia (approved):
Looks great!

PR #1481, opened 2025-07-02
[POP-2513] Make mask shares full length again

dkales:
This is in preperation for the fragile bits effort for iris codes. Up until now, mask code shares have had the same imaginary and real part, which meant that we could reduce the storage and memory costs by only storing half and multiplying the results by 2. This PR extends them to the full length again, such that future iris codes may have different real and imaginary parts. 

The DB and S3 loader are modified in a way that they extend half-sized mask codes to full ones if needed.

Also: DO NOT MERGE FOR NOW, this may need some sync for correct DB exporting in future.

Impact of merging this:
- Will increase memory requirements for InMemoryStore of IrisDB from 38.4kB to 51.2kB per iris code (a 33% increase over current).
- Processing time will increase slightly (mask dot products go from size 6400 to size 12800). For GPU, this time is relatively negligible compared to the remainder of the MPC protocol. For CPU, I do not have current breakdowns but I would also expect this to be not that significant.
- Will enable iris codes to have masks with different real/imaginary part in the future. The request shares currently sent via SNS are already in extended format.

### DIFF
--- a/iris-mpc-cpu/tests/e2e.rs
+++ b/iris-mpc-cpu/tests/e2e.rs
@@ -109,11 +109,7 @@ async fn start_hawk_node(
     Ok(handle)
 }
 
-#[ignore = "Expected to fail for now"]
-#[tokio::test]
-async fn e2e_test() -> Result<()> {
-    install_tracing();
-
+async fn e2e_test_main() -> Result<()> {
     let test_db = generate_full_test_db(DB_SIZE, DB_RNG_SEED);
     let db_left = test_db.plain_dbs(0);
     let db_right = test_db.plain_dbs(1);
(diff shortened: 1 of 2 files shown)


dkales:
@bgillesp @naure I tried to also handle changes in the HNSW variant, please take another look if I missed something

bgillesp (commented):
Leaving this review as a Comment to help avoid accidental merge, but the PR looks fine to me.  I can't speak to all of the changes in `iris-mpc-gpu`, but the changes in `iris-mpc-cpu` seem to cover all of the relevant touch points with the mask shares.

iris-mpc-gpu/src/server/actor.rs
```diff
@@ -3365,10 +3363,24 @@ impl InMemoryStore for ServerActor {
             self.device_manager.device_count(),
             IRIS_CODE_LENGTH,
         );
+
+        // Pad partial mask if needed, since the DB may contain partial masks, where the imaginary and real parts are the same and deduplicated.
```
leonanos8, line 3367: So with this handling here we can continue without performing any migration for the existing mask codes right?

dkales: For the DB masks, yes, however the S3 ones have hard-coded length checks that are now wrong and will need some handling

PR #1485, opened 2025-07-03
Fix saving prev modification

danielle-tfh:
### Notes
* Fixes bug in Genesis. During the first Genesis run the last modification ID is not saved since there is no delta. This means for the second run, we perform all modifications but there is no need since the Genesis included them.
* This PR then saves the largest modification id if there is no delta

### DIFF
--- a/iris-mpc-cpu/src/genesis/state_sync.rs
+++ b/iris-mpc-cpu/src/genesis/state_sync.rs
@@ -22,6 +22,9 @@ pub struct Config {
 
     // Identifier of the last modification ID to be indexed.
     pub max_modification_id: i64,
+
+    // Identifier of the max completed modification the node has seen
+    pub max_completed_modification: i64,
 }
 
 /// Constructor.
@@ -33,6 +36,7 @@ impl Config {
         last_indexed_id: IrisSerialId,
         max_indexation_id: IrisSerialId,
         max_modification_id: i64,
+        max_completed_modification: i64,
     ) -> Self {
         Self {
             batch_size,
@@ -41,6 +45,7 @@ impl Config {
             last_indexed_id,
             max_indexation_id,
             max_modification_id,
+            max_completed_modification,
         }
     }
 }
@@ -107,10 +112,10 @@ mod tests {
 
     impl Config {
         fn new_1() -> Self {
-            Self::new(64, 128, vec![3, 5], 50, 100, 100)
+            Self::new(64, 128, vec![3, 5], 50, 100, 100, 100)
         }
         fn new_2() -> Self {
-            Self::new(64, 128, vec![3, 5, 6], 150, 200, 200)
+            Self::new(64, 128, vec![3, 5, 6], 150, 200, 200, 200)
         }
     }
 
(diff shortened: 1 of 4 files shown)


iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -332,6 +334,13 @@ async fn exec_setup(
     let hawk_handle = GenesisHawkHandle::new(hawk_actor).await?;
     log_info(String::from("Hawk handle initialised"));
 
+    if modifications.is_empty() {
+        log_info(String::from("No modifications to apply. Therefore setting the last modification id as the largest completed modification id."));
+        let mut graph_tx = graph_store.tx().await?;
+        set_last_indexed_modification_id(&mut graph_tx.tx, max_completed_modification_id).await?;
+        graph_tx.tx.commit().await?;
+    }
```
bgillesp, line 342: It might be beneficial to set this value unconditionally at the end of delta (e.g. around current line 432 after any modifications are processed) and to remove the check `if ctx.modifications.is_empty()` on line 145 entirely so that delta runs even with empty modifications list.  This is because more modifications may be persisted/completed than are processed by the "relevant" modifications in `ctx.modifications`, so the last completed modification can always be updated to this `max_completed_modification_id` value.  (But this should only be done after the `ctx.modifications` entries have been processed.)

iris-mpc-store/src/lib.rs
```diff
@@ -522,18 +523,33 @@ WHERE id = $1;
...
+        .await?;
+
+        // Fetch the max id from modifications that are persisted and completed.
+        let max_id: Option<i64> = sqlx::query_scalar(
+            r#"
+            SELECT MAX(id) FROM modifications
+            WHERE persisted = true
+                AND status = 'COMPLETED'
+            "#,
+        )
```
bgillesp, line 544: Small issue I see with this query -- right now this catches all types of modifications that have been persisted with status `'COMPLETED'`, but our guarantee of consistency for reading from the production database is only for reauth, reset, and deletion modification entries (i.e. the plan is to pause submission of these request types to prod for a limited time while genesis is starting up).  It could happen that other modifications entries are actively being inserted into the database while genesis is loading, so parties could end up seeing different values for `MAX(id)`.  I think this should be fixed by just limiting this query with `AND request_type = ANY($1)` with the same `message_types` as above.

bgillesp (dismissed):
Nice find on this issue!  Looking good overall -- I think there's a small synchronization issue in the query selecting the "global max persisted id" that looks easy to fix, and I had one suggestion about where `set_last_indexed_modification_id` is called with the non-filtered maximum value.

iris-mpc-store/src/lib.rs
danielle-tfh: Good catch! Thank you :)

iris-mpc-cpu/src/genesis/state_sync.rs
```diff
@@ -22,6 +22,9 @@ pub struct Config {
 
     // Identifier of the last modification ID to be indexed.
     pub max_modification_id: i64,
+
+    // Identifier of the max completed modification the node has seen
+    pub max_completed_modification: i64,
```
carlomazzaferro, line 27: Possibly calling this `max_completed_modification_id` would indeed make it clearer

PR #1487, opened 2025-07-04
Add check to ensure latest persisted index matches the height of the …

danielle-tfh:
### Notes
* ensure consistency of graph and persisted last id

### DIFF
--- a/iris-mpc-cpu/src/hnsw/graph/graph_store.rs
+++ b/iris-mpc-cpu/src/hnsw/graph/graph_store.rs
@@ -605,7 +605,7 @@ mod tests {
         let mut graph_ops = tx.with_graph(StoreId::Left);
 
         // Insert links for each vector, using the same pattern as test_db
-        for i in 1..4 {
+        for i in 1..5 {
             let mut links = SortedNeighborhood::new();
             for j in 0..5 {
                 if i != j {
@@ -622,8 +622,8 @@ mod tests {
 
         let max_id = graph_ops.get_max_serial_id().await?;
         // serial_id is stored as u32, but we inserted 5 vectors, so largest should be >= 4
-        assert!(max_id >= 4);
-
+        assert!(max_id >= 5);
+        tx.tx.commit().await?;
         graph.cleanup().await?;
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


iris-mpc-cpu/src/hnsw/graph/graph_store.rs
```diff
@@ -550,6 +564,70 @@ mod tests {
...
+            let links = links.edge_ids();
+            graph_ops.set_links(vectors[i], links.clone(), 0).await?;
+            let links2 = graph_ops.get_links(&vectors[i], 0).await?;
+            assert_eq!(*links, *links2);
+        }
+
+        let max_id = graph_ops.get_max_serial_id().await?;
+        // serial_id is stored as u32, but we inserted 5 vectors, so largest should be >= 4
+        assert!(max_id >= 4);
+
```
bgillesp, line 626: Need a line like `tx.tx.commit().await?` here or else the test hangs.

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
```diff
@@ -550,6 +564,70 @@ mod tests {
...
+            }
+            let links = links.edge_ids();
+            graph_ops.set_links(vectors[i], links.clone(), 0).await?;
+            let links2 = graph_ops.get_links(&vectors[i], 0).await?;
+            assert_eq!(*links, *links2);
+        }
+
+        let max_id = graph_ops.get_max_serial_id().await?;
+        // serial_id is stored as u32, but we inserted 5 vectors, so largest should be >= 4
+        assert!(max_id >= 4);
```
bgillesp, line 625: Serial ids are 1-indexed throughout when represented in the DB, so `max_id >= 5` should test what we want here.  (Confirmed with a `println` that the current implementation gets `max_id` equal to 5 at this line.)

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
```diff
@@ -550,6 +564,70 @@ mod tests {
...
+                d.push(vector_store.eval_distance(&q, v).await?);
+            }
+            d
+        };
+
+        let mut tx = graph.tx().await?;
+        let mut graph_ops = tx.with_graph(StoreId::Left);
+
+        // Insert links for each vector, using the same pattern as test_db
+        for i in 1..4 {
```
bgillesp, line 608: Maybe `i in 0..5`?

bgillesp (approved):
Looks good, thank you!  One of the unit tests hangs currently because a Postgres transaction isn't committed, but that's a one-line fix.

PR #1489, opened 2025-07-04
Fix last persisted delta modification to always be stored

danielle-tfh:
### Notes
* always run delta to ensure we store max modification.


bgillesp (approved):
Looks good!  Took a quick look through, and the logic for setting the max indexed modification id seems right to me now.

PR #1491, opened 2025-07-04
Genesis skip deletions for staging

danielle-tfh:
### Notes
* ignore deletions for staging

### DIFF
--- a/iris-mpc-upgrade-hawk/src/genesis/mod.rs
+++ b/iris-mpc-upgrade-hawk/src/genesis/mod.rs
@@ -9,7 +9,10 @@ use chrono::Utc;
 use eyre::{bail, eyre, Report, Result};
 use iris_mpc_common::{
     config::{CommonConfig, Config},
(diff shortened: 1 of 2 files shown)


iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -422,6 +425,20 @@ async fn exec_delta(
                 "Applying modification: type={} id={}, serial_id={:?}",
                 modification.request_type, modification.id, modification.serial_id
             ));
+            if modification.request_type == smpc_request::IDENTITY_DELETION_MESSAGE_TYPE {
+                if ctx.config.environment != "prod" {
```
siajasl, line 429: As "prod" is a constant ... imho would be best to define it as such ... maybe in `iris-mpc-common.`

siajasl (approved):
One minor nitpick.  But approving anyway as nitpick is perhaps open to debate.

PR #1496, opened 2025-07-08
pass around batch sha and valid entries

carlomazzaferro:


iris-mpc-common/src/helpers/batch_sync.rs
```diff
@@ -11,6 +13,49 @@ pub struct BatchSyncState {
...
+            all_states,
+        }
+    }
+    pub fn sha_matches(&self) -> bool {
+        self.all_states
+            .iter()
+            .all(|s| s.batch_sha == self.my_state.batch_sha)
+    }
+
+    pub fn valid_entries(&self) -> Vec<bool> {
```
carlomazzaferro, line 41: - For each party’s vector, iterate over its entries; whenever any party has false at index i, set the accumulator at i to false.
- After folding all states, the accumulator contains true only for indices that were true in every party.

iris-mpc/src/server/mod.rs
```diff
@@ -583,6 +589,57 @@ async fn run_main_server_loop(
                 Some(Ok(Some(batch))) => batch,
             };
 
+            let batch_hash = sha256_bytes(batch.sns_message_ids.join(""));
+            let batch_valid_entries = batch.valid_entries.clone();
+
+            tracing::info!("Current batch hash: {}", hex::encode(&batch_hash[0..4]));
```
eaypek-tfh, line 595: [nit] shall we isolate the batch sync part in server into its own method or maybe consider moving this to the batch itself so we can call `batch.retain_after_sync(params)`?

eaypek-tfh (commented):
Very nice and clean! 🙌 

I didn't thought of valid batch entries sync when filing the ticket, good observation to include that as well! 💪 

I checked further on it and I couldn't find a place where batch entries are actually updated based on the `valid_entries` in the batch. It's like HNSW does not even make use of it. I see that [create_dummy_shares](https://github.com/worldcoin/iris-mpc/blob/f79d4ad9c0fbc409ca7851fc3a0741c581ea3ecd/iris-mpc/src/services/processors/batch.rs#L870-L886) is called in local party if it fails to parse shares [here](https://github.com/worldcoin/iris-mpc/blob/f79d4ad9c0fbc409ca7851fc3a0741c581ea3ecd/iris-mpc/src/services/processors/batch.rs#L724-L725) but the other parties could parse it and this could cause problems 🤔 

I see that GPU is getting rid of the invalid entries fully in all parties [here](https://github.com/worldcoin/iris-mpc/blob/f79d4ad9c0fbc409ca7851fc3a0741c581ea3ecd/iris-mpc-gpu/src/server/actor.rs#L837). Maybe we can do smt similar or just somehow ignore invalid entries? 🤔

carlomazzaferro:
> Very nice and clean! 🙌
> 
> I didn't thought of valid batch entries sync when filing the ticket, good observation to include that as well! 💪
> 
> I checked further on it and I couldn't find a place where batch entries are actually updated based on the `valid_entries` in the batch. It's like HNSW does not even make use of it. I see that [create_dummy_shares](https://github.com/worldcoin/iris-mpc/blob/f79d4ad9c0fbc409ca7851fc3a0741c581ea3ecd/iris-mpc/src/services/processors/batch.rs#L870-L886) is called in local party if it fails to parse shares [here](https://github.com/worldcoin/iris-mpc/blob/f79d4ad9c0fbc409ca7851fc3a0741c581ea3ecd/iris-mpc/src/services/processors/batch.rs#L724-L725) but the other parties could parse it and this could cause problems 🤔
> 
> I see that GPU is getting rid of the invalid entries fully in all parties [here](https://github.com/worldcoin/iris-mpc/blob/f79d4ad9c0fbc409ca7851fc3a0741c581ea3ecd/iris-mpc-gpu/src/server/actor.rs#L837). Maybe we can do smt similar or just somehow ignore invalid entries? 🤔

Interestingly, this comment: https://github.com/worldcoin/iris-mpc/blob/19767f7035cf124293915160a07991a169bc514f/iris-mpc-common/src/job.rs#L76-L79  hints at `valid_entries` being synced between parties, but I don't see where this is happening.

@naure any thoughts? We are designing right now the methodology to filter them out before the batch is processed

naure:
Hi, nice validation!

I don’t know about the `valid_entries` mechanism. Should I look into it?

carlomazzaferro:
> Hi, nice validation!
> 
> I don’t know about the `valid_entries` mechanism. Should I look into it?

I think it'd be good to understand how the GPU uses it, and the fact that it seems like we don't. We set the array if we fail to process, but we do not act on it. 

Otherwise, I've redesigned how this works. There can be more than 1 batch in flight, so it is possible that the batch hashes won't match when the nodes communicate. To mitigate this, I introduced a `HashMap` to hold the batch sha and valid entries for the inflight batches, and then remove the items when the batch is processed

@eaypek-tfh

eaypek-tfh:
> I think it'd be good to understand how the GPU uses it, and the fact that it seems like we don't. We set the array if we fail to process, but we do not act on it.

Re GPU, this is what I wrote in earlier comment (please also see the source of `retain` method [here](https://github.com/worldcoin/iris-mpc/blob/f79d4ad9c0fbc409ca7851fc3a0741c581ea3ecd/iris-mpc-gpu/src/server/mod.rs#L277)):

> I see that GPU is getting rid of the invalid entries fully in all parties [here](https://github.com/worldcoin/iris-mpc/blob/f79d4ad9c0fbc409ca7851fc3a0741c581ea3ecd/iris-mpc-gpu/src/server/actor.rs#L837). Maybe we can do smt similar or just somehow ignore invalid entries? 🤔

> There can be more than 1 batch in flight, so it is possible that the batch hashes won't match when the nodes communicate

Why do you think there can be more than 1 batch in flight? 🤔 We set the mpsc limit to 1 [here](https://github.com/worldcoin/iris-mpc/blob/f79d4ad9c0fbc409ca7851fc3a0741c581ea3ecd/iris-mpc/src/services/processors/batch.rs#L57)

PR #1505, opened 2025-07-10
Only load iris database up to max index id

danielle-tfh:
### Notes
* only load up to the specified max index for iris store
* also fixes the CI for context see https://github.com/worldcoin/iris-mpc/pull/1509


siajasl (approved):
LGTM

PR #1525, opened 2025-07-21
INFRA-4966 Change HNSW stage instance to x2iedn.8xlarge

kszarek:
At the last PoP meeting, we agreed to change the instances for testing purposes.

### DIFF
--- a/deploy/stage/common-values-ampc-hnsw.yaml
+++ b/deploy/stage/common-values-ampc-hnsw.yaml
@@ -35,11 +35,11 @@ startupProbe:
 
 resources:
   limits:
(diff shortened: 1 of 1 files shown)


jazzbee (approved):

deploy/stage/common-values-ampc-hnsw.yaml
```diff
@@ -50,7 +50,7 @@ podAnnotations:
 nodeSelector:
   kubernetes.io/arch: amd64
```
eaypek-tfh, line 51: do we still keep `kubernetes.io/arch: amd64`?

PR #1530, opened 2025-07-22
add check to verify that the PRF didnt get out of sync

sdwoodbury:
During testing in the staging environment, unusual values were observed in the anonymous statistics. It was desired to validate that the PRFs didn't get out of sync. This check will accomplish this. This may be only needed for testing or it may get added to the health check.


sdwoodbury:
@dkales  sry, didn't notice the comment until now i will fix this up and get it merged in

PR #1531, opened 2025-07-23
testing securty with fake tokens and junk ssh key

orlra:
https://linear.app/worldcoin/issue/INFRA-4811/secret-scanning-and-audit
test, dont merge.


orlra:
this was a test, so closing.

PR #1536, opened 2025-07-24
[POP-2699] Add TCP keepalive and metrics for reconnect

sdwoodbury:
Unexpected disconnects were observed during testing in staging with NGINX TLS sidecar. To troubleshoot this, this PR does the follwing:
- enables TCP keepalive after 30 seconds of inactivity
- adds metrics for reconnect
- adds metrics for creating new sessions
- changes the logging in `network/tcp/handle.rs` to  warn when the inbound or outbound forwarder returns early instead of when a message is received for a session that is no longer in use (this is a symptom of making new sessions)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2073,7 +2073,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
- "socket2",
+ "socket2 0.5.7",
  "tokio",
  "tower-service",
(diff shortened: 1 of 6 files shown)


dkales (approved):
LGTM, only consideration on my side would be to maybe use https://lib.rs/crates/socket2 for the socket options, as it is multi-platform by default.

PR #1537, opened 2025-07-24
[POP-2700] allow multiple root TLS certs

sdwoodbury:
It was requested that for application level TLS, a different root cert be used for each party. This PR makes that possible. However, this implementation seems no more secure than using 1 cert because any of the root certs provided could be used to sign any leaf cert. 

to set the environment variable:
OLD  
`export SMPC__TLS__ROOT_CERT="certs/root.crt"`
NEW  
`export SMPC__TLS__ROOT_CERTS='["certs/root1.crt","certs/root2.crt","certs/root3.crt"]'`


wojciechsromek:
I think we can close this one, thank you @sdwoodbury :pray:

PR #1539, opened 2025-07-26
bench-dot: Benchmark for galois_ring_pairwise_distance

naure:
- Benchmark the computation of distances.
- Two variants, with and without CPU cache.
- Single-threaded version.
    - Without batching. I found no impact of `batch_size`.
- Multi-threaded version.
    - With batching to dilute scheduling overhead.

```bash
cargo bench -p iris-mpc-cpu --bench dot
```


naure:
I made it `batch_size=100` as a realistic size.

PR #1555, opened 2025-08-04
Finish genesis 100 and 101 tests

sdwoodbury:
- add a common `MpcNode` struct which holds the CPU iris, graph, modifications, and persisted state tables, and the GPU iris table, with helper functions to clear and initialize said tables. 
- execute plaintext genesis and compare the graph output to `exec_genesis()`
- ensure s3 deletions are empty for the 100 test
- gate behind the `db_dependent` feature
- ensure the genesis tests run in serial


iris-mpc-upgrade-hawk/tests/workflows/genesis_101/runner.rs
```diff
@@ -0,0 +1,233 @@
...
+                .await
+            });
+        }
+
+        while let Some(r) = join_set.join_next().await {
+            r.unwrap()?;
+        }
+
+        // hack to get around an "address already in use" error, emitted by HawkHandle
+        let service_ports = vec!["4003".into(), "4004".into(), "4005".into()];
```
bgillesp, line 80: Odd issue.  Fix works for now, and maybe not even worth trying to hunt down the cause since "repeated execution within a single process" is not a common use case of this code.

iris-mpc-upgrade-hawk/tests/workflows/genesis_101/runner.rs
```diff
@@ -0,0 +1,233 @@
...
+        Ok(())
+    }
+
+    async fn exec_assert(&mut self) -> Result<(), TestError> {
+        let config = &self.configs[0];
+        let plaintext_irises = get_irises();
+        let expected = Arc::new(
+            MpcNode::simulate_genesis(DEFAULT_GENESIS_ARGS, config, &plaintext_irises)
+                .await
+                .unwrap(),
```
bgillesp, line 114: Small tweak here for accuracy -- this should simulate genesis execution twice with `max_indexation_id` set to 50 and then 100 respectively.  The intended usage of the plaintext genesis function call is that it consumes an input state struct and produces an output state struct, and then the args of the output state struct can be modified before feeding it back into the plaintext genesis function again.  If we want to just merge for now, this change can wait for a later PR, as the results of the two different versions of genesis execution should be the same in this case anyway.

bgillesp (commented):
This is looking great, thank you!  Just a few minor comments then should be good to merge.

bgillesp (approved):
Good to go now!

PR #1562, opened 2025-08-07
[POP-2778] fix prf check

sdwoodbury:
prf check fails to receive a message due to a timeout. this is blocking staging. somehow state_check() succeeds even though it is doing basically the same thing. 

attempt to fix the issue by checking only 2 sessions instead of all of them

also update logging to make `squeeze_challenge()` only return an error and not emit logs. emitting 128 logs per MPC node per `prf_check()` is too noisy.


sdwoodbury:
@dkales , how about doing PRF check on every session but only logging the error once instead of once per session? i agree it's hard to believe that this is blocking staging. it does spam the logs too much in the event of an error though.

PR #1566, opened 2025-08-08
Fix dependency compatibility issue in CPU Integration Tests CI/CD workflow

bgillesp:
This PR fixes an issue with the "CPU Integration Tests" CI/CD workflow, where docker compilation failed because a new release of the `cargo-edit` utility updated to use dependencies which require `rustc` version 1.86.0 or greater, while our development environment uses 1.85.0.

Modifying the Dockerfile to fix the version of `cargo-edit` to the previous release resolves the issue for now, but we might consider revising our deployment rust to 1.89.0 at some point, to avoid issues that might come up as a result of using older versions.


bgillesp:
Without any changes from `main`, the workflow still fails here, complaining that some updated dependencies during installation of `cargo-edit` aren't compatible with rustc 1.85.0:

```
7.462 error: failed to compile `cargo-edit v0.13.7`, intermediate artifacts can be found at `/tmp/cargo-installaRCFzz`.
7.462 To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
7.462 
7.462 Caused by:
7.462   rustc 1.85.0 is not supported by the following packages:
7.462     cargo-util-schemas@0.8.2 requires rustc 1.86
7.462     cargo_metadata@0.21.0 requires rustc 1.86.0
7.462     clap-cargo@0.16.0 requires rustc 1.86
7.462   Try re-running `cargo install` with `--locked`
```

Now trying to see if running this installation with the `--locked` flag as suggested fixes the issue for now.  Would we consider updating the project to Rust 1.89.0, the most recent stable release?

bgillesp:
I just confirmed that `cargo-edit` released a new version `0.13.7` with release notes "Dependency updates" just earlier today.  Now trying to explicitly install the version prior to this, `0.13.6`, to see if this resolves the issue for now.

carlomazzaferro (approved):

PR #1569, opened 2025-08-12
[POP-2714] Implement delta protocol for plaintext genesis indexer

bgillesp:
This PR adds the genesis delta sub-protocol to plaintext genesis implementation, providing support for processing modifications present in the source database.  It additionally fixes a small bug in plaintext state initialization, where the in-memory iris stores were initialized from the destination database, but should be initialized instead from the source database.

### DIFF
--- a/iris-mpc-cpu/src/genesis/plaintext.rs
+++ b/iris-mpc-cpu/src/genesis/plaintext.rs
@@ -1,7 +1,9 @@
 use std::{collections::HashMap, sync::Arc};
 
-use eyre::{OptionExt, Result};
-use iris_mpc_common::{iris_db::iris::IrisCode, IrisSerialId, IrisVectorId, IrisVersionId};
+use eyre::{bail, OptionExt, Result};
+use iris_mpc_common::{
+    helpers::smpc_request, iris_db::iris::IrisCode, IrisSerialId, IrisVectorId, IrisVersionId,
+};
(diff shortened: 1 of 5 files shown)


iris-mpc-cpu/src/genesis/plaintext.rs
siajasl, line 1: General comment in respect of this code file.  Is this intended only to be executed from e2e test engine ?  If so then I would propose moving it there.

bgillesp: Good question -- the intent for this file is to give a "reference implementation" of genesis, i.e. in the absence of complex service configuration and MPC protocols, what is the "intended" plaintext computation being done by the executable.  I would like to leave it in the same context as the full server implementation, as I think that this is of independent utility in the codebase as a sanity check and reference point for the simple logic being executed.  The e2e tests then help to ensure the invariant that "functionality of the genesis executable is synchronized with the functionality described by the plaintext reference implementation".  I might add a module level doc comment to the file to clarify this intent.

siajasl: OK so this is a lower level of test.  I can understand it's utility but it is still test logic.  Perhaps annotate with a test flag so that it does not form part of the main binary footprint.  Have approved anyway.

PR #1573, opened 2025-08-13
[POP-2717] Implement genesis 104 e2e test

bgillesp:
This PR integrates the updates to the plaintext genesis simulation from #1569 (modifications support) with the genesis e2e tests, and implements the genesis 104 e2e test exercising handling of modifications applied between two runs of the protocol.


siajasl (approved):
Approved.  Will circle back in 1582 review.

PR #1575, opened 2025-08-14
au/simplify-eval: In eval_distance_batch, support a single query

naure:
Separate PR to simplify an API that’s being reimplemented.


mcalancea (approved):
Looks good to me!

PR #1576, opened 2025-08-14
Iris worker pool

naure:
*Includes #1575*

*Supersedes #1565*


iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -0,0 +1,188 @@
...
+        next_counter: Arc::new(AtomicU64::new(0)),
+    }
+}
+
+fn worker_thread(ch: Receiver<IrisTask>, iris_store: SharedIrisesRef<ArcIris>) {
+    while let Ok(task) = ch.recv() {
+        match task {
+            IrisTask::DotProductPairs { pairs, rsp } => {
+                let pairs = {
+                    let store = iris_store.data.blocking_read();
```
mcalancea, line 133: Just confirming my understanding: there's no danger of contention here, writes to data only occur when inserting/updating new irises, which happens only for centers towards the end of the pipeline. And not at all during intra-batch and searching.

iris-mpc-common/src/config/mod.rs
```diff
@@ -255,6 +255,9 @@ pub struct Config {
 
     #[serde(default = "default_batch_sync_polling_timeout_secs")]
     pub batch_sync_polling_timeout_secs: u64,
+
+    #[serde(default = "default_tokio_threads")]
```
mcalancea, line 259: So the default here is to allow full overlap between Tokio's threads and the thread pool?

mcalancea (approved):
Approved. Left a couple of questions for my own understanding.

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
naure: Correct. I’d like some type safety in there though. In fact, I’m trying to move the store exclusively in the pool and remove direct accesses from other places. Continuing this in https://github.com/worldcoin/iris-mpc/pull/1594

naure: In the mean time, if there is contention on the lock, that is a bug because that may make the algorithm non-deterministic and unsync the MPC.

iris-mpc-common/src/config/mod.rs
naure: To start yes. There might be phases with a lot of work in tokio: the remaining MPC computation besides dot products, high rate of small network packets, copying data around and book-keeping. To be seen what’s the optimal config and whether we need additional knobs (e.g. limits on the dot product pool).

PR #1579, opened 2025-08-17
[POP-2718] Implement genesis 105 e2e test

bgillesp:
This PR extends the functionality for initializing and updating modifications data in a test Postgres configuration and in the plaintext genesis state struct, and implements the genesis 105 e2e test exercising mixed handling of modifications initialized before a run, and then updated prior to a second run of the genesis protocol.  The PR also fixes some small incorrect edge case behavior in the plaintext implementation of the genesis indexer (delta subprotocol), so that the `last_indexed_modification_id` is properly updated in the presence of non-finalized modifications.

### DIFF
--- a/.github/workflows/test-gpu.yaml
+++ b/.github/workflows/test-gpu.yaml
@@ -10,7 +10,7 @@ concurrency:
 jobs:
   e2e:
     runs-on: arc-gpu-amd64-runner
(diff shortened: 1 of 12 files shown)


siajasl (approved):
Approving but will circle back in review of 1582.

PR #1582, opened 2025-08-18
Refactor and simplify genesis e2e runners

bgillesp:
This PR factors out code common to the different genesis e2e runners, including initialization code and assertion logic.  In addition, the genesis e2e `README.md` should now accurately reflect the assertions being conducted in the code.


iris-mpc-upgrade-hawk/tests/workflows/genesis_100.rs
siajasl, line 1: Genesis args should also include dB snapshot/backup flags as these can be passed into the binary.  Out of scope for this PR.

iris-mpc-upgrade-hawk/tests/workflows/README.md
```diff
@@ -1,112 +1,104 @@
...
-There are zero deletions on S3
+- Databases are initialized
+- S3 deletions: `[]`
 
 Test:
-index genesis up to 100
+- Update the GPU database with simulated modifications:
+    - Id 1: `ResetUpdate` modification for serial id 5, persisted
+    - Id 2: `Reauth` modification for serial id 15, persisted
+    - Id 3: `Uniqueness` modification for serial id 25, persisted
```
carlomazzaferro, line 65: I do not understand what a `Uniqueness` modification here refers to

siajasl (approved):
More refactoring could be done but am approving nonetheless.

iris-mpc-upgrade-hawk/tests/workflows/README.md
bgillesp: Yes, that's correct!  `Uniqueness` modifications may be present in the modifications table during production deployment, and should not impact genesis execution when their associated serial id is outside the range of previously indexed ids -- so this is just confirming that it does nothing in this context.  (Uniqueness modifications should never appear for an already indexed serial id since such a modification appears only when an iris/serial id is first inserted in the database.)

PR #1591, opened 2025-08-20
au/phase-metrics-2: Reduce amount of metrics with "histogram"

naure:
Follow-up to #1567 

Avoid spamming DataDog with low-level metrics. Instead send aggregations (max / avg / median / sum). See [metrics_exporter_statsd docs](https://docs.rs/metrics-exporter-statsd/latest/metrics_exporter_statsd/#histograms).

We only do this for low-level / high-volume metrics. The other `histogram!` calls report raw values.


naure:
Let’s skip this for now unless it becomes a problem.

PR #1593, opened 2025-08-21
[POP-2808] add `build_plaintext.rs`

mcalancea:


bgillesp (approved):
Looks great!  I played around a little bit to gauge how well the multi-threading works to accelerate the performance and found that on my laptop for larger batch sizes it saves about 55% in time to construct a graph over batch size 1, but that most of the performance gain took place when increasing batch size to maybe 4 or 5.  This definitely could be a result of `insert::insert` posing a nontrivial bottleneck as you suggested earlier.  We'll have to keep it in mind for later potential optimizations.

PR #1594, opened 2025-08-21
Au/iris workers store

naure:


iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -665,14 +667,16 @@ impl<'a> InMemoryStore for IrisLoader<'a> {
...
             &mut self.irises,
+            self.iris_pools.clone(),
             [left_code, right_code],
             [left_mask, right_mask]
         ) {
             let iris = GaloisRingSharedIris::try_from_buffers(self.party_id, code, mask)
                 .expect("Wrong code or mask size");
-            side.insert(vector_id, iris);
+            let iris = pool.numa_realloc_blocking(iris).unwrap();
+            store.insert(vector_id, iris);
```
mcalancea, line 679: Does it make sense to bring the store itself to the NUMA node of the pool that is going to use it ?

mcalancea (approved):
Looks good.

iris-mpc-cpu/src/execution/hawk_main.rs
naure: Although not strictly necessary, it does make sense and this is planned.

PR #1607, opened 2025-08-27
(POP 2815) Use search instead of search_to_insert for non-centers

mcalancea:


bgillesp (approved):
Looks good, thanks!  I did just notice that for some testing environments the default HNSW parameters are used, which sets `ef_constr = 320` and `ef_search = 256`, so we should be careful not to mistakenly assume that this provides a free performance gain in these settings.

PR #1620, opened 2025-09-02
au/fast-metrics: Fast metrics with local aggregation (POP-2813)

naure:
[POP-2813 / investigate-correctness-of-metrics-on-datadog](https://linear.app/worldcoin/issue/POP-2813/investigate-correctness-of-metrics-on-datadog)

**Why**

- `FastHistogram` is a metrics collector that pre-aggregates the values.
- It is fast because it is a plain old struct; no locks, no lookups, no background thread, nothing.
- There will be multiple `FastHistogram` active at a time for a same metric. Example:
    - One per thread, using a thread-local variable (`PAIRWISE_DISTANCE_METRICS`).
    - One per HNSW search, using a temporary variable (`metric_edges` in `layer_search_batched`).
    - One per MPC session, stored in a struct (`TcpSession.metric`).
- Aggregated metrics are sent to the `DataDog` agent and cloud, instead of individual values.
- Rates of events per second are measured directly, which should be more reliable and precise than DataDog aggregated rates.

**How to use**

The set of metrics reported by multiple `flush()` of multiple `FastHistogram` for the same name will be further aggregated as distributions in `DataDog`. In particular, it makes sense to combine the same aggregation function of `FastHistogram` and of `DataDog`.

Example:
    - `min:my_metric.min`: Minimum across threads, and minimum within a time period.
    - `sum:my_metric.sum`: Sum of sums.
    - …

Rates:
    - `sum:my_metric.count_rate`: Total rate of records per second (calls to the `record` method).
    - `sum:my_metric.sum_rate`: Total rate of items per second (sum of `value`, e.g. network bytes).
    - `avg:my_metric.sum_rate`: Rate per `FastHistogram` instance, which could mean for instance per thread depending how it is used. Similarly, slowest or fastest with min and max.

**Current metrics setup**

- Iris Workers:
    - Duration of dot product computations. `pairwise_distance.per_pair_duration`
    - Latency including queuing. `iris_worker.latency`
    - Batch size of requests. `pairwise_distance.batch_size`
- Networking: (In #1634)
    - Packet sizes. `outbound_packet_bytes`
    - Number of messages per packet. `outbound_packet_messages`
    - Latency of preparing the packets. `outbound_buffer_latency`
- HNSW, for each layer:
    - Breadth `search_edges_layer`
    - Depth `search_depth`

And for each one, the usual variants are provided: `.count`, `.sum`, `.min`, `.max`, `.avg`.

### DIFF
--- a/.github/workflows/temp-branch-build-and-push.yaml
+++ b/.github/workflows/temp-branch-build-and-push.yaml
@@ -3,7 +3,7 @@ name: Branch - Build and push docker image
 on:
   push:
     branches:
-      - "leonidasnanos-pop-2659-2d-hd-and-implementation-of-the-required-wiring"
+      - "leonidasnanos-pop-2855-implement-2d-stats-cleaning-of-the-cummulative-buckets"
 
 concurrency:
   group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
--- a/iris-mpc-common/src/helpers/statistics.rs
+++ b/iris-mpc-common/src/helpers/statistics.rs
@@ -237,33 +237,89 @@ impl BucketStatistics2D {
         }
     }
 
-    /// Fill bucket counts for the 2D histogram.
-    /// buckets_2d is expected in row-major order (left index major):
-    /// buckets_2d[left_idx * n_buckets_per_side + right_idx]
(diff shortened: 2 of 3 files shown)


mcalancea (approved):
Cool!

PR #1625, opened 2025-09-04
au/borrow-iris: Avoid Arc::clone and Vec in dot product workers (POP-2853)

naure:
Avoid costly `Arc::clone` and `Vec` allocations in Iris workers. We can use iterators and references instead.


mcalancea (approved):
Cool, thank you!

PR #1626, opened 2025-09-04
Mihai/POP-2850: Iris store HashMap to Vec

mcalancea:
Replaces store `Hashmap` with `Vec<Option>`. Additionally, makes getters return references when possible.


iris-mpc-cpu/src/hawkers/shared_irises.rs
```diff
@@ -132,13 +157,13 @@ impl<I: Clone> SharedIrises<I> {
     }
 
     pub fn last_vector_ids(&self, n: usize) -> Vec<VectorId> {
+        // Mihai: was this logic correct previously?
```
naure, line 160: It looks back at the last n insertions. There are odd cases where size<n, or where some vectors have been removed, then we return fewer than n items. Let me know if that’s wrong.

naure: It’s used by a feature called "LUC" if you’d like to poke around and double-check.

iris-mpc-cpu/src/hawkers/shared_irises.rs
```diff
@@ -10,7 +10,8 @@ use crate::execution::hawk_main::state_check::SetHash;
 /// Storage of inserted irises.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 pub struct SharedIrises<I: Clone> {
-    pub points: HashMap<SerialId, (VersionId, I)>,
+    pub points: Vec<Option<(VersionId, I)>>,
```
naure, line 13: Just wondering if that could be not-pub now?

naure (approved):
Great! That will save memory, be faster, and cleaner too.

iris-mpc-cpu/src/hawkers/shared_irises.rs
```diff
@@ -52,18 +56,25 @@ impl<I: Clone> SharedIrises<I> {
...
-            .points
-            .insert(vector_id.serial_id(), (vector_id.version_id(), iris));
+        let serial_id = vector_id.serial_id() as usize;
 
-        self.next_id = self.next_id.max(vector_id.serial_id() + 1);
+        // Extend underlying Vec to accomodate the new serial_id
+        if self.points.len() <= serial_id {
+            self.points
+                .extend((0..serial_id - self.points.len() + 1).map(|_| None));
+        }
```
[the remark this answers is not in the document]
mcalancea: Indeed!

iris-mpc-cpu/src/hawkers/shared_irises.rs
naure: FYI, Local Uniqueness Check, meaning time-local (last `n` here) and geography-local (provided by the upstream client as `luc_ids`).

iris-mpc-cpu/src/hawkers/shared_irises.rs
mcalancea: I made it private and added a getter.

PR #1655, opened 2025-09-12
[POP-2867] Graph diff functionality

mcalancea:
- Implements diffing functionality as traits (for neighborhoods, layers and graphs) for flexibility.
- Existing implementations offer "edge diffing" where the graphs are assumed to be node-equivalent. An `ensure_node_equivalence` method is also provided. 
- Existing impls are based on simple "Jaccard similarity" (size of intersection / size of union) for neighborhoods and different aggregating versions for graphs and layers (single value aggregates + top `n` lowest similarity in each layer).


bgillesp (approved):
Looking great!  I like the new visitor pattern for the differ a lot -- led to very readable implementations for both the Jaccard and the explicit neighbor versions.

PR #1671, opened 2025-09-19
wip: anon stats separately for reauth and uniqueness

carlomazzaferro:


iris-mpc-common/src/config/mod.rs
```diff
@@ -384,6 +388,10 @@ fn default_match_distances_2d_buffer_size() -> usize {
     1 << 13 // 8192
 }
 
+fn default_reauth_match_distances_min_count() -> usize {
```
carlomazzaferro, line 391: Unsure if we need this, was just a consideration on how to limit the amount of reauth bucket stats to send

PR #1676, opened 2025-09-23
[POP-2951] Slow-but-perfect KNN in plaintext

mcalancea:
Usage: 
```
cargo run --release --bin generate_ideal_neighborhoods -- --num-irises 10000 --k 320 --results-file results.txt --num-threads 4
```


iris-mpc-cpu/bin/generate_ideal_neighborhoods.rs
bgillesp, line 1: For serialization, would it be worth encoding results in a binary format to save on the file size of the program output?  In JSON the numbers representing neighbor serial ids are expanded in size by about a factor of 2 relative to using `u32` values directly, so for multi-GB result sets this is quite a substantial savings.  Not sure if there are any gotchas in terms of doing this in a streaming fashion as in the current implementation.

iris-mpc-cpu/bin/generate_ideal_neighborhoods.rs
```diff
@@ -0,0 +1,228 @@
...
+    #[arg(long, default_value = "iris-mpc-cpu/data/store.ndjson")]
+    path_to_iris_codes: PathBuf,
+
+    /// The k for k-NN
+    #[arg(long)]
+    k: usize,
+
+    /// Path to the results file
+    #[arg(long)]
+    results_file: PathBuf,
```
bgillesp, line 55: Modify PR description to include this parameter in usage note since it's required.

iris-mpc-cpu/bin/generate_ideal_neighborhoods.rs
```diff
@@ -0,0 +1,228 @@
...
+    #[arg(long, default_value_t = 1)]
+    num_threads: usize,
+
+    /// Path to the iris codes file
+    #[arg(long, default_value = "iris-mpc-cpu/data/store.ndjson")]
+    path_to_iris_codes: PathBuf,
+
+    /// The k for k-NN
+    #[arg(long)]
+    k: usize,
```
bgillesp, line 51: Modify PR description to include this parameter in usage note since it's required.

iris-mpc-cpu/bin/generate_ideal_neighborhoods.rs
```diff
@@ -0,0 +1,228 @@
...
+    let stream = Deserializer::from_reader(reader).into_iter::<Base64IrisCode>();
+    let mut irises: Vec<IrisCode> = Vec::with_capacity(num_irises);
+
+    let (limit, skip, step) = match args.irises_selection {
+        IrisSelection::All => (num_irises, 0, 1),
+        IrisSelection::Even => (num_irises * 2, 0, 2),
+        IrisSelection::Odd => (num_irises * 2, 1, 2),
+    };
+
+    let stream_iterator = limited_iterator(stream, Some(limit))
```
bgillesp, line 187: Consider using `stream.take(limit)` rather than `limited_iterator` -- I wrote this function a while ago to make it easier to use `Option<usize>` to optionally limit the iterator in question, but in this setting it isn't needed.

bgillesp (commented):
Really great binary!  Extremely usable and resilient.  I had just one larger comment about the serialization format, but otherwise it's looking good. :+1:

iris-mpc-cpu/bin/generate_ideal_neighborhoods.rs
mcalancea: I'm inclined against it because it complicates recovering the state of the file if the program is interrupted. I see that the produced file for `n=10^4` and `k=320` is `16MB`. What changes for `n=10^6` is a factor of `100` for `n` and a factor of `1.5` for the average number of digits per neighbor, so it would put it at about `2.4GB`. 

For this size I think it's reasonable to execute the computation in the current format and shrink the file after it's complete, if we find its size unwieldy.

bgillesp: Sounds reasonable -- I may be using it for somewhat larger runs producing a file in the 25-30GB range, but reformatting after execution should be acceptable if I need to shuttle the resulting file around.

bgillesp (approved):
Looking great, thanks for the revisions! :pray:

PR #1687, opened 2025-09-25
[POP-2942] Test Utils: Delete original code 1

siajasl:
In [POP-2869](https://linear.app/worldcoin/issue/POP-2869/test-utils-create-subcrate) a new sub-crate called iris-mpc-utils was initialized from code scattered across the solution … but with the original code left in place.  

This is the first step towards removing original code.   Some binaries are deleted as well as test-utils embedded in production code.


siajasl:
After discussions it was decided not to proceed due to circularity issues.

PR #1696, opened 2025-09-29
Ability to disable anon stats per-batch, from SQS messages

carlomazzaferro:
### Why
Reduce batch latency under high load by skipping anonymized statistics (1D, 2D, mirror) collection and aggregation when requested by backend.

### What
Introduce a per-batch flag that disables anon stats end-to-end (SQS → batch → GPU kernels).
- New request flag: disable_anonymized_stats in Uniqueness SQS message JSON.
- Batch-level control: If any Uniqueness request in a batch sets the flag, the entire batch disables anon stats.
- GPU changes: CUDA openResults paths skip writing match-distance buffers and counters; 2D stats computation is fully bypassed.

~Note: tests aren't perfect, as they test the case in which every batch is processed with the flag being set. In real world, this is unlikely to be the case - there will be batches with, and batches without such flag~

Edit: added the test cases with probabilistic setting of disablement of anon stats for better real-world testing

### DIFF
--- a/iris-mpc-gpu/src/server/actor.rs
+++ b/iris-mpc-gpu/src/server/actor.rs
@@ -2150,9 +2150,7 @@ impl ServerActor {
         orientation: Orientation,
     ) -> (PartialResultsWithRotations, Vec<OneSidedDistanceCache>) {
         // we try to calculate the bucket stats here if we have collected enough of them
-        if !self.disable_anonymized_stats_for_current_batch {
-            self.try_calculate_bucket_stats(eye_db, orientation);
-        }
+        self.try_calculate_bucket_stats(eye_db, orientation);
 
         let old_distance_cache_counters = match orientation {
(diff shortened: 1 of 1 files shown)


iris-mpc-gpu/src/server/actor.rs
```diff
@@ -2139,7 +2150,9 @@ impl ServerActor {
         orientation: Orientation,
     ) -> (PartialResultsWithRotations, Vec<OneSidedDistanceCache>) {
         // we try to calculate the bucket stats here if we have collected enough of them
-        self.try_calculate_bucket_stats(eye_db, orientation);
+        if !self.disable_anonymized_stats_for_current_batch {
```
dkales, line 2153: This if seems redundant with the internal one of try_calculate_bucket_stats?

dkales (dismissed):

PR #1707, opened 2025-10-02
[POP-2956] rotation aware trick dot

sdwoodbury:
adds two new versions of `trick_dot()`:
- `rotation_aware_trick_dot()`: associates `self` with a rotation, and performs the dot product as if `self` has previously been pre-processed by `rotate_coefs_left()` or `rotate_coefs_right()`.
- `rotation_aware_trick_dot_padded()`: adds padding to each row so that, for example, a rotation of -15 (or `Left(15)` is achieved by selecting index 0..800 from the padded row.

a unit test was used to verify correctness of `rotation_aware_trick_dot()`

a benchmark was added to compare the 3 versions of `trick_dot()` in two scenarios: CPU bound and RAM bound

`trick_dot()` was refactored to be shared by `GaloisRingIrisCodeShare` and `GaloisRingTrimmedMaskCodeShare`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2584,11 +2584,9 @@ dependencies = [
  "itertools 0.13.0",
  "metrics 0.22.3",
  "metrics-logger",
- "num-format",
  "num-traits",
  "num_cpus",
  "num_enum",
- "perf-event",
  "prost 0.13.3",
  "rand",
  "rand_chacha",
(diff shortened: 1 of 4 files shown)


sdwoodbury:
while working on [POP-2962](https://linear.app/worldcoin/issue/POP-2962/batch-rotation-aware-dot-products) i realized that `GaloisRingTrimmedMaskCodeShare` doesn't have the rotation aware dot product, and also that this code can be refactored for better code reuse as follows:

```
fn trick_dot<const D: usize>(left: &[u16; D], right: &[u16; D]) -> u16 {
       ...
}

// same for rotation_aware_trick_dot
```

and then `GaloisRingIrisCodeShare` and `GaloisRingTrimmedMaskCodeShare` would call the same generic function. I have this in my branch for `POP-2962` but can move it to this PR if desired....

naure (approved):
Great work!

PR #1711, opened 2025-10-06
[POP-2992] Implement construct-graph-ptxt binary

bgillesp:
PR implements a small binary to construct an HNSW graph from plaintext iris code input, with output serialized to file using the standard "single-graph" binary output format.  This binary will be used primarily for upcoming data analysis tasks.

### DIFF
--- a/iris-mpc-cpu/bin/construct_graph_ptxt.rs
+++ b/iris-mpc-cpu/bin/construct_graph_ptxt.rs
@@ -1,0 +1,120 @@
+use std::{
+    fs::File,
+    io::{BufReader, BufWriter},
+    path::PathBuf,
+    sync::Arc,
+};
+
+use clap::Parser;
+use eyre::Result;
+use tracing::info;
+
+use iris_mpc_cpu::{
+    hawkers::plaintext_store::PlaintextStore,
+    hnsw::{vector_store::VectorStoreMut, GraphMem, HnswParams, HnswSearcher},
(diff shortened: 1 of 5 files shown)


iris-mpc-cpu/bin/construct_graph_ptxt.rs
```diff
@@ -0,0 +1,120 @@
...
+    info!("Parsing CLI arguments");
+    let args = Args::parse();
+
+    let mut params = HnswParams::new(args.ef, args.ef, args.M);
+    if let Some(q) = args.layer_probability {
+        params.layer_probability = q;
+    }
+
+    info!("Opening iris codes input stream");
+    let file = File::open(args.iris_codes_path.as_path()).unwrap();
```
iliailia, line 75: You could repurpose the `read_json` function from `py_bindings/io.rs`

iris-mpc-cpu/bin/construct_graph_ptxt.rs
```diff
@@ -0,0 +1,120 @@
...
+            )
+            .await?;
+
+        counter += 1;
+        if counter % 1000 == 0 {
+            info!("Inserted {} plaintext entries", counter);
+        }
+    }
+
+    info!("Persisting HNSW graph to file");
```
iliailia, line 114: You can repurpose `write_bin` from `py_bindings/io.rs`.

iliailia (approved):
Thanks! Good idea to use the parallel insert algorithm.

PR #1713, opened 2025-10-06
[POP-2916] Change `intra_batch_rule` to `AND`

mcalancea:
See [linear issue](https://linear.app/worldcoin/issue/POP-2916/investigate-mismatch-between-hnsw-and-gpu) for context.

See [this thread](https://inversed.slack.com/archives/C08FBFZSTHP/p1759509084751299?thread_ts=1758205823.283759&cid=C08FBFZSTHP) for discussion on `AND` being implemented by the GPU and the fact that this change does resolve the failing test case.


bgillesp (approved):
Looks good to go -- thanks for your work in identifying the issue here!

PR #1716, opened 2025-10-07
[POP-2965] Compute minimal distance over rotations 

iliailia:
This adds a new method to `VectorStore` that takes rotations of a query iris, computes the distances between them and a given iris vector and returns the minimal distance lifted to 32-bit shares for further comparisons. Batching is supported. `Aby3Store` contains an optimized implementation via oblivious minimal trees.


naure:
In this PR, the search algorithm has to provide the rotations: `query: &[Self::QueryRef; ROTATIONS]`.

However, my understanding of this was to move the concept of rotations into the iris-worker. This way, the search algorithm does not have to be aware of rotations, it only manipulates the one central `query: Self::QueryRef`. See [POP-2962/batch-rotation-aware-dot-products-in-iris-worker](https://linear.app/worldcoin/issue/POP-2962/batch-rotation-aware-dot-products-in-iris-worker).

This has two advantages:
1.  The search code does not become more complex than it already is.
2. The iris worker is in a better place to be tuned and optimized when it is responsible for all rotations at once.

PR #1720, opened 2025-10-08
Mihai/pop 2952

mcalancea:


bgillesp (changes requested):
Looking great, thank you!  See below for one quick suggestion that would be helpful to control memory usage. :pray:

bgillesp (approved):

PR #2007, opened 2026-02-27
feat: continuous rerand

philsippl:
Spec and implementation of continuous rerandomization of the iris share databases.

Spec is provided in [docs/specs/rerandomization.md](https://github.com/worldcoin/iris-mpc/blob/ps/cont-rerand/docs/specs/rerandomization.md).

**Goals:**

- Continuous rerandomization of shares in persistent storage (in-memory representation stays untouched)
- Rerand performed by separate server than matching server (for network performance reasons)
-  Minimal impact on normal server (e.g. matching, startup, ...)
- Support for GPU and HNSW versions 

**High level design:**
- Rerand server operating in chunks over DB
- The rerand servers can be at most 1 chunk (or 1 epoch at the boundary) apart from each other
- Synchronization with matching server required:
  - at startup: while the matching server is loading the db, we freeze the rerand
  - when applying modifications: iris share table is always locked, when there is a writer 
- Rerand servers first write to "staging" schema and then apply in chunks to reduce lock times
- Relies on the current system property that modifications are guaranteed to eventually arrive on all parties
  - This PR also makes modification handling a bit more robust by deleting only after storing a modification and recovering potentially lost modifications.
   
   related: https://github.com/worldcoin/ampc-common/pull/71


iris-mpc-upgrade/src/s3_coordination.rs
```diff
@@ -0,0 +1,416 @@
...
+    party: u8,
+    chunk_id: u32,
+    version_map: &[(i64, i16)],
+) -> Result<()> {
+    let prefix = format!("{}/chunk-{}", epoch_party_prefix(epoch, party), chunk_id);
+
+    let hash = version_map_hash(version_map);
+    upload_marker(s3, bucket, &format!("{prefix}/version-hash"), hash.to_vec()).await?;
+
+    let body = serde_json::to_vec(version_map)?;
```
dkales: The collection of all documents here will get quite large, no? I guess the auto-cleanup of S3 will handle this? Otherwise one could only upload this on hash mismatch?

dkales (commented):
A few minor comments, overall strategy seems good. Have not looked at the e2e tests yet.

iris-mpc-upgrade/tests/continuous_rerand_e2e.rs
```diff
@@ -0,0 +1,682 @@
...
+            let r = simulate_server_startup(&env.harness, p).await;
+            println!("[phase 4]   party {} server startup: {:?}", p, r.is_ok());
+        }
+
+        wait_epoch_done(&env.harness, 0).await?;
+        stop_all(t, h).await;
+
+        let ep = assert_consistent_rerand_epoch(&env.harness, &[]).await?;
+        assert!(ep >= 1);
+        verify_fingerprints(&env.harness, &env.fingerprints, &[]).await?;
```
dkales: Maybe in general this should also assert that the shares are actually different, not only that the fingerprint of the PR is fine.

iris-mpc/src/server/mod.rs
```diff
@@ -138,33 +172,75 @@ pub async fn server_main(config: Config) -> Result<()> {
...
+            .filter(|(i, _)| *i != config.party_id)
+            .map(|(_, (h, p))| -> eyre::Result<_> { Ok((h.as_str(), p.parse::<usize>()?)) })
+            .collect::<eyre::Result<Vec<_>>>()?;
+        rerand_store::freeze_and_verify_watermarks(&iris_store.pool, &peer_addrs).await?;
     }
+    // Worker is now frozen with verified equal watermarks.
+    // Everything from here until freeze release must be wrapped so that
+    // errors always release the freeze.
+    let frozen_result = async {
+        let rerand_lock_conn = rerand_store::acquire_apply_lock(&iris_store.pool).await?;
```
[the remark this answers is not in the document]
philsippl: Yeah it's just there for extra safety, leaving a comment.

philsippl:
[paused] do not merge for now

PR #2053, opened 2026-03-25
[POP-3625] IrisWorkerPool trait: decouple distance computation from local worker pool

mcalancea:
## Summary

Introduces the `IrisWorkerPool` trait that abstracts iris distance computation behind opaque `QueryId` handles, decoupling callers from the concrete `IrisPoolHandle`. This is Phase 1 of the "Iris Db Sharding and Memory Optimization" design — the trait surface is ready for a future remote worker implementation.

- **`Aby3Query` becomes a lightweight handle** — just `(QueryId, rotation, mirrored)`, no iris data
- **`LocalIrisWorkerPool`** wraps `IrisPoolHandle` + query cache; `cache_queries` does the full pipeline: preprocess → rotate → mirror → NUMA-realloc
- **`Aby3Store<D, W>`** is generic over `W: IrisWorkerPool`
- **Cache/evict lifecycle** in `handle_job`: cache at start, evict at end (including identity update QIDs)
- **Genesis** fixed: fetches irises from store, caches before search, evicts after
- **`insert_irises` returns store checksum** (`u64`) — the commutative `SetHash` over all stored VectorIds, available for immediate consistency verification

## Reading order (commit by commit)

1. **Trait + impl + helpers** — `IrisWorkerPool` trait, `LocalIrisWorkerPool`, `QueryId`/`QuerySpec`, leaf helpers (`mirrored()`, `allocate_next_id`)
2. **Store switchover** — `Aby3Query` → type alias, `Aby3Store<D,W>` generic, distance functions rewired through trait. `DistanceSimple`/`DistanceMinimalRotation` structs collapsed into `DistanceFn` methods — with `W::compute_dot_products` taking a `DistanceMode`, the two reduced to identical pipelines (dot products → lift → optional `oblivious_min`), so the shared structure is now visible inline.
3. **Hawk main integration** — `HawkActor` worker_pools, `HawkRequest` with raw irises + QueryIds, `handle_job` cache/evict lifecycle, identity update tracking, intra-batch/search adapted
4. **Genesis integration** — `Batch` carries `irises_to_cache`, caching before search, eviction after
5. **Benchmarks + binaries** — mechanical adaptation

## Key design decisions

- `std::sync::RwLock` for query cache (never held across `.await`, read-dominant during search)
- Both worker pools cache ALL irises (mirror queries cross eyes)
- `compute_pairwise_distances` second operand is `QueryId` not `QuerySpec` (raw iris has no rotation)
- Cache miss returns `Result` instead of panicking
- `VectorStoreMut::insert` uses `allocate_next_id` under write lock (TOCTOU fix)
- `insert_irises` returns `u64` store checksum (design doc specifies `StoreChecksum`)

## Known remaining work

- `cache_queries` generates all 63 variants per iris; a rotation/mirror mask parameter would reduce this to 7 (hawk main) or 1 (genesis) — see TODO in trait definition
- `BatchQuery` rotated/interpolated fields are now dead weight in CPU path (still used by GPU)

## Test plan

- [x] 112 unit tests pass
- [x] Clippy clean, fmt clean
- [ ] Docker integration test (e2e)
- [ ] Dev cluster deploy


iris-mpc-common/src/galois_engine.rs
```diff
@@ -205,6 +205,18 @@ pub mod degree4 {
...
+        /// `remap_new_to_mirrored_index` preserves the `b` component and the
+        /// trimmed range corresponds to `b == 0`.
+        pub fn mirrored(&self) -> Self {
+            // mirrored_mask uses flip_imaginary = false, so no imaginary flip needed
+            let mut res = self.clone();
+            for i in 0..MASK_CODE_LENGTH {
+                res.coefs[GaloisRingIrisCodeShare::remap_new_to_mirrored_index(i)] = self.coefs[i];
+            }
+            res
+        }
```
bgillesp: Can we make the usage of the index mirroring here more explicit, as below?  Your existing documentation does explain the nuance, but I fear that a subtle bug could slip through in the future if the non-standard usage isn't isolated.

```suggestion
        /// Remap new indices to mirrored variant for trimmed shares.
        /// 
        /// NOTE: Remapping of indices to mirrored indices works unmodified for
        /// trimmed shares because full resolution `remap_new_to_mirrored_index`
        /// preserves the `b` component and the trimmed range corresponds to `b == 0`.
        fn remap_new_to_mirrored_index_trimmed(i: usize) -> usize {
            GaloisRingIrisCodeShare::remap_new_to_mirrored_index(i)
        }

        /// Mirror the mask share.
        pub fn mirrored(&self) -> Self {
            // mirrored_mask uses flip_imaginary = false, so no imaginary flip needed
            let mut res = self.clone();
            for i in 0..MASK_CODE_LENGTH {
                res.coefs[Self::remap_new_to_mirrored_index_trimmed(i)] = self.coefs[i];
            }
            res
        }
```

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -477,6 +480,501 @@ fn worker_thread(ch: Receiver<IrisTask>, iris_store: SharedIrisesRef<ArcIris>, n
...
+    ///
+    /// The worker performs the full preprocessing pipeline on each iris:
+    /// mirror, preprocess (Lagrange interpolation), and generate all 31
+    /// rotations for both normal and mirrored variants.
+    ///
+    /// Caching an already-cached `QueryId` is a no-op.
+    ///
+    // TODO: Accept a rotation/mirror mask so callers can request only the
+    // variants they need. Currently every call generates all 31 rotations ×
+    // 2 orientations (63 variants), but:
```
bgillesp: 62 variants

iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
```diff
@@ -171,23 +166,6 @@ where
...
-    pub async fn eval_pairwise_distances(
-        &mut self,
-        pairs: Vec<Option<(ArcIris, ArcIris)>>,
-    ) -> Result<Vec<DistanceShare<D::Ring>>> {
-        if pairs.is_empty() {
-            return Ok(vec![]);
-        }
-
-        self.distance_fn.eval_pairwise_distances(self, pairs).await
-    }
```
bgillesp, line 189: I would prefer to revert the concrete implementation of `eval_pairwise_distances` back to sitting in `distance_fn.rs` if that's possible, using a light wrapper here to expose the functionality externally.  The distance fn module focuses on all functionality that depends on the differences between `Simple` and `MinRotation`, and the new implementation of `eval_pairwise_distances` is the only place in `aby3_store.rs` that the two variants are exposed directly.

bgillesp: Suggested modification is now in PR #2084.

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -477,6 +480,524 @@ fn worker_thread(ch: Receiver<IrisTask>, iris_store: SharedIrisesRef<ArcIris>, n
...
+    fn cache_queries(
+        &self,
+        queries: Vec<(QueryId, ArcIris)>,
+    ) -> impl Future<Output = Result<()>> + Send;
+
+    /// Compute dot products for batches of (query_spec, targets).
+    ///
+    /// Each `QuerySpec` selects a specific preprocessed rotation from the
+    /// cache. Returns one `Vec<RingElement<u16>>` per batch.
+    fn compute_dot_products(
```
bgillesp, line 602: Would prefer for DistanceMode to be incorporated as a configuration field for individual implementations rather than a dynamic argument in the trait function signatures.  In general we only use a single DistanceMode at once (otherwise we'd need to be juggling multiple HNSW graphs), so encoding as a fixed configuration field better represents the contract that only one distance mode choice is valid in a particular execution environment.

Proposed fix is in PR #2085.

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -477,6 +480,524 @@ fn worker_thread(ch: Receiver<IrisTask>, iris_store: SharedIrisesRef<ArcIris>, n
...
+    ) -> impl Future<Output = Result<Vec<RingElement<u16>>>> + Send;
+
+    /// Evict cached queries, freeing memory.
+    fn evict_queries(&self, query_ids: Vec<QueryId>) -> impl Future<Output = Result<()>> + Send;
+
+    /// Delete irises by replacing them with party-specific dummy sentinels
+    /// that produce max-distance in dot products.
+    fn delete_irises(
+        &self,
+        party_id: usize,
```
bgillesp: `party_id` seems like it should be a config field in the implementer rather than a dynamic argument in the trait definition.  It doesn't make sense for a specific worker pool to ever be able to specify different values over the course of a run.

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -477,6 +480,524 @@ fn worker_thread(ch: Receiver<IrisTask>, iris_store: SharedIrisesRef<ArcIris>, n
...
+            let pairs: Vec<_> = irises
+                .into_iter()
+                .map(|iris| (QueryId::new(), iris))
+                .collect();
+            let specs: Vec<_> = pairs.iter().map(|(qid, _)| QuerySpec::new(*qid)).collect();
+            this.cache_queries(pairs).await?;
+            Ok(specs)
+        }
+    }
+}
```
bgillesp: Can these be migrated to separate functions in `test_utils.rs`?  This is not core functionality of the trait (basically only used for testing), so it would be nicer if we could keep the trait definition a bit leaner.

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -477,6 +480,524 @@ fn worker_thread(ch: Receiver<IrisTask>, iris_store: SharedIrisesRef<ArcIris>, n
...
+    /// (e.g., `numa_realloc`, `reserve`, `wait_completion`).
+    pub fn inner(&self) -> &IrisPoolHandle {
+        &self.inner
+    }
+}
+
+/// Build 31 `ArcIris` rotations from code and mask rotation vecs.
+fn zip_rotations(
+    code_rots: Vec<iris_mpc_common::galois_engine::degree4::GaloisRingIrisCodeShare>,
+    mask_rots: Vec<iris_mpc_common::galois_engine::degree4::GaloisRingTrimmedMaskCodeShare>,
```
bgillesp: Import rather than qualified references for better readability.

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -477,6 +480,524 @@ fn worker_thread(ch: Receiver<IrisTask>, iris_store: SharedIrisesRef<ArcIris>, n
...
+) -> Vec<ArcIris> {
+    code_rots
+        .into_iter()
+        .zip(mask_rots)
+        .map(|(code, mask)| {
+            Arc::new(crate::protocol::shared_iris::GaloisRingSharedIris { code, mask })
```
bgillesp: Import here as well rather than qualified.

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -477,6 +480,524 @@ fn worker_thread(ch: Receiver<IrisTask>, iris_store: SharedIrisesRef<ArcIris>, n
...
+                    .filter(|(qid, _)| !cache.contains_key(qid))
+                    .collect()
+            };
+            if new_queries.is_empty() {
+                return Ok(());
+            }
+
+            // Preprocess + rotate, collecting all resulting ArcIris values.
+            let mut entries: Vec<(QueryId, CachedQuery)> = Vec::with_capacity(new_queries.len());
+            for (query_id, iris) in &new_queries {
```
bgillesp: Nit: slightly simpler to iterate over `new_queries` rather than `&new_queries`, then remove deref and clone in final `entries.push(...)` call.

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -315,11 +315,18 @@ pub struct HawkActor {
     /// A size used by the start-up loader.
     loader_db_size: usize,
     /// In-memory storage for the secret-shared iris codes for both left and right eyes.
+    /// Used by workers for distance computation and by external mutation paths (deletions).
     iris_store: BothEyes<Aby3SharedIrisesRef>,
```
bgillesp, line 319: The existence of this data structure in the hawk actor violates the contract that the worker pool has sole ownership of iris code data.  This won't be compatible with workers running on separate nodes, since the primary iris data won't be loaded by the main server at all.

iris-mpc-cpu/src/execution/hawk_main/intra_batch.rs
```diff
@@ -70,23 +73,24 @@ async fn per_session(
...
-            let iris1_proc =
-                &search_queries[batch.i_eye][pair.task.i_request][pair.task.i_rotation].iris_proc;
-            let iris2 = &search_queries[batch.i_eye][pair.earlier_request]
+            let spec_a = search_queries[batch.i_eye][pair.task.i_request][pair.task.i_rotation];
+            let id_b = raw_queries[batch.i_eye][pair.earlier_request]
                 .center()
-                .iris;
-            Some((iris1_proc.clone(), iris2.clone()))
+                .query_id;
+            Some((spec_a, id_b))
```
bgillesp, line 89: Are these functionalities equivalent?  This looks to me like the old version may actually have been logically incorrect: `iris1_proc` and `iris2` both are derived from the same side of the query from either Normal or Mirror orientations, so I think this means that previously intra_batch compared normal to normal irises and mirrored to mirrored irises, the latter of which is equivalent to comparing normal to normal irises.  I think the proper behavior would be to compare mirrored to normal irises as well, which is I believe what the new implementation does.  Am I understanding this correctly?

iris-mpc-cpu/src/execution/hawk_main/intra_batch.rs
```diff
@@ -54,13 +57,13 @@ pub async fn intra_batch_is_match(
 #[instrument(level = "trace", target = "searcher::network", skip_all)]
 async fn per_session(
     search_queries: &BothEyes<VecRequests<VecRotations<Aby3Query>>>,
+    raw_queries: &BothEyes<VecRequests<VecRotations<Aby3Query>>>,
```
bgillesp: Consider renaming arg: `unmirrored_queries` (similarly above)

iris-mpc-cpu/src/execution/hawk_main/intra_batch.rs
bgillesp: (Perhaps the test on line 195 below confirms this observation.)

iris-mpc-cpu/src/hawkers/aby3/test_utils.rs
```diff
@@ -327,16 +328,16 @@ pub async fn shared_random_setup<R: RngCore + Clone + CryptoRng>(
...
-            .map(|id| Aby3Query::new_from_raw(shared_irises[id][role].clone()))
-            .collect::<Vec<_>>();
+        let irises: Vec<ArcIris> = (0..database_size)
+            .map(|id| Arc::new(shared_irises[id][role].clone()))
+            .collect();
         let store = store.clone();
         let task: JoinHandle<Result<(Aby3StoreRef, GraphMem<Aby3VectorRef>)>> =
             tokio::spawn(async move {
                 let mut store_lock = store.lock().await;
+                let queries = store_lock.workers.cache_irises(irises).await?;
```
bgillesp: Do we need to later evict this queries from the cache?

iris-mpc-cpu/src/hawkers/shared_irises.rs
```diff
@@ -37,16 +37,18 @@ impl<I: Clone> SharedIrises<I> {
         let next_id = points_map.keys().map(|v| v.serial_id()).max().unwrap_or(0) + 1;
 
         let mut points = vec![None; next_id as usize];
+        let mut set_hash = SetHash::default();
         for (v, iris) in points_map.into_iter() {
             points[v.serial_id() as usize] = Some((v.version_id(), iris));
+            set_hash.add_unordered(v);
```
bgillesp, line 43: Good catch.

iris-mpc-cpu/src/execution/hawk_main/test_utils.rs
```diff
@@ -111,20 +122,98 @@ pub fn make_request_intra_match(batch_size: usize, party_id: usize) -> HawkReque
...
+            (
+                GaloisRingSharedIris::generate_shares_locally(iris_rng, iris.clone())[party_id]
+                    .clone(),
+                GaloisRingSharedIris::generate_mirrored_shares_locally(iris_rng, iris.clone())
+                    [party_id]
+                    .clone(),
+            )
+        })
+        .collect();
+
```
bgillesp, line 175: Might be more correct to modify `left_shares` and `right_shares` so last entries are mirror/side swap of first entries before feeding them to `receive_batch_shares`.  (Overwriting individual fields in the resulting `BatchQuery` might not overwrite everything that depends on the input shares, which could produce an inconsistent batch state).

bgillesp (changes requested):
Quite a significant block of work, nice going!  Recommended process:
- Review PRs #2084 and #2085, and merge 2085 to 2084 to 2053 if the changes look reasonable.
- Look through comments on this PR and make updates/respond
- Only merge main back into this branch after we've finished with all discussion/updates on the current changeset, to minimize confounding diffs

iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
mcalancea: Merged, thank you for preparing it!

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
mcalancea: Merged, thank you for preparing it!

I think we'll need more work here to tighten the coupling (make sure store uses mode of the worker impl, which in turn is well coordinated with the worker binary in the remote case)

iris-mpc-cpu/src/execution/hawk_main/intra_batch.rs
mcalancea: Yes, it was. We should make sure TFH use the fixed a version if they rerun their integration tests.

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
mcalancea: State of the art AI arithmetic I'm afraid

iris-mpc-cpu/src/hawkers/aby3/test_utils.rs
mcalancea: Yes. I know there's been discussion for a while in the Rust community about introducing linear types (values must be consumed exactly once) which would be a nice way of addressing this (cache method returns a guard object which must be consumed with evict).

I'll think about other ways of doing this more reliably, but won't block this PR with it.

iris-mpc-cpu/src/execution/hawk_main.rs
bgillesp: From external conversation for context: this issue will be fixed in follow-up work when we implement the network/sharded iris worker pool interface.

bgillesp (approved):
Looking great to me, thanks for the revisions!

PR #2076, opened 2026-04-10
fix s3 checks

sdwoodbury:
- continue on failure to delete s3 checkpoint. 
- add logging of the execution args, config, and s3 region configuration.
- update the `temp` workflow that builds genesis for testing on the dev cluster, to use `Dockerfile.genesis.dev.hawk`, which uses a script as an entrypoint, allowing testing genesis with different max_height, etc, without building a new container.

### DIFF
--- a/iris-mpc-upgrade-hawk/src/genesis/mod.rs
+++ b/iris-mpc-upgrade-hawk/src/genesis/mod.rs
@@ -1150,7 +1150,7 @@ async fn verify_s3_checkpoint_access(
         .send()
         .await
     {
-        tracing::warn!("S3 checkpoint bucket delete check failed: {e}");
+        log_warn(format!("S3 checkpoint bucket delete check failed: {e}"));
     }
 
     Ok(())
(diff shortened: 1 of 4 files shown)


iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -1138,13 +1143,15 @@ async fn verify_s3_checkpoint_access(
...
-    s3_client
+    if let Err(e) = s3_client
         .delete_object()
         .bucket(bucket)
         .key(&key)
         .send()
         .await
-        .map_err(|e| eyre!("S3 checkpoint bucket delete check failed: {e}"))?;
+    {
+        tracing::warn!("S3 checkpoint bucket delete check failed: {e}");
```
bgillesp: Should use `log_warn` function for consistency with other outputs in the file.

bgillesp (approved):
Looks good!  I can't approve the change to the GitHub workflow myself, since Philipp is the code owner there.

wojciechsromek (approved):

PR #2077, opened 2026-04-14
[POP-3724] implement genesis graph sync procotol

sdwoodbury:


iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -1717,3 +1836,209 @@ async fn upload_and_sync_genesis_checkpoint(
...
+        .filter_map(Result::ok)
+        .collect();
+    if all_hashes.len() != checkpoints.len() {
+        log_warn("some checkpoint hashes failed to parse".into());
+    }
+    for (src, dest) in all_hashes
+        .into_iter()
+        .take(hashes.len())
+        .zip(hashes.iter_mut())
+    {
```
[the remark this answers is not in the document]
sdwoodbury: hashes[idx] need not correspond to checkpoints[idx]

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -1679,6 +1740,64 @@ async fn validate_consistency_of_stores(
...
+    common_set.retain(|_k, count| *count == others_hashes.len() + 1);
+    for (idx, hash) in my_checkpoint_hashes.iter().enumerate() {
+        if common_set.contains_key(hash) {
+            let r = my_checkpoints.get(idx).cloned();
+            if r.is_none() {
+                bail!("unreachable condition in get_common_checkpoint()");
+            }
+            return Ok(r);
+        }
+    }
```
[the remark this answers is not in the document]
sdwoodbury: that is not the intended behavior. though it would be useful.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -346,20 +335,29 @@ async fn exec_setup(
...
     let server_coord_config = &config
         .server_coordination
         .clone()
         .unwrap_or_else(|| panic!("Server coordination config is required for server operation"));
+
     // Coordinator: await server start.
-    let (is_ready_flag, verified_peers, my_uuid) = start_coordination_server(
+    let extra_routes = Router::new().route(
+        GRAPH_CHECKPOINT_ROUTE,
+        get(move || async move { serde_json::to_string(&hashes).unwrap() }),
```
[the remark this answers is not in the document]
sdwoodbury: fair but serializing a array of u8 should never fail

iris-mpc-cpu/src/genesis/genesis_checkpoint.rs
```diff
@@ -186,22 +190,45 @@ pub async fn save_checkpoint_state<V: VectorStore>(
     Ok(())
 }
 
-pub async fn cleanup_old_checkpoints<V: VectorStore>(
+/// DANGER: requires that the 3 parties have
+/// reached consensus on a common checkpoint.
+/// if the parties have not reached agreement, then
+/// calling this function could make rollback impossible
+pub async fn cleanup_checkpoints<V: VectorStore>(
```
mcalancea, line 197: My take on this has fluctuated, but at the moment I'm cozying up to the safety of not deleting backups at all during genesis. My estimate for the cost of storing _everything_ until we finish is < 200$ a month (with it being pro-rata, it will actually be less). Or at least we could mark the last backup of a run as special and never delete the special ones, which would be insignificant in terms of cost but would offer more safety. 

@bgillesp thoughts?

iris-mpc-store/src/lib.rs
```diff
@@ -448,15 +448,22 @@ WHERE id = $1;
...
 
-    pub async fn rollback(&self, db_len: usize) -> Result<()> {
+    pub async fn rollback(&self, last_indexed_id: usize) -> Result<()> {
         let mut tx = self.pool.begin().await?;
+        self.rollback_tx(&mut tx, last_indexed_id).await?;
+        tx.commit().await?;
+        Ok(())
+    }
 
+    pub async fn rollback_tx(
```
mcalancea: The naming of this can be improved, right now it suggests we might be rolling back a transaction. `rollback_in_tx` or similar? Or maybe `delete_irises_after_id` instead of `rollback` in both places.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -1751,3 +1876,212 @@ async fn upload_and_sync_genesis_checkpoint(
...
+        .collect();
+    if checkpoints.len() != db_checkpoints_len {
+        log_warn("failed to convert some GenesisCheckpointRow to GenesisCheckpointState".into());
+    }
+    let hashes: Vec<_> = checkpoints
+        .iter()
+        .map(|x| blake3::Hash::from_hex(x.blake3_hash.as_bytes()))
+        .filter_map(Result::ok)
+        .collect();
+    if hashes.len() != checkpoints.len() {
```
mcalancea: This situation will cause a bug, because the checkpoints and the hashes will be misaligned and the wrong checkpoint will be returned for the common hash. It would be nice to prevent this at the type-level even by returning a vec/array of tuples instead of two separate containers.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -1713,6 +1777,67 @@ async fn validate_consistency_of_stores(
...
+        bail!("invalid number of parties");
+    }
+    find_common_checkpoint(my_checkpoint_hashes, my_checkpoints, others_hashes)
+}
+
+/// Pure subset logic for [`get_common_checkpoint`].
+/// Finds the first checkpoint whose hash appears in all parties'
+/// hash lists (mine + the X `others_hashes`).  Zero hashes are ignored
+/// because they represent empty slots in a [`GraphCheckpointHashes`] array.
+fn find_common_checkpoint(
```
mcalancea: Given the tiny scale we're operating at here (3 parties, 10 hashes per party), I'd go with just a couple nested loops checking that all other parties contain the current candidate. Simpler to read.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -392,6 +387,46 @@ async fn exec_setup(
         bail!("Shutdown")
     }
 
+    let graph_checkpoint = get_common_checkpoint(config, hashes, graph_checkpoints).await?;
```
mcalancea, line 394: I suggest we add the possibility for a manual override on the actual checkpoint that we want. It covers some cases where the snapshot selected by the current algorithm fails for some reasons but a different suitable snapshot exists; these cases are pretty obscure and shouldn't happen in normal operation as far as I can tell, but it would be nice to handle them at the CLI level if we ever need to.

iris-mpc-cpu/src/genesis/genesis_checkpoint.rs
mcalancea: Although it might make sense to keep deleting from the checkpoint db itself to ensure consistency.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -636,6 +671,7 @@ async fn exec_delta(
                 )
                 .await?;
                 log_info(String::from("S3 checkpoint created after delta"));
+                hawk_handle.sync_peers(false, None).await?;
```
mcalancea: The upload method itself calls sync_peers at the end. Does this achieve something different?

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -841,6 +877,7 @@ async fn exec_indexation(
                     "Final checkpoint created at iris_id={}",
                     last_indexed_id
                 ));
+                hawk_handle.sync_peers(false, None).await?;
```
mcalancea: Same question as above.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -1751,3 +1876,212 @@ async fn upload_and_sync_genesis_checkpoint(
     hawk_handle.sync_peers(false, Some(done_rx)).await?;
     Ok(())
 }
+
+async fn get_most_recent_checkpoints(
```
mcalancea: General question about consensus code: can it go into a different file so we don't bloat mod.rs too much? It might help agents navigate stuff more efficiently.

iris-mpc-upgrade-hawk/tests/workflows/genesis_108.rs
```diff
@@ -0,0 +1,155 @@
...
+    // After the second run, party 0's genesis graph checkpoint is deleted from its
+    // CPU-side `genesis_graph_checkpoint` table. This puts party 0 into a state where
+    // it has indexed irises in its iris store but has no record of where the corresponding
+    // HNSW graph was checkpointed on S3.
+    //
+    // The third genesis run must therefore exercise the graph-rollback path:
+    //   - Parties 1 and 2 still have their checkpoint entries and agree on a hash.
+    //   - Party 0 has no entry and contributes a null hash to the consensus protocol.
+    //   - The majority (parties 1 & 2) win the consensus → party 0 downloads and loads
+    //     the S3 checkpoint graph instead of rebuilding from scratch.
```
mcalancea: This is not true right? Party 0 does have a checkpoint which is not the most recent and all the parties agree to use it.

mcalancea (commented):
Looks good to me overall! I left some comments on particulars. 

I'd like to highlight the point on how to handle deletions less aggressively and the one about the manual override for the checkpoint as topics for a brief discussion before merging.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
sdwoodbury: i see the issue now...at the least, checkpoints whose hash fails to convert should be excluded from the list.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
sdwoodbury: this made the integration tests pass. but that was before hawk_session.sync_peers was updated. will try without it.

iris-mpc-upgrade-hawk/tests/workflows/genesis_108.rs
sdwoodbury: yup. that's a hallucination.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
sdwoodbury: it's on the todo list.

iris-mpc-upgrade-hawk/src/genesis/mod.rs
sdwoodbury: i don't think it's that much simpler. but i made the change anyway.

iris-mpc-cpu/src/genesis/genesis_checkpoint.rs
sdwoodbury: the tradeoff is on the infra side - all 3 parties would have to manually modify their databases, which could require a new deployment.

bgillesp: We were talking about exactly this point earlier, actually -- specifically we were looking at keeping the final snapshot of each run as an "archival" snapshot that isn't pruned automatically.  This could be done by adding an extra "archival" field in the snapshots table, and pruning only items not marked as archival.  I forget what we concluded about sequencing, but can this feature be added as a follow-up PR after this one is merged?  It seems fairly independent from the rest of the functionality here.

mcalancea (approved):

PR #2078, opened 2026-04-15
[POP-3736] Handle inter-party timing drift in genesis sync_peers

bgillesp:
- **Fix timeout crash in `sync_peers` during genesis S3 checkpoints.** When one party finishes its S3 graph snapshot upload well before the others, it enters the sync consensus loop first and its network receives time out before the slower parties arrive. The receive logic is now retried independently of sends (avoiding mpsc channel buildup) with a 1-second backoff between retries.
- **Add 5-minute outer timeout on `sync_peers` in the `JobRequest::Sync` handler.** This bounds the retry loop so that persistent failures (crashed peer, broken connection) surface as errors rather than hanging indefinitely.


mcalancea (approved):
Looks sane.

PR #2086, opened 2026-04-20
optimize debug and release profiles for sccache

sdwoodbury:


Cargo.toml
```diff
@@ -93,8 +93,22 @@ ampc-server-utils = { git = "https://github.com/worldcoin/ampc-common.git", rev
 # This improves performance and makes panic propagation more reliable.
 # bench gets its setting from release, and test from dev.
 [profile.release]
+incremental = true
```
mcalancea: This part I'm a bit worried about; I think it reduces the codegen-units (Claude says it's implicitly reduced from 256 to 16), which can make release builds produce less optimized code (regardless if they are clean or incremental).

This change is independent of the sccache stuff right? Is it also required? If it is, we should hard-code the codegen-units so it doesn't get overridden.

mcalancea (commented):
Left a comment on the incremental release builds. 

We should ask somebody from TFH to stamp this.

Cargo.toml
sdwoodbury: good point. there was no need to make release incremental. i removed it

sdwoodbury:
this can be added in ~/.cargo/config.toml

PR #2088, opened 2026-04-22
test: consolidate and expand CPU e2e tests

mcalancea:
## Summary

Consolidates and expands the actor-level CPU e2e test suite. Two tests live in `iris-mpc-cpu/tests/e2e.rs`:

**`e2e_test`** (the existing random-driven test) — now covers every `TestCase` variant at least once. It was previously reaching 9 of 19 variants with the old seed and had deletions + reset_updates disabled via stale `= 0` TODO flags.

- `NUM_BATCHES` 5 → 20, `MAX_BATCH_SIZE` 5 → 10.
- `MAX_DELETIONS_PER_BATCH` and `MAX_RESET_UPDATES_PER_BATCH` both 0 → 3 (deletion and reset_update paths are supported in the actor).
- `TestCaseGenerator` tracks a per-variant pick counter; the test asserts every variant in `default_test_set()` was picked at least once. Fails loudly if a future seed or batch-count change ever silently under-covers.

**`e2e_uniqueness_test`** (new) — 5 deterministic batches built on a single canonical iris X, exercising the search/decision pipeline directly. Built on an `Expectation` enum attached per-request:

- Batch 1: seed X + its 30 non-mirror rotations → slot 0 `UniqueInsert`; rotations all intra-batch-blocked by slot 0.
- Batch 2: all 62 rotation/mirror variants → `DbMatchAt` for non-mirror, `MirrorAttackDetected` for mirror.
- Batch 3: `[reauth X'=X+15, Y=X+30, Z=X-15]` → reauth succeeds; Y blocked intra-batch by the reauth; Z matches the *original* X because reauth mutations apply only after the batch's searches.
- Batch 4: `[Y, Z]` after reauth persists → Y now matches X' in DB; Z is out of rotation window vs X' and inserts fresh, proving reauth displaced X.
- Batch 5: `[delete(seed_db_index), uniqueness(X')]` → deletion acts before same-batch search, so X' re-enrolls at a fresh serial.

On any expectation violation, `check_expectations` collects all failures and panics once with a self-diagnosing per-slot table including each request's `purpose`, expected outcome, observed `ServerJobResult` fields, specific problems, and a source pointer. Runs with tracing always print the table on success too.

### Supporting changes

- Align the bootstrap `HnswSearcher` with the `HawkActor` runtime searcher: `new_linear_scan` + `LINEAR_SCAN_MAX_GRAPH_LAYER=1`, geometric layer distribution `1/20` on both sides.
- Fresh RNG seeds.
- `install_tracing` uses `try_init` so both tests running in the same binary don't fight over the global subscriber.
- `iris-mpc-common::test`: `TestCase` derives `Hash`; `default_test_set()` is `pub`; `TestCaseGenerator` gains `picked_test_cases()` accessor.
- `iris_mpc_cpu::hnsw` re-exports `LayerDistribution`.

## Test plan

- [x] `cargo test --release -p iris-mpc-cpu --test e2e -- --include-ignored --test-threads=1` — both tests pass (~72s together).
- [x] Full `TestCase::default_test_set()` coverage verified on the new seed.
- [x] Cherry-picked onto `mihai/iris-worker-pool-trait-clean` — both tests pass there too.
- [x] `cargo fmt --check` + `cargo clippy -p iris-mpc-cpu --tests --release` — clean.


iris-mpc-cpu/tests/e2e.rs
```diff
@@ -146,24 +160,71 @@ async fn start_hawk_node(
...
+    let (h0, h1, h2) = tokio::join!(
+        start_hawk_node(&args0, db_left, db_right),
+        start_hawk_node(&args1, db_left, db_right),
+        start_hawk_node(&args2, db_left, db_right),
+    );
+    Ok((h0?, h1?, h2?))
+}
+
+/// Run an async test on a 64 MB stack to avoid platform-dependent stack
+/// overflows in release mode.
```
sdwoodbury, line 205: I know that this works as-is but I'm unsure why the `stack_size` is set on the `std::thread` instead of the `runtime::Builder`. I think that threads spawned by the tokio runtime do not inherit the stack size of the `std::thread`.

iris-mpc-cpu/tests/e2e.rs
```diff
@@ -236,3 +264,745 @@ async fn e2e_test_async() -> Result<()> {
...
+    /// Blocked by an earlier mutating slot in the same batch.
+    IntraBatchBlockedBy { earlier_request_id: String },
+    /// `full_face_mirror_attack_detected` raised, mirror match at the given index.
+    MirrorAttackDetected { expected_mirror_match_index: u32 },
+    /// Reauth target updated.
+    ReauthSucceeds { target_index: u32 },
+}
+
+impl Expectation {
+    /// Source pointer printed in failure messages, to orient debugging.
```
sdwoodbury: nit: "source file" instead of "source pointer"

iris-mpc-cpu/tests/e2e.rs
```diff
@@ -236,3 +264,745 @@ async fn e2e_test_async() -> Result<()> {
...
+            .into_iter()
+            .next()
+            .expect("rotations_from_range returns one element")
+    }
+}
+
+/// Rotation+mirror variant of (x_left, x_right). For `mirror=true` the eyes
+/// swap sides AND each is mirrored — the canonical full-face-mirror-attack
+/// construction, matching Mirror-orientation search expectations.
+fn build_mirror_variant_pair(
```
sdwoodbury: nit: this should not be called "build_mirror_variant..." because the function doesn't always do that. something like "build_iris_pair" is more suitable.

sdwoodbury (dismissed):

iris-mpc-cpu/tests/e2e.rs
mcalancea: Fixed in 8a9394c5.

iris-mpc-cpu/tests/e2e.rs
mcalancea: Renamed to `build_iris_pair` in 8a9394c5.

iliailia (approved):

PR #2090, opened 2026-04-26
Add documentation for fields of some Result structs

bgillesp:
Added doc strings for fields of the Result structs for uniqueness, reauth, and identity match (reset and recover) types, correlated with documentation in the `signup-service` repository, and with the construction logic in the main `process_job_result` function.


mcalancea (approved):
Looks good!

PR #2093, opened 2026-04-28
test: Improve E2E tests with a mirror batch match + fix reporting

dkales:
This PR adds an E2E test case specifically for intra-batch mirror matches.
The GPU version was actually already handling this correctly, the only thing that was missing was reporting the matching request id from the mirror matches as well.
However, this is just a cosmetic/logging fix, as the real decision of `is_match` was already computed correctly.

A few more details:

In iris-mpc-gpu/src/server/actor.rs:868-885, the CompactQuery for each side is built like this:

code_query / mask_query (the "a" operand) → get_iris_interpolated_requests_preprocessed(side, orientation) — this is orientation-aware: in Mirror mode for Eye::Left it returns right_mirrored_iris_interpolated_requests_preprocessed (opposite eye, mirrored).
code_query_insert / mask_query_insert (the "b" operand) → get_iris_requests_rotated_preprocessed(side) at iris-mpc-gpu/src/server/mod.rs:180-188 — this method takes no orientation parameter and always returns the normal, same-side rotated requests.
The intra-batch self-check in actor.rs:1762-1870 calls compute_dot_products, which dots code_query against code_query_insert, which is the expected mirror - normal dot product for the mirror orientation.

### DIFF
--- a/iris-mpc-common/src/test.rs
+++ b/iris-mpc-common/src/test.rs
@@ -374,6 +374,8 @@ pub struct TestCaseGenerator {
     /// skip invalidating requests in the current batch, since we expect
     /// them to be processed
     skip_invalidate: bool,
+    /// Do we already have a mirror of a template in the current batch?
+    have_batch_mirror: bool,
     /// duplicates in the current batch, used to test the batch
     /// deduplication mechanism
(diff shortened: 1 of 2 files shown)


mcalancea (approved):
LGTM!

PR #2094, opened 2026-04-28
[POP-3785] move graph checkpoint code to a common location

sdwoodbury:
in preparation for POP-3783, move the genesis graph checkpoint code into `iris-mpc-cpu/src/graph_checkpoint/` so that `hawk` and `genesis` can share code. 

No functionality was changed in this PR, though `aws/clients.rs` was modified to add the checkpoint s3 client to the `AwsClients` struct.


bgillesp (approved):
Looks fine, thanks!

PR #2096, opened 2026-04-29
[POP-3689] Make quick-sort partial in sorted neighborhoods

iliailia:
This change prunes the quick-sort sorting network to sort only the first `k` elements.

Tested on a graph with 100,000 irises, this results in a net savings: ~0.1% less bytes sent and ~0.9% less rounds per batch.


bgillesp (approved):
Looking great, thanks!

PR #2097, opened 2026-04-29
[POP-3788] add GraphMutation

sdwoodbury:
Add the data model and persistence layer for graph mutations, without changing existing functionality. This will unblock development of the checkpoint sidecar service. 

note that there are now two functions to download the graph, one for a IrisVectorId and one for a PlaintextIris. This is to facilitate decoding different graph versions in the future, which will require a concrete type rather than a generic.


iliailia (dismissed):
LGTM!

iliailia (approved):

PR #2098, opened 2026-04-30
Fully decouple HawkActor from iris storage and worker-pool concretes

mcalancea:
## Summary

Decouples `HawkActor` from iris storage and worker-pool concrete types, and unifies the construction path between hawk-main and genesis.

### `IrisWorkerPool` is the only iris-data interface

- `IrisWorkerPool` is dyn-compatible: `BoxFuture` returns, no `Clone` supertrait, blanket impl for `Arc<T: ?Sized + IrisWorkerPool>`.
- `Aby3Store` drops the `W` generic; `pub workers: Arc<dyn IrisWorkerPool>`.
- `HawkActor.worker_pools: BothEyes<Arc<dyn IrisWorkerPool>>`. `HawkSession` and the search/insert pipeline stay concrete.
- All iris reads — including the genesis persistence thread — go through `worker_pools[side].fetch_irises(...)`.

### `HawkActor` holds no iris data and no concrete pool handles

- `iris_store` field, `iris_store(StoreId)` getter, `refresh_registries`, and `assert_registry_consistency` are gone.
- `loader_handles` and `IrisPoolHandle` no longer leak past the pool's own constructor.
- Field set: `args, searcher, prf_key, registry, graph_store, worker_pools, anon_stats_store, networking, error_ct, party_id`.

### Registry construction is independent of the iris load

- New `Store::stream_iris_ids` runs an index-only `(id, version)` scan.
- `InitializedWorkers` produces metadata-only registries from the same scan, in parallel with the full iris load.
- After load, registry-side and pool-side checksums are cross-checked; mismatch = loader dropped or duplicated rows.
- Each eye gets an independent `Arc<RwLock>` registry — runtime inserts allocate `next_id` per eye, and a shared Arc would burn two ids per logical insert.

### Construction goes through `HawkActorPrelude`

- The prelude holds only the network handle.
- Genesis runs `prelude.sync_peers()` + iris-DB rollback before deciding the final `max_serial_id`.
- Hawk-main has no pre-load step today but uses the same shape; ready for any future iris-count negotiation.
- `prelude.finalize(initialized, graph)` is the only path that constructs a fully-loaded `HawkActor`. Iris loading is always done before the actor exists.

## Note: `fake_db_size` removed from Hawk main

Hawk main no longer honors `SMPC__FAKE_DB_SIZE > 0` — the `LocalInitMode::FakeDb` mode and its server branch are deleted, and the server now bails if a non-zero value is set. All current production deploys set `SMPC__FAKE_DB_SIZE: "0"`, but **we are uncertain whether anyone uses the fake-db shortcut for local benchmarks, dev loops, or CI smokes**. The `config.fake_db_size` field itself is left intact (the older GPU server still reads it). If anyone on the team relies on the Hawk-main fake-db path, the relevant commit (`a84aad5a`) can be reverted in isolation without disturbing the rest of the PR.

## Test plan
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (no warnings)
- [x] `cargo test -p iris-mpc-cpu --lib` (174/174)
- [x] `cargo test -p iris-mpc-upgrade-hawk --lib`
- [x] cpu-integration-tests reproduced locally and pass with the per-eye-Arc fix
- [ ] Full CI


sdwoodbury (dismissed):

iris-mpc-cpu/src/execution/hawk_main/worker_pool_initializer.rs
```diff
@@ -0,0 +1,296 @@
...
+        right_mask: &[u16],
+    ) {
+        for (pool, code, mask) in izip!(
+            &self.iris_pools,
+            [left_code, right_code],
+            [left_mask, right_mask]
+        ) {
+            let iris = GaloisRingSharedIris::try_from_buffers(self.party_id, code, mask)
+                .expect("Wrong code or mask size");
+            pool.insert(vector_id, iris).unwrap();
```
iliailia, line 272: does it make sense to panic with the message here by switching to `.expect("iris worker thread disconnected during load")`?

iris-mpc-cpu/src/execution/hawk_main/worker_pool_initializer.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    fn increment_db_size(&mut self, _index: usize) {
+        self.db_size += 1;
+    }
+
+    fn reserve(&mut self, additional: usize) {
+        let additional = additional + (additional as f64 * IRIS_STORE_RESERVE_EXTRA) as usize;
+        for side in &self.iris_pools {
+            side.reserve(additional).unwrap();
```
iliailia, line 283: does it make sense to panic with the message here by switching to `.expect("iris worker thread disconnected during load")`?

iris-mpc-cpu/src/execution/hawk_main/worker_pool_initializer.rs
```diff
@@ -0,0 +1,296 @@
...
+            if expected != got {
+                return Err(eyre!(
+                    "worker pool checksum mismatch on side {side}: \
+                     registry expected {expected:#x}, pool reported {got:#x} \
+                     — loader dropped or duplicated rows vs the registry scan"
+                ));
+            }
+        }
+
+        tracing::info!(
```
iliailia, line 219: Mb, adding `mode` would be valuable.

iliailia (dismissed):
Thanks, left a few minor comments

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -474,23 +470,51 @@ pub struct HawkInsertPlan {
...
-            [(); 2].map(|_| GraphMem::new()),
-            [(); 2].map(|_| Aby3Store::<HawkOps>::new_storage(None)),
-        )
-        .await
+/// Partial actor with only the network handle. Lets MPC peer
+/// communication run before iris data is loaded — genesis needs
+/// `sync_peers` to complete before iris-DB rollback fixes
+/// `max_serial_id`. `finalize` consumes the prelude with an
+/// initializer's output and a graph to produce a `HawkActor`.
+pub struct HawkActorPrelude {
```
bgillesp: This construction of a `HawkActorPrelude` that is first initialized, then finalized, and gives access to a single networking call, feels too heavy for what it accomplishes (genesis needs networking before finalizing `HawkActor`).  Could we instead just have a function call which initializes and returns only the networking, and then have the user pass the networking as an arg into the constructor for a HawkActor?

In fact, could we refactor this setup into a `HawkActorBuilder` struct, so that we reduce the propagation of `from_cli*` constructors, and provide a canonical path for getting a `HawkActor` in the end?  It seems a bit confusing/indirect right now that there are paths to getting a `HawkActor` both through `HawkActorPrelude::new/finalize` and the `HawkActor` test utility functions.  To handle the networking thing, have a separate function call that generates the networking from a `HawkArgs` as suggested above, then an optional builder function to pass in the networking if it's been manually constructed/used externally.  If not specified for the builder, it constructs a new networking when building based on the provided args.

mcalancea: Thanks for the suggestions. I'm walking back the prelude thing. The actor builder might be a better fit for a subsequent PR?

bgillesp (approved):
This is looking great, thanks for the change on the hawk actor construction design.

PR #2112, opened 2026-05-06
[POP-3798] Optimized Neighborhood Encoding

iliailia:
## Summary
- New module `iris_mpc_cpu::hnsw::graph::encoded_neighborhood::EncodedNeighborhood`: a compact at-rest format for sorted, strictly-increasing `Vec<u32>` neighborhood IDs.
- Wire layout is a 3-byte header (`k: u16 LE`, `b: u8`) followed by an MSB-first Rice-coded body. IDs are gap-encoded (`ids[0]` absolute, then `ids[i] - ids[i-1] - 1`) and Rice/Golomb-coded with a power-of-two divisor `2^b`.
- The Rice parameter `b` is chosen per neighborhood as `floor(log2(μ))` over the exact mean of the encoded values — about half a bit too high vs. the bit-optimal `floor(log2(μ·ln 2))`, but cheap to compute and good for any neighborhood size.
- Includes a criterion benchmark suite, round-trip / malformed neighborhood / fuzz tests, and an analytical-size sanity check.

## Wire format

```
+-------------+-----+-----------------------------+
| k (u16 LE)  | b   | Rice-coded body, MSB-first  |
+-------------+-----+-----------------------------+
 bytes 0..2    2     3..
```

Each Rice-coded value `v` is `q = v >> b` zero bits, a `1` terminator (self-delimiting), then the low `b` bits of `v`. Body is right-padded to whole bytes. Module docs include a worked example pinning the bytes of `[0, 5, 9]`.

## Benchmarks

M1 Pro, criterion default sample/warmup, single-thread.

**Encode/decode at k=450 across universes** (≈ HNSW M=64 neighborhoods at varying database sizes):

| Universe | encode  | decode  | encode thrpt | decode thrpt |
|----------|---------|---------|--------------|--------------|
| n=1M     | 6.27 µs | 8.37 µs | 71.8 Melem/s | 53.8 Melem/s |
| n=10M    | 7.36 µs | 9.82 µs | 61.2 Melem/s | 45.8 Melem/s |
| n=100M   | 8.58 µs | 11.49 µs| 52.4 Melem/s | 39.2 Melem/s |
| n=2³²    | 11.01 µs| 14.60 µs| 40.9 Melem/s | 30.8 Melem/s |

**Scaling at n=10M across k** (per-element cost amortizes once k ≳ 100):

| k     | encode  | decode  | encode thrpt | decode thrpt |
|-------|---------|---------|--------------|--------------|
| 10    | 302 ns  | 272 ns  | 33.0 Melem/s | 36.8 Melem/s |
| 100   | 1.78 µs | 2.24 µs | 56.1 Melem/s | 44.7 Melem/s |
| 450   | 7.31 µs | 10.72 µs| 61.5 Melem/s | 42.0 Melem/s |
| 2000  | 33.03 µs| 42.52 µs| 60.6 Melem/s | 47.0 Melem/s |

**Cap-slack sensitivity** (separate one-off run; not in the committed bench suite): encode timing was flat within ~0.7 % across `slack ∈ {2, 3, 4, 6, 8, 16}` for k ∈ {10, 450, 2000, 65535}. The `+ 4` slack in `BitWriter::with_capacity` is empirically safe — the body Vec doesn't reallocate. Reasoning is documented above the cap line.

**Encoded size vs. analytical model.** `encoded_size_matches_model` asserts the produced byte length is within ±10 % of the `(k·(b+2) + 24) / 8` model across `(universe, expected_b) ∈ {(1M, 11), (10M, 14), (100M, 17), (2³², 23)}` at k=450.

## Test plan
- [x] `cargo test -p iris-mpc-cpu --lib hnsw::graph::encoded_neighborhood` — 20 tests pass (round-trip across single-element / consecutive / handcrafted / widely-spaced / max-k / high-universe / random seeds; encode error paths; malformed-blob decode paths; garbage-input fuzz; size-vs-model).
- [x] `cargo bench -p iris-mpc-cpu --bench encoded_neighborhood` — runs cleanly (numbers above).

## Follow up
- Add conversion from the current `VectorId` to `u32`.
- Wire the encoder into the `GraphPg` write path and the decoder into the read path.

### DIFF
--- a/iris-mpc-cpu/src/hnsw/graph/encoded_neighborhood.rs
+++ b/iris-mpc-cpu/src/hnsw/graph/encoded_neighborhood.rs
@@ -40,6 +40,11 @@ impl EncodedNeighborhood {
         if ids.len() > MAX_K {
             return Err(EncodeError::TooLarge(ids.len()));
         }
+        for i in 1..ids.len() {
+            if ids[i] <= ids[i - 1] {
+                return Err(EncodeError::NotSorted(i));
+            }
+        }
         let k = ids.len() as u16;
         let b = 0u8;
         let mut out = Vec::with_capacity(HEADER_LEN);
@@ -86,5 +91,27 @@ mod tests {
         let decoded = encoded.decode().expect("decode empty");
         assert!(decoded.is_empty());
         assert_eq!(encoded.as_bytes().len(), HEADER_LEN);
+    }
+
+    #[test]
+    fn rejects_unsorted() {
+        let err = EncodedNeighborhood::encode(&[5, 3]).unwrap_err();
+        assert!(matches!(err, EncodeError::NotSorted(1)), "got {:?}", err);
+    }
+
+    #[test]
+    fn rejects_duplicates() {
+        let err = EncodedNeighborhood::encode(&[5, 5]).unwrap_err();
+        assert!(matches!(err, EncodeError::NotSorted(1)), "got {:?}", err);
+    }
+
+    #[test]
+    fn rejects_too_large() {
+        let oversized: Vec<u32> = (0..(MAX_K as u32 + 1)).collect();
+        let err = EncodedNeighborhood::encode(&oversized).unwrap_err();
+        match err {
+            EncodeError::TooLarge(n) => assert_eq!(n, MAX_K + 1),
+            other => panic!("unexpected error: {other:?}"),
+        }
     }
 }
--- a/iris-mpc-cpu/src/hnsw/graph/encoded_neighborhood.rs
+++ b/iris-mpc-cpu/src/hnsw/graph/encoded_neighborhood.rs
@@ -1,3 +1,50 @@
+//! Compact at-rest encoding for HNSW graph neighborhoods.
+//!
+//! A neighborhood is a sorted, strictly-increasing slice of `u32` IDs. We
+//! delta-encode it (`ids[0]` kept as an absolute value, then each later id as
+//! `ids[i] - ids[i-1] - 1`) and Rice-code the resulting `k` values into a
+//! packed bitstream. Decoding reverses this: read the header, Rice-decode `k`
+//! symbols, then prefix-sum them back into ids.
+//!
+//! # Wire format
+//!
+//! ```text
+//! +-------------+-----+-----------------------------+
+//! | k (u16 LE)  | b   | Rice-coded body, MSB-first  |
+//! +-------------+-----+-----------------------------+
+//!  bytes 0..2    2     3..
+//! ```
+//!
+//! Each Rice-coded value `v` is written as `q = v >> b` zero bits, a `1`
+//! terminator, then the low `b` bits of `v`. The Rice parameter `b` (which is
+//! a power-of-two Golomb divisor) is the same for every symbol in a blob and
+//! is chosen as described below. The body is zero-padded on the right to a
+//! whole number of bytes.
+//!
+//! # Choosing `b`
+//!
+//! For values drawn from a geometric distribution with mean `μ`, the
+//! bit-optimal Rice parameter is `b ≈ floor(log2(μ · ln 2))`. We approximate
+//! this as `floor(log2(μ))` (about half a bit too high on average — at most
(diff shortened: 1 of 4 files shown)


iris-mpc-cpu/src/hnsw/graph/encoded_neighborhood.rs
```diff
@@ -0,0 +1,560 @@
...
+    QuotientOverflow(usize),
+    #[error("decoded id overflow at gap {0}")]
+    IdOverflow(usize),
+}
+
+impl EncodedNeighborhood {
+    /// Encode a sorted-ascending, strictly-increasing slice of u32 IDs.
+    pub fn encode(ids: &[u32]) -> Result<Self, EncodeError> {
+        if ids.is_empty() {
+            return Err(EncodeError::Empty);
```
bgillesp: It seems overly strict to require that neighborhoods be non-empty -- is there a reason we can't just interpret an empty Rice-coded body segment as an empty neighborhood?

iris-mpc-cpu/src/hnsw/graph/encoded_neighborhood.rs
```diff
@@ -0,0 +1,560 @@
...
+        }
+        if ids.len() > MAX_K {
+            return Err(EncodeError::TooLarge(ids.len()));
+        }
+        for i in 1..ids.len() {
+            if ids[i] <= ids[i - 1] {
```
bgillesp, line 92: I think this should be strict inequality to ensure strict increasing property.

iris-mpc-cpu/src/hnsw/graph/encoded_neighborhood.rs
```diff
@@ -0,0 +1,560 @@
...
+    31u8.saturating_sub(mean_gap.leading_zeros() as u8)
+}
+
+fn rice_encode(writer: &mut BitWriter, value: u32, b: u8) {
+    let q = if b == 0 { value } else { value >> b };
+    writer.write_zeros(q);
+    writer.write_one();
+    if b > 0 {
+        let mask = (1u32 << b) - 1;
+        writer.write_bits(value & mask, b);
```
bgillesp: I think `write_bits` already only writes the low `b` bits from the input -- mask probably isn't necessary?

iris-mpc-cpu/src/hnsw/graph/encoded_neighborhood.rs
```diff
@@ -0,0 +1,560 @@
...
+    }
+}
+
+fn rice_decode(reader: &mut BitReader<'_>, b: u8, gap_idx: usize) -> Result<u32, DecodeError> {
+    debug_assert!(b <= 31, "rice parameter b must be 0..=31");
```
bgillesp, line 361: Why doesn't b = 32 work?  For u32 inputs I think this would be just writing the deltas as 32 bits and delimiting by 1s for the unary q=0.  Not efficient, but it seems like it may be valid?

bgillesp (commented):
Great implementation, thank  you!  A few minor comments, plus see #2117 for some optimizations on the BitReader/Writer structs.  Otherwise it's looking good.

iris-mpc-cpu/src/hnsw/graph/encoded_neighborhood.rs
iliailia: It's the negation of strict increasing property. The following element can't be equal or smaller than its predecessor.

iris-mpc-cpu/src/hnsw/graph/encoded_neighborhood.rs
iliailia: Yes, it's valid, but `b=32` leads to strictly worse encoding sizes. `compute_b` actually enforces `b <= 31`, so this is just a safe check that we don't pass inefficient `b` computed somewhere else.

iris-mpc-cpu/src/hnsw/graph/encoded_neighborhood.rs
bgillesp: Oh whoops, sorry, yeah that's right.

bgillesp (approved):

PR #2114, opened 2026-05-06
[POP-3799] improve logging

sdwoodbury:
Within iris-mpc-cpu, propagate spans across spawned tasks. For unit tests which spawn multiple MPC parties, attach spans to identify which logs are emitted by which party.

### DIFF
--- a/iris-mpc-common/src/test.rs
+++ b/iris-mpc-common/src/test.rs
@@ -29,7 +29,6 @@ use std::{
     collections::{HashMap, HashSet},
     ops::Range,
 };
-use tracing::{info_span, Instrument};
 use uuid::Uuid;
 
 const THRESHOLD_ABSOLUTE: usize = IRIS_CODE_LENGTH * 345 / 1000; // 0.345 * 12800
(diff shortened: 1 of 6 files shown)


iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -1952,18 +1959,19 @@ impl HawkHandle {
                     is_match_batch(search_queries, step1.missing_vector_ids(), sessions_search)
                         .await?;
 
-                step1.step2(&missing_is_match, intra_results.await??)
+                step1.step2(&missing_is_match, intra_results.await???)
```
mcalancea, line 1962: We should be able to knock this down to just `??`.

iris-mpc-common/src/test.rs
```diff
@@ -1366,8 +1379,63 @@ impl TestCaseGenerator {
...
+        // idx == u32::MAX means no insertion happened (NON_MATCH_ID sentinel)
+        // This can happen for invalidated requests
+        if idx != u32::MAX {
+            // Check if another party already processed this insertion
+            if let std::collections::hash_map::Entry::Vacant(e) = self.inserted_responses.entry(idx)
+            {
+                // First party processing this insertion
+                // Bounds check: new insertion idx must be >= initial_db_len
+                if idx < initial_db_len {
+                    tracing::warn!(
```
mcalancea, line 1405: Just warn, not error or panic? It seems like an unsound thing to happen.

mcalancea:
On `verify_graph_consistency` in `iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs`:

There's existing graph-diff infrastructure under `iris-mpc-cpu/src/hnsw/graph/graph_diff/` that covers what this helper does:

```rust
fn verify_graph_consistency(role: usize, mpc: &GraphMem<VectorId>, pt: &GraphMem<VectorId>) -> Result<()> {
    ensure_node_equivalence(mpc, pt)
        .map_err(|e| eyre!("[role {role}] {e:?}"))?;
    let diff = run_diff(mpc, pt, ExplicitNeighborhoodDiffer::new(SortBy::Index));
    let has_diffs = diff.0.iter()
        .any(|l| l.diffs.iter().any(|(_, d)| !d.only_in_lhs.is_empty() || !d.only_in_rhs.is_empty()));
    if has_diffs {
        bail!("[role {role}] neighborhoods differ:\n{diff}");
    }
    ensure!(mpc.entry_points == pt.entry_points, "[role {role}] entry points differ");
    Ok(())
}
```

Side benefit: since this returns `Err` instead of panicking, the call site can collect results across all three roles before failing — currently a divergence at role 0 short-circuits the `for role in 0..3` loop and you never learn whether roles 1 and 2 also disagreed.

iris-mpc-cpu/src/execution/hawk_main/scheduler.rs
```diff
@@ -3,10 +3,34 @@ use eyre::{eyre, Result};
...
-use tokio::{sync::mpsc::UnboundedReceiver, task::JoinError};
+use tokio::{sync::mpsc::UnboundedReceiver, task::JoinError, task::JoinHandle};
+use tracing::{Instrument, Span};
 
 const N_EYES: usize = 2;
 
+/// Helper to spawn a task with the current tracing span preserved.
+/// This ensures that child tasks inherit and maintain the parent span context.
+/// The future should return a Result type.
+pub fn spawn_with_span<F, T>(task: F) -> JoinHandle<Result<T>>
```
mcalancea, line 14: This naming suggests the span itself might be new (for example, a child span attached to the parent). The difference can be important (especially for hierarchical consumers of the spans). Maybe the name can be improved?

mcalancea (commented):
Looks good! Should reuse the differ if not a headache. Note the comment by Claude above.

iris-mpc-cpu/src/execution/hawk_main.rs
sdwoodbury: rust analyzer gives me an error with only two question marks.

iris-mpc-cpu/src/execution/hawk_main/scheduler.rs
sdwoodbury: i updated this to take the span as an argument. that should eliminate the possibility of confusion and be more flexible.

mcalancea (dismissed):

mcalancea (approved):

iliailia (approved):

PR #2116, opened 2026-05-06
[POP-3753] add explicit synchronization to genesis

sdwoodbury:
This is still needed at the end of `exec_indexation`, at least for the integration tests. Otherwise one of the 3 MPC parties could do `sync_state()` (previously called `sync_peers`), exit, and then have the other parties try to do `health_check` + `sync_state()` and have their networking stack fail because one of the parties already quit. 

This was needed to make tests pass after fully implementing the WAL.


iliailia (approved):
LGTM!

PR #2118, opened 2026-05-07
[POP-3753] implement write ahead log

sdwoodbury:
Use a `GraphMutation` enumeration instead of building up destructive overwrites via `ConnectPlan`s. `ConnectPlan` becomes a list of `GroupedMutations` which is a list of `GraphMutation`s associated with each request (modification). 

this is a large PR but the commits are picked to focus on non-overlapping areas so that the commits can be reviewed in sequence instead of all at once. 

~this branch needs to be rebased onto main after PR #2114  and PR #2116 are merged. (this will also fix the errors caused by missing dependencies)~


iris-mpc-common/src/test.rs
sdwoodbury, line 1: changes in this file were supposed to be in PR 2114

iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
sdwoodbury, line 1: changes from this file were also supposed to be in PR 2114

iris-mpc/src/server/mod.rs
```diff
@@ -592,22 +588,20 @@ async fn load_database(
 
     // Try to load graph from S3 checkpoint first, then fall back to
     // Postgres graph representation for temporary legacy compatibility.
-    // TODO simplify this logic once graph DB tables are removed.
+    // TODO: apply GraphMutations to the checkpoint
```
sdwoodbury, line 591: i think this will be addressed in a future PR

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -432,18 +488,80 @@ impl<V: Ref + Display + FromStr + Ord> Layer<V> {
...
+
+    pub fn insert_node(&mut self, id: V, neighbors: Vec<V>) {
+        self.set_links(id.clone(), neighbors.clone());
+    }
+
+    pub fn add_neighbor(&mut self, id: V, neighborhoods: Vec<V>) {
+        for nbhd in &neighborhoods {
+            if let Some(neighbor_links) = self.links.get_mut(nbhd) {
+                match neighbor_links.binary_search(&id) {
+                    Err(i) => {
```
mcalancea: `binary_search + insert` keeps `neighbor_links` sorted. The invariant: `set_links` must receive already-sorted neighbors. `insert_prepare_batch` does sort them (`searcher.rs:1611`), so this holds today — but a future caller bypassing the planner could break it silently (binary_search on an unsorted Vec returns wrong results, leading to wrong-position inserts and undetectable graph corruption). Add a `// invariant: neighbor_links is kept sorted; relies on the planner sorting before set_links` comment so the dependency is explicit.

iris-mpc-cpu/src/hnsw/graph/mutation.rs
```diff
@@ -41,6 +83,6 @@ impl<Vector: Ord> GraphMutation<Vector> {
     /// are modified in an incompatible way. Keeping this separate allows old
     /// mutations to remain stored and handled appropriately during format upgrades.
     pub fn get_version() -> i32 {
```
mcalancea: **Version bump 1 → 2 needs a v1 compatibility path on read.** After commit `cc0d74f8`, v1 rows still deserialize (no variants were removed), but on replay v1's `InsertNode` no longer triggers backlinks — they're now an explicit `AddNeighbor` mutation that v1 rows don't contain. Without a `if row.mutation_version == 1 { synthesize AddNeighbor from each InsertNode }` step on read, **any deployment with v1 rows in its WAL silently replays into a graph with no backlinks**. Either land the synthesis or document 'drop the WAL on upgrade and restart from a fresh checkpoint' in the deploy story.

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -432,18 +488,80 @@ impl<V: Ref + Display + FromStr + Ord> Layer<V> {
...
     }
 
+    /// Order-agnostic checksum of this layer's link map. Two layers with the
+    /// same `(node, set-of-neighbors)` content produce the same checksum even
+    /// if their `HashMap` iteration order or internal `Vec` ordering differ.
+    pub fn checksum(&self) -> u64 {
+        self.set_hash.checksum()
+    }
+
+    pub fn insert_node(&mut self, id: V, neighbors: Vec<V>) {
```
mcalancea: Was: 'Insert a node with bidirectional links to its neighbors' (doc removed in `cc0d74f8`). Now it's literally `self.set_links(id.clone(), neighbors.clone())` — no backlinks. Any caller who doesn't also emit `AddNeighbor` produces a graph node with no incoming edges. Either rename (`set_node`?), inline at the one call site in `insert_apply`, or add a doc comment making the new (lack of) semantics explicit.

iris-mpc-cpu/src/execution/hawk_main/state_check.rs
```diff
@@ -27,14 +27,72 @@ pub struct SetHash {
...
     pub fn remove(&mut self, value: impl Hash) {
         self.accumulator = self.accumulator.wrapping_sub(Self::hash(value));
     }
 
+    /// Add a `(key, set-of-items)` pair to the accumulator without any
+    /// dependence on the iteration order of `items`. Each `(key, item)` pair
+    /// is hashed and folded in individually via wrapping addition, and a
+    /// one-shot key marker is added so that an empty `items` collection is
+    /// still distinguishable from the key being absent altogether.
+    pub fn add_unordered_set<K, I, T>(&mut self, key: K, items: I)
```
mcalancea, line 53: No unit tests for `add_unordered_set` / `remove_unordered_set` covering the four properties the design relies on: order-independence over `items`, empty-vs-absent key distinction (the marker is the whole point), add+remove identity, multiset semantics on duplicate items. ~20 lines of tests would lock all of these in.

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -432,18 +488,80 @@ impl<V: Ref + Display + FromStr + Ord> Layer<V> {
...
+        self.set_links(id.clone(), neighbors.clone());
+    }
+
+    pub fn add_neighbor(&mut self, id: V, neighborhoods: Vec<V>) {
```
mcalancea: Per-neighborhood cost: M+2 SipHasher finalizations (from `remove_unordered_set` + `add_unordered_set`) plus an O(M) `Vec::insert` shift inside the inner loop. With remove-then-add pairs per backlink update, this is ~2(M+2) SipHasher calls per touched neighbor. Worth a microbench against checksum read frequency before merge — if `set_hash::checksum()` is rarely called, sort-at-read could be cheaper than fold-on-every-write.

iris-mpc-cpu/src/hnsw/graph/mutation.rs
```diff
@@ -1,27 +1,69 @@
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, Default, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct GroupedMutations<V: Ord>(pub Vec<GraphMutation<V>>);
+
 /// Represents a diff to apply to an existing graph.
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum GraphMutation<Vector: Ord> {
```
mcalancea, line 15: **WAL replay still depends on code for end-state derivation.** `AddNeighbor` + `RemoveInvalidNeighbors` (commit `cc0d74f8`) made *which* nodes get touched explicit — good. But each touched node's final link list is still re-derived at replay time by `Layer::add_neighbor` / `Layer::compact_node` / `Layer::set_links`. Any future change to dedup, ordering, or missing-target handling in those functions silently changes how old WAL entries replay. 

Should think about how to enforce this contract.

iris-mpc-cpu/src/hnsw/searcher.rs
```diff
@@ -2029,40 +2048,51 @@ mod tests {
...
-
-        // Verify that each neighbor's updated neighborhood has exactly 4
-        // elements (the original 4 neighbors plus the newly inserted vector,
-        // trimmed to M_max=4 in layer 0)
-        for ((id, _lc), nbhd) in &plan.updates {
-            if *id != next_id {
-                assert_eq!(nbhd.len(), 4);
-            }
+        // Verify the grouped plan was created correctly
+        assert_eq!(grouped_plans.len(), 1);
```
mcalancea, line 2117: Old test asserted `updates.len() == 6` and that each backlink-receiver's neighborhood was trimmed to M_max=4. New test only checks the InsertNode shape — no assertion that any `Compact` was produced, never mind correct target / `to_remove`. This commit added the Compact-attribution logic and the test simultaneously stopped covering it. Please restore assertions on the produced `Compact` mutations (count, per-(id,layer) target, `to_remove` contents).

iris-mpc-cpu/src/hnsw/searcher.rs
```diff
@@ -1725,29 +1707,66 @@ impl HnswSearcher {
...
+                    .iter()
+                    .filter(|v| !compacted_set.contains(v))
+                    .cloned()
+                    .collect();
+                if !to_remove.is_empty() {
+                    let group_idx = insert_to_group
+                        .get(id)
+                        .copied()
+                        .or_else(|| nbhd_last_group.get(&(id.clone(), *layer)).copied())
+                        .unwrap_or(last_some_idx);
```
mcalancea, line 1761: Attaching `Compact` to `nbhd_last_group = max(group_idx that touched N)` is correct *because* no later group touches N, so apply-time state of N equals planner-time `final_nbhds[N]`. This invariant is load-bearing for correctness and easy to break silently in a future refactor (e.g. `max` → `min` or `first`). Worth a one-line `// invariant:` comment on this block.

iris-mpc-cpu/src/execution/hawk_main/test_utils.rs
```diff
@@ -90,19 +93,20 @@ pub async fn init_graph(actor: &mut HawkActor) -> Result<()> {
...
+            } else {
+                UpdateEntryPoint::False
             };
-            graph.insert_apply(plan).await;
+            let mutations = vec![GraphMutation::InsertNode {
+                id: id(i),
+                layers: vec![(0, edges(i))],
+                update_ep,
+            }];
+            graph.insert_apply(mutations);
```
mcalancea, line 117: After commit `cc0d74f8`, `InsertNode` no longer adds backlinks (that's now `AddNeighbor`). This init helper emits only `InsertNode`, so the produced graph has *no* backlinks at all. Compounds with the topology change from commit 5 (the `build_layer_updates` removal). Fix: emit a paired `AddNeighbor` alongside each `InsertNode`.

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
```diff
@@ -289,6 +271,81 @@ impl<V: VectorStore> GraphPg<V> {
...
+                "CREATE TABLE {main_table} AS TABLE {backup_table}"
+            ))
+            .execute(&self.pool)
+            .await?;
+        }
+
+        Ok(())
+    }
+
     pub async fn insert_hawk_graph_mutations(
```
mcalancea, line 283: **Determinism convention break in the underlying schema.** `hawk_graph_mutations.id` uses `GENERATED ALWAYS AS IDENTITY`, which drifts across parties on rollback/retry. Migration `20250313141530_modifications_table_v2` deliberately moved `modifications.id` *off* IDENTITY and onto a `BIGINT PRIMARY KEY` + `BEFORE INSERT` trigger that assigns `MAX(id)+1` from committed contents — precisely so the id matches across parties. The new column reintroduces the drift, and `assert_modifications_consistency` (sync.rs:334) compares the resulting `graph_mutation_id` cross-party. Either use the same trigger-based pattern in the migration, or make `modification_id` the natural key.

iris-mpc/src/server/mod.rs
```diff
@@ -592,22 +588,20 @@ async fn load_database(
...
-            graph_loader.load_graphs_from_checkpoint(both_eyes);
-            Ok(())
+            download_graph_checkpoint(s3_client, checkpoint_bucket, &state).await?
         } else {
-            tracing::info!("No S3 checkpoint found, loading from PostgreSQL");
-            graph_loader
-                .load_graph_store(graph_store, parallelism)
-                .await
-        }
+            tracing::info!("No S3 checkpoint found, defaulting to empty graph");
```
mcalancea, line 600: Dramatic operational event (either fresh genesis or 'we lost the graph') logged at `info!`. Should be `warn!` minimum, or gate behind an explicit `--allow-empty-graph` flag so a misconfigured deployment can't silently come up empty.

Mihai: Claude is being a bit snarky here, but yeah we should probably be explicit about when we want to start with the empty graph.

iris-mpc-cpu/src/execution/hawk_main.rs
```diff
@@ -1733,51 +1708,44 @@ impl HawkMutation {
...
-                        UpdateEntryPoint::SetUnique { layer } => {
-                            graph.set_entry_point(plan.inserted_vector, layer).await?;
-                        }
-                        UpdateEntryPoint::Append { layer } => {
-                            graph.add_entry_point(plan.inserted_vector, layer).await?;
-                        }
-                    }
+            // Insert graph mutation into hawk_graph_mutations and link to modification
+            if let Some(ref modification_key) = mutation.modification_key {
+                if let Some(modification) = modifications.get_mut(modification_key) {
```
mcalancea, line 1732: Two nested `if let Some(...)` with no `else` branches: a `SingleHawkMutation` with no `modification_key`, or with a key that doesn't match the map, is silently discarded with no log or error. Either both branches need explicit handling (warn + skip, or error), or there should be a `debug_assert!` documenting that these cases can't happen.

iris-mpc/src/server/mod.rs
```diff
@@ -592,22 +588,20 @@ async fn load_database(
...
+            [GraphMem::new(), GraphMem::new()]
+        };
+        graph_loader.load_graphs_from_checkpoint(both_eyes);
```
mcalancea, line 603: The old `tests_db::test_graph_load` was removed in commit 7 along with `load_graph_store`. No replacement test covers the new checkpoint download → WAL replay → graph reconstruction. Given the production load path doesn't yet replay WAL (TODO at line 591), an end-to-end test would make the eventual wiring much safer.

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
```diff
@@ -289,6 +271,81 @@ impl<V: VectorStore> GraphPg<V> {
...
+            let backup_table = format!("\"{}\".{}_backup", schema, table_name);
+
+            // Drop main table if it exists
+            sqlx::query(&format!("DROP TABLE IF EXISTS {main_table} CASCADE"))
+                .execute(&self.pool)
+                .await?;
+
+            // Recreate main table from backup
+            sqlx::query(&format!(
+                "CREATE TABLE {main_table} AS TABLE {backup_table}"
```
mcalancea: **`CREATE TABLE main AS TABLE backup`** copies columns + data but **not** PKs, IDENTITY, NOT NULLs, indices, or triggers. After this restore: `hawk_graph_mutations.id` loses its IDENTITY (next insert fails on NOT NULL), `persistent_state` loses its `(domain, key)` PK (breaking the `ON CONFLICT` upsert in `set_last_indexed_modification_id`), and `genesis_graph_checkpoint` loses its constraints. Pre-existing pattern but new tables have constraints the old link tables didn't. Currently only reachable via `exec_use_backup_as_source` (`#[allow(dead_code)]`).

**Question**: do we still need backup/restore at all now that recovery goes through S3 checkpoint + WAL? If kept, fix with `TRUNCATE main; INSERT INTO main SELECT * FROM backup;` (preserves schema).

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
```diff
@@ -331,6 +388,26 @@ impl<V: VectorStore> GraphPg<V> {
         Ok(rows.into_iter().collect())
     }
 
+    pub async fn get_hawk_graph_mutations_after(
```
mcalancea, line 323: No unit tests for `get_hawk_graph_mutations_after` (especially `None` vs `Some(id)`-with-rows-above/below the threshold), `insert_hawk_graph_mutations` (round-trip + `RETURNING id`), `set_last_indexed_modification_id` (insert + upsert path), or `get_max_hawk_graph_mutation_id` (empty vs non-empty). These are now load-bearing for delta-replay.

iris-mpc-bins/bin/iris-mpc-cpu/graph_mem_cli.rs
```diff
@@ -52,50 +55,39 @@ enum Command {
...
         file,
         dbg,
+        s3_bucket,
+        party_id,
+        aws_region,
         command,
     } = Cli::parse();
-    let db_context = DbContext::new(&db_url, &schema).await;
+
+    let region_provider = S3Region::new(aws_region.unwrap_or_else(|| "eu-north-1".to_string()));
```
mcalancea: The doc comment on `aws_region` says 'defaults to env/instance metadata', but `.region(region_provider)` unconditionally overrides whatever `from_env()` would have chosen. So `AWS_REGION=us-east-1 graph_mem_cli ...` silently runs in `eu-north-1`. Fix: only call `.region(...)` when `aws_region.is_some()`.

iris-mpc-bins/bin/iris-mpc-cpu/init_test_dbs.rs
```diff
@@ -451,9 +454,25 @@ async fn main() -> Result<()> {
 }
 
 async fn init_dbs(args: &Args) -> Vec<DbContext> {
+    let region = S3Region::new(args.aws_region.clone());
```
mcalancea: Same env-region override pattern as `graph_mem_cli.rs:71` — `.region(region)` always wins over `from_env()`. Less misleading here because `--aws-region` has clap `default_value` (no `Option`/doc-comment mismatch), but `AWS_REGION` is still silently ignored. Same fix.

iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
```diff
@@ -1591,10 +1599,9 @@ mod tests {
 
         for (role, mpc_graph) in mpc_graphs.into_iter().enumerate() {
             let mpc_graph = mpc_graph.unwrap_or_else(|| panic!("party {role} did not finish"));
-            assert_eq!(
-                mpc_graph, plaintext_graph,
-                "MPC graph for party {role} differs from plaintext graph",
-            );
+            if let Err(e) = ensure_node_equivalence(&mpc_graph, &plaintext_graph) {
```
mcalancea, line 1606: `ensure_node_equivalence` checks layer count + per-layer node *set* only — does **not** compare neighbor list contents or order. The deleted `verify_graph_consistency` checked every node's exact neighbor list. Two graphs with identical node sets but different links would pass here. Suggest: add `assert_eq!(mpc_graph, plaintext_graph)` as a primary strict check, and use `run_diff(..., ExplicitNeighborhoodDiffer::new(SortBy::Index))` as the failure-message detail. Also: the panic lost the `[Role {role}]` prefix the old code carried — useful for 3-party CI failures.

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
```diff
@@ -358,502 +435,75 @@ impl<V: VectorStore> GraphPg<V> {
...
+        Ok(row.get("id"))
     }
 
-    /// Ensures that graph_entry and graph_links table are empty. For testing only
-    pub async fn clear_tables(&mut self) -> Result<()> {
-        let entry_table = self.entry_table();
-        let links_table = self.links_table();
+    /// Update the last_indexed_modification_id in persistent_state.
+    /// This tracks the highest modification_id that has been persisted to hawk_graph_mutations.
+    pub async fn set_last_indexed_modification_id(&mut self, modification_id: i64) -> Result<()> {
```
mcalancea, line 415: This writer (called from `HawkMutation::persist`) writes to `("hawk", "last_indexed_modification_id")`. The genesis path writes to `("genesis", ...)` via `state_accessor::set_last_indexed_modification_id` — and has readers. **Nothing reads the hawk-domain value.** Three options: (a) delete the writer until a real consumer exists, (b) point at the genesis domain so the two flows share a watermark, or (c) land the hawk-side consumer in this PR. Currently it's a write-only side effect on every persist.

iris-mpc-bins/bin/iris-mpc-cpu/db_sanity_check.rs
```diff
@@ -283,85 +282,101 @@ async fn main() -> Result<()> {
...
+        .get_hawk_graph_mutations_after(checkpoint_state.graph_mutation_id)
+        .await?;
+    rpt!(
+        rpt,
+        "  Applying {} mutation row(s) after checkpoint (graph_mutation_id={:?})...",
+        mutation_rows.len(),
+        checkpoint_state.graph_mutation_id,
+    );
+    for row in &mutation_rows {
+        let both_eyes: BothEyes<Vec<GraphMutation<VectorId>>> =
```
mcalancea: Same `bincode::deserialize::<BothEyes<Vec<GraphMutation<VectorId>>>>(&row.serialized_mutations)` pattern as `test_utils::get_both_eyes` (commit 12). When production WAL replay lands (server/mod.rs TODO), this will be three copies. Extract a method on `GraphMutationRow` (e.g. `fn deserialize_mutations(&self) -> Result<BothEyes<Vec<GraphMutation<VectorId>>>>`) and drop the new `bincode` dep from `iris-mpc-bins`.

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
mcalancea: `binary_search + insert` keeps `neighbor_links` sorted. The invariant: `set_links` must receive already-sorted neighbors. `insert_prepare_batch` does sort them (`searcher.rs:1611`), so this holds today — but a future caller bypassing the planner could break it silently (binary_search on an unsorted Vec returns wrong results, leading to wrong-position inserts and undetectable graph corruption). Add a `// invariant: neighbor_links is kept sorted; relies on the planner sorting before set_links` comment so the dependency is explicit.

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
mcalancea: I'll further note here that these backup methods weren't used during Genesis (the backups were manual, done by Wojciech). So I suspect they aren't used anywhere.

mcalancea (approved):

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
sdwoodbury: I was wary of mixing the state of hawk and genesis (rules out (b)). i expected the sidecar service to use this value in the future. the constants will have to be moved to a common location. 

since this is a feature branch where additional work is required before merging to main, is it really necessary to pick from (a) or  (c) ?

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
sdwoodbury: The below call chain happens once per batch. Given that M=320 and N=number of nodes in the graph (N >> M), the options are:
- Hashing per graph update ~ O(M^2) 
- Hashing per state_check() ~ O(N*M)  

looks like the current implementation wins

```                                                                                                                                                                          
  state_check()                                                                                                                                                                 
    └─ state_check_side()  [state_check.rs:397]                                                                                                                                 
         └─ session.checksum()  [state_check.rs:558]                                                                                                                            
              └─ graph_store.read().checksum()  [layered_graph.rs:295]
                   └─ for each layer:                                                                                                                                           
                        set_hash.add_unordered((lc, layer.set_hash.checksum()))                                                                                                 
```

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
sdwoodbury: good point. i'll make modification_id the primary key.

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
sdwoodbury: i'd like to get rid of the backup/restore

iris-mpc/src/server/mod.rs
sdwoodbury: given the size of the current pull request, i think this should be deferred to a follow up PR

iris-mpc-cpu/src/execution/hawk_main.rs
sdwoodbury: added logging for the inner `if let` and a comment for the outer one.

iris-mpc/src/server/mod.rs
sdwoodbury: we may want to add a cmdline arg. note that before the WAL, it was valid to start the server with an empty database and hence an empty graph. The current code matches the old behavior. I think this should at least be out of scope for this PR.

iris-mpc-cpu/src/hnsw/graph/mutation.rs
sdwoodbury: as far as i can tell, the only way to make the graph mutations not depend on an implementation detail which could possibly change later is to include the entire new  neighborhood of the node. the point of the WAL was to purposely stop doing that so that the DB doesn't get clobbered. 

to catch a regression, i an add some carefully crafted unit tests. another option is giving the GraphMutation enum new versions of the same operation, if we ever want to support multiple versions of the same operation. but that seems unlikely.

sdwoodbury: added unit tests, removed the version field from GraphMutation, and made a comment about the possibility of a versioned enum

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
mcalancea: No, that's fine, let's leave it at this. I was just confused by the intention because "indexed" was genesis-only vocabulary for me.

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
mcalancea: Ah sorry, I personally meant sorting the entire neighborhood and hashing the sorted version (with a single SipHash) regardless of the modifications themselves. Similar to what is done now. 

I thought Claude meant the same but I see now it didn't. O(N * M) makes no sense, of course.

iris-mpc-cpu/src/hnsw/searcher.rs
```diff
@@ -1565,24 +1545,40 @@ impl HnswSearcher {
...
+    /// an optional `RemoveNode` followed by an `InsertNode`.
     ///
-    /// After the `ConnectPlan` structs are created, the final state of all
-    /// updated neighborhoods is inspected to see if any are too large and need
-    /// to be compacted.  All such neighborhoods are compacted in one step at
-    /// the end of the function call.
+    /// For every `InsertNode` found, the function computes bilateral neighborhood
+    /// updates (intra-batch inserts are tracked so they see each other's links),
+    /// then checks whether any neighborhood has grown past the M limit.  If so,
+    /// a `Compact` mutation is appended to the `GroupedMutations` of whichever
```
bgillesp, line 1583: I'm concerned about the logic surrounding this.  Because of the way that neighborhood compaction is done, all of the compactions have to be computed at the end as a batch.  If the `Compact` mutations are appended to a `GroupedMutations` which caused the neighborhood to grow past the limit, then the removal of nodes may not reflect the state of the corresponding neighborhood when the compaction was calculated.  Example: node 1000 and 1001 are added to neighborhood 50, but the first one, node 1000, causes neighborhood 50 to surpass the compaction limit.  When the compaction is calculated with both nodes 1000 and 1001 included, 1001 is one of the edges removed by the compaction.  If this compaction op is applied in the group when the 1000 was inserted, then the 1001 is added in a later group, the final result is that compacted neighborhood 50 includes 1001 in the end, which it shouldn't.  This is the reason that compaction operations were appended to the last update in the previous version -- this is the point in the graph evolution that the compactions are actually calculated.

sdwoodbury: per today's conversation, this will be done in a follow up PR so that Sidecar work will be unblocked sooner.

PR #2119, opened 2026-05-08
Streaming S3 upload + download primitives for graph checkpoints

mcalancea:
## Summary

Adds two streaming counterparts to the existing buffered `upload_graph` / `download_graph` paths in `iris-mpc-cpu/src/graph_checkpoint/s3_client/`. Both avoid materializing the full serialized payload in memory — important for graphs that approach or exceed available RAM.

### Upload (`streaming.rs`)
- `stream_serialize_and_upload_with(s3, bucket, key, |w| ..., part_size, parallelism)` — serializer runs on `spawn_blocking` and writes into a bounded async duplex pipe; read side spawns up to `parallelism` concurrent `UploadPart` tasks via a `JoinSet`.
- Per-part retry (`UPLOAD_PART_MAX_RETRIES = 3`, 2s backoff) for transient S3 errors; fails fast on permanent ones.
- Aborts the multipart upload on any failure path (serializer error, S3 error, task panic).
- Peak memory outside the serializer's working set is `(parallelism + 1) * part_size` — at defaults (8 × 100 MiB) ≈ 900 MiB, vs. ~130 GiB for the buffered path at projected graph size.

### Download (`streaming_download.rs`)
- `stream_download_and_deserialize<T>(s3, bucket, key) -> (T, Blake3Hash)` — issues a single `GetObject`, tees the body through BLAKE3 into a bounded duplex pipe, deserializes on a `spawn_blocking` task via `SyncIoBridge`.
- Returns the deserialized value plus the BLAKE3 digest of the bytes, so callers verify against the recorded checkpoint hash in one pass.
- Wire-compatible with the existing buffered `download_and_hash` path's hex hash — a checkpoint written by either upload path is readable by either download path.
- Core helper `deserialize_and_hash_from<R: AsyncRead, T>(reader, capacity)` is generic so tests don't need S3.

## Test plan

- [x] `cargo test -p iris-mpc-cpu --lib -- graph_checkpoint::s3_client` — 9 unit tests pass without S3:
  - Upload: byte-equality vs. buffered `bincode`, back-pressure on a payload larger than `part_size`, serializer-error propagation, serializer-panic propagation.
  - Download: round-trip vs. buffered `bincode + blake3::hash`, back-pressure on a payload much larger than pipe capacity, truncated payload errors cleanly (no hang), reader error propagates, empty input fails cleanly.
- [x] `cargo clippy -p iris-mpc-cpu --lib --tests --no-deps` — clean.
- [ ] localstack-gated upload integration test (`#[ignore]` by default; runs in localstack-equipped CI).
- [ ] No equivalent download integration test in this PR — exercised by the upcoming `RebuildFromCheckpoint` materializer work.

### DIFF
--- a/iris-mpc-cpu/src/graph_checkpoint/s3_client/streaming.rs
+++ b/iris-mpc-cpu/src/graph_checkpoint/s3_client/streaming.rs
@@ -1,0 +1,390 @@
+//! Streaming serialize + S3 multipart upload.
+//!
+//! `stream_serialize_and_upload` serializes a value directly into an S3
+//! multipart upload without materializing the full byte buffer in memory.
+//! The serializer runs on a blocking thread and writes into a bounded async
+//! duplex pipe; the read side spawns up to `parallelism` concurrent
+//! `UploadPart` tasks as full chunks become available. Outside of the
+//! serializer's own working set, peak memory is roughly
+//! `(parallelism + 1) * part_size` plus the AWS SDK's request buffers.
+//!
+//! Compare to [`super::multipart::upload_graph`], which requires the caller
+//! to pass a fully buffered `Bytes` payload (doubling memory for large
+//! graphs).
+
+use std::{io::Write, sync::Arc};
+
+use aws_sdk_s3::{
+    types::{CompletedMultipartUpload, CompletedPart},
+    Client as S3Client,
+};
+use eyre::{eyre, Result};
+use tokio::{
+    io::{AsyncReadExt, DuplexStream},
+    sync::Semaphore,
+    task::JoinSet,
+};
+use tokio_util::io::SyncIoBridge;
+
+/// Default part size used by [`stream_serialize_and_upload`]. Also the size
+/// of the duplex pipe, which bounds back-pressure on the serializer.
+pub const DEFAULT_STREAMING_PART_SIZE: usize = 100 * 1024 * 1024;
+
+/// Default cap on concurrent in-flight `UploadPart` tasks.
+pub const DEFAULT_STREAMING_PARALLELISM: usize = 8;
+
+/// Serialize directly into an S3 multipart upload without buffering the full
+/// payload in memory.
+///
+/// `serialize` runs on a `spawn_blocking` thread, so it must be `Send +
+/// 'static`; callers typically move owned data or an `Arc`/owned guard into
+/// the closure. The closure receives a `&mut dyn Write` and should return
+/// `Ok(())` on success.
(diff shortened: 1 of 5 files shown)


iris-mpc-cpu/src/graph_checkpoint/s3_client/streaming.rs
```diff
@@ -0,0 +1,372 @@
...
+                }
+            }
+        }
+    }
+
+    if let Some(e) = first_error {
+        join_set.abort_all();
+        return Err(e);
+    }
+
```
sdwoodbury, line 360: `abort_all()` will have no effect after `join_next()` is exhausted.

iris-mpc-cpu/src/graph_checkpoint/s3_client/streaming.rs
```diff
@@ -0,0 +1,372 @@
...
+            let _permit = permit;
+            let res = client
+                .upload_part()
+                .bucket(&bucket)
+                .key(&key)
+                .upload_id(&upload_id)
+                .part_number(pn)
+                .body(buf.into())
+                .send()
+                .await
```
sdwoodbury: nit: it might be helpful to wrap this in a retry loop, so that the code sleeps and retries a few times before failing.

sdwoodbury (dismissed):

sdwoodbury:
we might want to merge this into the WAL feature branch instead. That would avoid the need to rebase the feature branch onto main, which would affect any work off of that branch.

mcalancea:
@sdwoodbury yep, that's the intention, I'll do it before asking for review formally.

sdwoodbury (approved):
This looks good.

I think this could still benefit from HTTP ranges. the logic of `multipart::download_graph` could be modified to write the `join_set` output to a stream rather than assembling `final_data`. For simplicity, A single range could be requested at a time. Although not the fastest, this would at least handle transient network errors without having to restart the download. And it would not require changes to `deserialize_and_hash_from()`

mcalancea:
> This looks good.
> 
> I think this could still benefit from HTTP ranges. the logic of `multipart::download_graph` could be modified to write the `join_set` output to a stream rather than assembling `final_data`. For simplicity, A single range could be requested at a time. Although not the fastest, this would at least handle transient network errors without having to restart the download. And it would not require changes to `deserialize_and_hash_from()`

Aye, good point 👍 , I included it.

sdwoodbury (approved):

PR #2124, opened 2026-05-13
Refactor GraphMutation for uniform edge ops and two-phase apply

bgillesp:
## Summary

Restructures the `GraphMutation` enum so node-level and edge-level operations are cleanly separated, edge ops are uniform with an explicit direction, and `GraphMem::insert_apply` is two-pass.

**Structural changes (in `iris-mpc-cpu/src/hnsw/graph/mutation.rs`):**
- New `EdgeDirection` enum with `Outgoing` / `Incoming` / `Bidirectional` variants. `Bidirectional` is stored as a single op (semantic intent), not split into primitives.
- `AddEdges` and `RemoveEdges` are now single-layer with an explicit `direction: EdgeDirection`. Both have mirrored shapes (`{ id, layer, to_add | to_remove, direction }`).
- `AddNode` reshaped from `{ id, layers: Vec<(usize, Vec<V>)>, update_ep }` to `{ id, max_graph_layer: usize, update_ep }` — purely node-level, no forward neighborhood data. Forward links move into `AddEdges` ops emitted alongside `AddNode` in the same group.

**Apply path (in `iris-mpc-cpu/src/hnsw/graph/layered_graph.rs`):**
- `GraphMem::insert_apply` is now two-pass — `AddNode` / `RemoveNode` first, then `AddEdges` / `RemoveEdges`. This eliminates a class of ordering bugs (an edge op listed before its node op in the input Vec now applies correctly).
- Direction dispatch in the edge arms with `warn!` paths for missing referents.
- Four `Layer` edge primitives are now uniform in signature (`(&mut self, id: &V, to_add|to_remove: Vec<V>)`) and style: `add_outgoing_edges`, `add_incoming_edges`, `remove_outgoing_edges`, `remove_incoming_edges`.

**Searcher (in `iris-mpc-cpu/src/hnsw/searcher.rs`):**
- `insert_prepare_batch` reads new-node neighborhoods from `AddEdges` ops in the same group (Outgoing or Bidirectional) instead of from `AddNode.layers`.
- All call sites of the old enum shape across `hawk_main.rs`, `insert.rs`, and the test_utils helpers migrated.

**`GroupedMutations` stays a dumb `Vec` wrapper** — no construction-time validation added. Deferred to the future modification-id design per the linked planning doc.

### DIFF
--- a/iris-mpc-cpu/src/hnsw/graph/test_utils.rs
+++ b/iris-mpc-cpu/src/hnsw/graph/test_utils.rs
@@ -18,8 +18,9 @@ use crate::{
                 node_equiv::ensure_node_equivalence,
                 run_diff,
             },
+            mutation::EdgeType,
             neighborhood::Neighborhood,
-            GraphMutation,
+            GraphMutation, UpdateEntryPoint,
         },
         vector_store::{VectorStore, VectorStoreMut},
         SortedNeighborhood,
@@ -375,7 +376,7 @@ impl DbContext {
         let ep_mutation = GraphMutation::AddNode {
             id: vectors[0],
             height: 1,
-            update_ep: crate::hnsw::graph::mutation::UpdateEntryPoint::SetUnique { layer: 0 },
+            update_ep: UpdateEntryPoint::SetUnique { layer: 0 },
         };
         left_graph.insert_apply(vec![ep_mutation]);
 
@@ -392,13 +393,13 @@ impl DbContext {
                 GraphMutation::AddNode {
                     id: vectors[i],
                     height: 1,
-                    update_ep: crate::hnsw::graph::mutation::UpdateEntryPoint::False,
+                    update_ep: UpdateEntryPoint::False,
                 },
                 GraphMutation::AddEdges {
-                    id: vectors[i],
+                    base: vectors[i],
                     layer: 0,
-                    to_add: neighbors,
-                    direction: crate::hnsw::graph::mutation::EdgeDirection::Outgoing,
+                    neighbors,
+                    edge_type: EdgeType::Base,
                 },
             ];
             left_graph.insert_apply(mutations);
--- a/iris-mpc-cpu/src/hnsw/graph/test_utils.rs
+++ b/iris-mpc-cpu/src/hnsw/graph/test_utils.rs
@@ -374,8 +374,8 @@ impl DbContext {
         // Set entry point via InsertNode with SetUnique
         let ep_mutation = GraphMutation::AddNode {
             id: vectors[0],
-            layers: vec![(0, vec![])],
-            update_ep: crate::hnsw::graph::mutation::UpdateEntryPoint::SetUnique { layer: 0 },
+            max_graph_layer: 0,
+            update_ep: crate::hnsw::graph::mutation::UpdateEntryPoint::SetUnique { layer: 1 },
         };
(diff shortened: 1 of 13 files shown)


iris-mpc-cpu/src/hnsw/graph/test_utils.rs
```diff
@@ -388,12 +388,20 @@ impl DbContext {
...
-                id: vectors[i],
-                layers: vec![(0, neighbors)],
-                update_ep: crate::hnsw::graph::mutation::UpdateEntryPoint::False,
-            };
-            left_graph.insert_apply(vec![mutation]);
+            let mutations = vec![
+                GraphMutation::AddNode {
+                    id: vectors[i],
+                    height: 1,
+                    update_ep: crate::hnsw::graph::mutation::UpdateEntryPoint::False,
```
sdwoodbury: nit: i know i did the same thing without realizing but including the entire path is a pet peeve of mine (i know it's the AI agent doing this)

sdwoodbury (commented):
I like the shape of `GraphMutation`- the `AddEdges` and `RemoveEdges` variants with a field that tells what type of operation to perform. I think the naming is confusing though. 

`to_add` isn't always "to add", sometimes  `id` is the one being added. Same with `to_remove`. 

and `EdgeDirection` is really describing the type of operation to perform. This is usually either to set the node's neighbors or add `id` to each of the nodes' neighborhoods. 

I have trouble visualizing direction in this scenario. I don't have a sense of where the origin is so Incoming and Outgoing feel nebulous. 

Consider renaming to the following:
```
pub enum AffectedEdges {
    Anchor,    // Update the anchor's list
    Neighbors, // Update the neighbors' lists
    All,       // Update both (Symmetric)
}
```

```
enum GraphMutation {
  AddEdges {
    anchor: ...,
    layer: ...,
    neighbors: ...,
    affected_edges: ...,
  },
  RemoveEdges {
    anchor: ...,
    layer: ...,
    neighbors: ...,
    affected_edges: ...,
  },
  ...
```

the naming isn't perfect but at least by having the field names in `GraphMutation` match the variants of the other enum, it's immediately clear which neighborhood is getting updated (the single neighborhood - of anchor, the neighborhoods of neighbors, or all).

bgillesp:
@sdwoodbury good points -- I renamed edge mutations as follows, let me know what you think:

```
AddEdges {
    base: Vector,
    neighbors: Vec<Vector>,
    layer: usize,
    edge_type: EdgeType,
},
RemoveEdges {
    base: Vector,
    neighbors: Vec<Vector>,
    layer: usize,
    edge_type: EdgeType,
},
```

```
/// Type of edges between `base` and the nodes listed in `neighbors` affected by
/// edge mutations.
///
/// - `Base`: affects nodes listed in `neighbors` found in `base`'s neighbor
///   list (forward edges from `base`).
/// - `Neighbors`: affects instances of `base` in each neighbor node's neighbor
///   list (back-edges into `base`).
/// - `All`: affects both of the above types of edges (symmetric edges).
#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
pub enum EdgeType {
    /// Affects forward edges from the base node
    Base,

    /// Affects back edges into the base node
    Neighbors,

    /// Affects both forward edges from and back edges into the base node
    All,
}
```

sdwoodbury (approved):
looks great!

PR #2132, opened 2026-05-15
[POP-3836] Apply WAL to graph checkpoint on startup

sdwoodbury:
note that `genesis` does not add to `hawk_graph_mutations` so there is no double-apply risk when `hawk` starts up with a checkpoint created by `genesis`

### DIFF
--- a/iris-mpc-common/src/helpers/sync.rs
+++ b/iris-mpc-common/src/helpers/sync.rs
@@ -13,6 +13,9 @@ pub struct SyncState {
     /// Bincode-serialized `BothEyes<Vec<GraphMutation<IrisVectorId>>>` for each
     /// modification in `modifications` (parallel by index).  `None` means this
     /// party has no WAL entry for that modification.
+    ///
+    /// using serde(default) for backwards compatibility.
+    #[serde(default)]
     pub graph_mutation_bytes: Vec<Option<Vec<u8>>>,
 }
 
--- a/iris-mpc-common/src/helpers/sync.rs
+++ b/iris-mpc-common/src/helpers/sync.rs
@@ -221,7 +221,7 @@ impl SyncResult {
             grouped.entry(m.id).or_default().push(m);
         }
 
-        tracing::info!("Grouped modifications: {:?}", grouped);
+        tracing::info!("Grouped modifications: {}", grouped.len());
 
         let completed_max_mod_ids: Vec<Option<i64>> = self
             .all_states
@@ -240,9 +240,9 @@ impl SyncResult {
             let mod_id_diff = max_id.saturating_sub(min_id) as usize;
             if mod_id_diff > self.my_state.common_config.get_max_modifications_lookback() {
                 panic!(
-                    "Modification ID difference across nodes is too large: {:?}. Min: {:?}, Max: {:?}. \
+                    "Modification ID difference across nodes is too large. Min: {:?}, Max: {:?}. \
              Can not safely handle this case, consider bumping lookback. Crashing!",
-                    completed_max_mod_ids, min_id, max_id
+                    min_id, max_id
                 );
             }
         }
--- a/iris-mpc-common/src/helpers/sync.rs
+++ b/iris-mpc-common/src/helpers/sync.rs
@@ -301,7 +301,7 @@ impl SyncResult {
                             roll_forward.status = ModificationStatus::Completed.to_string();
                             roll_forward.persisted = any_persisted;
                             tracing::warn!(
-                                "Updating modification row from {:?} to {:?}",
+                                "Planning to update modification row from {:?} to {:?}",
                                 local_m,
                                 roll_forward
                             );
--- a/iris-mpc-common/src/helpers/sync.rs
+++ b/iris-mpc-common/src/helpers/sync.rs
@@ -10,6 +10,10 @@ pub struct SyncState {
     pub modifications: Vec<Modification>,
     pub next_sns_sequence_num: Option<u128>,
     pub common_config: CommonConfig,
+    /// Bincode-serialized `BothEyes<Vec<GraphMutation<IrisVectorId>>>` for each
+    /// modification in `modifications` (parallel by index).  `None` means this
+    /// party has no WAL entry for that modification.
+    pub graph_mutation_bytes: Vec<Option<Vec<u8>>>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -384,6 +388,7 @@ mod tests {
             modifications,
             next_sns_sequence_num: None,
             common_config: CommonConfig::from(config),
+            graph_mutation_bytes: vec![],
         }
     }
 
@@ -790,18 +795,21 @@ mod tests {
                 modifications: vec![],
                 next_sns_sequence_num: Some(100),
                 common_config: CommonConfig::default(),
+                graph_mutation_bytes: vec![],
             },
             SyncState {
                 db_len: 20,
                 modifications: vec![],
                 next_sns_sequence_num: Some(200),
                 common_config: CommonConfig::default(),
+                graph_mutation_bytes: vec![],
             },
             SyncState {
                 db_len: 30,
                 modifications: vec![],
                 next_sns_sequence_num: Some(150),
                 common_config: CommonConfig::default(),
+                graph_mutation_bytes: vec![],
             },
         ];
 
@@ -814,6 +822,7 @@ mod tests {
             modifications: vec![],
             next_sns_sequence_num: None,
             common_config: CommonConfig::default(),
+            graph_mutation_bytes: vec![],
         };
         let all_states = vec![
             state_with_none_sequence_num.clone(),
@@ -836,18 +845,21 @@ mod tests {
                 modifications: vec![],
                 next_sns_sequence_num: None, // NodeX - advanced but empty queue
                 common_config: CommonConfig::default(),
+                graph_mutation_bytes: vec![],
             },
             SyncState {
                 db_len: 20,
                 modifications: vec![],
                 next_sns_sequence_num: Some(123), // Other nodes still have messages
                 common_config: CommonConfig::default(),
+                graph_mutation_bytes: vec![],
             },
             SyncState {
                 db_len: 30,
                 modifications: vec![],
                 next_sns_sequence_num: Some(123),
                 common_config: CommonConfig::default(),
+                graph_mutation_bytes: vec![],
             },
         ];
 
@@ -967,18 +979,21 @@ mod tests {
                 modifications: vec![],
                 next_sns_sequence_num: Some(100),
                 common_config: CommonConfig::from(config1),
+                graph_mutation_bytes: vec![],
             },
             SyncState {
                 db_len: 20,
                 modifications: vec![],
                 next_sns_sequence_num: Some(100),
                 common_config: CommonConfig::from(config2),
+                graph_mutation_bytes: vec![],
             },
             SyncState {
                 db_len: 20,
                 modifications: vec![],
                 next_sns_sequence_num: Some(100),
                 common_config: CommonConfig::from(config3),
+                graph_mutation_bytes: vec![],
             },
         ];
 
--- a/iris-mpc-common/src/helpers/sync.rs
+++ b/iris-mpc-common/src/helpers/sync.rs
@@ -240,9 +240,9 @@ impl SyncResult {
             let mod_id_diff = max_id.saturating_sub(min_id) as usize;
             if mod_id_diff > self.my_state.common_config.get_max_modifications_lookback() {
                 panic!(
-                    "Modification ID difference across nodes is too large. Min: {:?}, Max: {:?}. \
+                    "Modification ID difference across nodes is too large: {:?}. Min: {:?}, Max: {:?}. \
              Can not safely handle this case, consider bumping lookback. Crashing!",
-                    min_id, max_id
+                    completed_max_mod_ids, min_id, max_id
                 );
             }
         }
--- a/iris-mpc-cpu/src/graph_checkpoint/synchronize.rs
+++ b/iris-mpc-cpu/src/graph_checkpoint/synchronize.rs
@@ -1,9 +1,11 @@
 use std::collections::HashMap;
 
+use aws_sdk_s3::Client;
 use eyre::{bail, eyre, Result};
 use iris_mpc_common::{helpers::sync::SyncResult, IrisVectorId};
 use itertools::izip;
 
+use super::{download_graph_checkpoint, GraphCheckpointState};
 use crate::{
     execution::hawk_main::{BothEyes, HawkOps},
     hawkers::aby3::aby3_store::Aby3Store,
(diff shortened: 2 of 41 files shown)


iris-mpc-cpu/src/graph_checkpoint/synchronize.rs
```diff
@@ -0,0 +1,109 @@
...
+                .and_modify(|e| {
+                    // if one party has not written the graph mutation yet (e would be None) but the other had,
+                    // then update the graph mutation
+                    if e.is_none() && graph_mutation.is_some() {
+                        *e = graph_mutation;
+                    }
+                })
+                .or_insert(graph_mutation);
+            if entry.is_some() && graph_mutation.is_some() && *entry != graph_mutation {
+                bail!("graph mutation mismatch between parties. modification id: {}; party A: {:?}; party B: {:?}", modification.id, entry, graph_mutation);
```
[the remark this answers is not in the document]
sdwoodbury: this is incorrect. it should absolutely fail.

migrations/20260425000002_hawk_graph_mutations_v2.up.sql
```diff
@@ -15,3 +15,12 @@ ALTER TABLE modifications
...
     DROP COLUMN IF EXISTS mutation_version;
+
+ALTER TABLE hawk_graph_mutations
+    DROP CONSTRAINT IF EXISTS hawk_graph_mutations_pkey;
+
+ALTER TABLE hawk_graph_mutations
+    DROP COLUMN IF EXISTS id;
+
+ALTER TABLE hawk_graph_mutations
+    ADD PRIMARY KEY (modification_id);
```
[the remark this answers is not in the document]
sdwoodbury: this is in a feature branch which has not been deployed yet.

iris-mpc-cpu/src/hnsw/graph/graph_store.rs
```diff
@@ -290,6 +290,8 @@ impl<V: VectorStore> GraphPg<V> {
             r#"
             INSERT INTO hawk_graph_mutations (modification_id, serialized_mutations)
             VALUES ($1, $2)
+            ON CONFLICT (modification_id) DO UPDATE
```
mcalancea, line 293: Both `insert_hawk_graph_mutations` methods (here and at `:436`) are now upserts post-PR, but the name still reads as a plain insert and there's no docstring saying otherwise. Suggest renaming to `upsert_hawk_graph_mutations`.

iris-mpc-common/src/helpers/sync.rs
```diff
@@ -233,9 +240,9 @@ impl SyncResult {
             let mod_id_diff = max_id.saturating_sub(min_id) as usize;
             if mod_id_diff > self.my_state.common_config.get_max_modifications_lookback() {
                 panic!(
-                    "Modification ID difference across nodes is too large: {:?}. Min: {:?}, Max: {:?}. \
+                    "Modification ID difference across nodes is too large. Min: {:?}, Max: {:?}. \
              Can not safely handle this case, consider bumping lookback. Crashing!",
-                    completed_max_mod_ids, min_id, max_id
+                    min_id, max_id
                 );
```
mcalancea, line 246: This panic lost the `completed_max_mod_ids` debug-dump that the original carried, which is the only thing telling you *which* parties are at the spread extremes when this fires. The trim is incidental to the PR's scope and degrades diagnostics on a code path that only runs when something is already wrong cross-party — suggest restoring the `{:?}` unless there's a specific reason for the change.

iris-mpc-cpu/src/graph_checkpoint/synchronize.rs
```diff
@@ -0,0 +1,287 @@
...
+/// `sync_modifications()` has its graph mutations present in the local
+/// `hawk_graph_mutations` table.
+///
+/// It is assumed that one party could have None for a graph mutation and
+/// another party could have Some for a graph mutation. for every modification
+/// in to_update, the peer will search SyncResult.all_states for a Some(graph_mutation)
+/// to roll forward.
+///
+/// The graph mutations are applied all together in a transaction.
+pub async fn sync_graph_mutations(
```
mcalancea, line 34: This function has the most novel logic in the PR (peer-merge semantics, None→Some upgrade, Some/Some mismatch, `to_update` filtering) and zero unit-test coverage — the test module below only exercises `apply_graph_mutations`. SyncResult-fixture tests for None/None/Some, Some/Some-match, Some/Some-mismatch-bail, empty `to_update`, and all-None cases would run ~80 lines and pin the cross-party semantics in CI.

iris-mpc-cpu/src/graph_checkpoint/synchronize.rs
```diff
@@ -0,0 +1,287 @@
...
+                    }
+                })
+                .or_insert(graph_mutation);
+            if entry.is_some() && graph_mutation.is_some() && *entry != graph_mutation {
+                bail!(
+                    "graph mutation mismatch between parties. modification id: {}",
+                    modification.id
+                );
+            }
+        }
```
mcalancea: The `and_modify`/`or_insert` + post-hoc `entry.is_some() && graph_mutation.is_some() && *entry != graph_mutation` check is correct but takes three trace-throughs to convince yourself of. An explicit `match (existing, incoming) { ... }` (None→insert, Some(None)+Some→upgrade, Some(Some(a))+Some(b) if a==b→noop, else→bail) reads more directly and pairs 1:1 with the unit-test cases in the previous comment.

iris-mpc-cpu/src/graph_checkpoint/synchronize.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    Ok(both_eyes)
+}
+
+/// Deserializes and applies a sequence of WAL mutation rows to an in-memory
+/// graph, mutating `both_eyes` in place.
+///
+/// Rows must be ordered by `modification_id` ascending (which is the ordering
+/// guaranteed by all `get_hawk_graph_mutations_*` queries).
+pub fn apply_graph_mutations(
```
mcalancea, line 168: Doc comments at `:107` and `:137-138` state the 'rows must be ascending by `modification_id`' invariant; today it's enforced only by the SQL `ORDER BY` at the call site. Add a `debug_assert!(mutation_rows.windows(2).all(|w| w[0].modification_id < w[1].modification_id))` so any future non-DB caller (test fixture, sidecar reader, replay tool) can't silently violate it.

iris-mpc-cpu/src/graph_checkpoint/synchronize.rs
```diff
@@ -0,0 +1,287 @@
...
+///
+/// It is assumed that one party could have None for a graph mutation and
+/// another party could have Some for a graph mutation. for every modification
+/// in to_update, the peer will search SyncResult.all_states for a Some(graph_mutation)
+/// to roll forward.
+///
+/// The graph mutations are applied all together in a transaction.
+pub async fn sync_graph_mutations(
+    sync_result: &SyncResult,
+    graph_pg: &GraphPg<Aby3Store<HawkOps>>,
```
mcalancea, line 36: This function never touches the deserialized form of the mutation bytes — it just shuffles them between peers via `graph_pg.begin_tx()` and `graph_pg.insert_hawk_graph_mutations()`, both generic over `V: VectorStore`. The concrete `Aby3Store<HawkOps>` binding is unnecessary; making it generic is a one-line change that unblocks reuse from plaintext/genesis paths. `apply_graph_mutations` and `load_graph_and_roll_forward` legitimately stay concrete to `IrisVectorId`.

iris-mpc/src/server/mod.rs
```diff
@@ -578,21 +602,37 @@ async fn init_hawk_actor(
...
+                    None => tracing::info!(
+                    "checkpoint predates WAL tracking; replaying entire WAL on top of checkpoint"
+                ),
+                }
+                graph_store.get_hawk_graph_mutations_after(after).await?
+            }
         };
-        Ok::<_, eyre::Report>(both_eyes)
+
+        load_graph_and_roll_forward(s3_client, checkpoint_bucket, checkpoint, wal_mutation_rows)
```
mcalancea, line 636: The 'pre-WAL checkpoint → replay entire WAL' branch (case where `graph_mutation_id == None`) assumes the WAL contains every modification since the checkpoint, which only holds in the immediate-upgrade window before any later checkpoint trims the WAL. If you ever end up with a pre-WAL checkpoint *and* a partially-trimmed WAL, this silently misses the gap. 

This case should just panic maybe?

iris-mpc/src/server/mod.rs
```diff
@@ -171,6 +170,9 @@ pub async fn server_main(config: Config) -> Result<()> {
             sync_result.clone(),
         )
         .await?;
+        // insert the WAL entries so that init_hawk_actor rolls the graph forward to
+        // the correct height.
+        sync_graph_mutations(&sync_result, &graph_store).await?;
```
mcalancea, line 175: WAL sync only runs inside the `enable_modifications_sync` block — with that flag off, each party replays only its local WAL, with no peer reconciliation. This should panic after the first state check kicks in, but I wonder whether this flag should exist at all? When would we want to not sync?

iris-mpc-cpu/src/graph_checkpoint/synchronize.rs
```diff
@@ -0,0 +1,287 @@
...
+) -> Result<BothEyes<GraphMem<IrisVectorId>>> {
+    let mut both_eyes = if let Some(state) = checkpoint {
+        tracing::info!(
+            "Loading graph from common S3 checkpoint, hash: {}",
+            state.blake3_hash
+        );
+        download_graph_checkpoint(s3_client, checkpoint_bucket, &state).await?
+    } else {
+        tracing::info!("No S3 checkpoint found, defaulting to empty graph");
+        [GraphMem::new(), GraphMem::new()]
```
mcalancea, line 151: Silent default to empty graph here also covers the 'no common checkpoint across parties' case (when `get_common_checkpoint` returned `None`) — which in a steady-state cluster is alarming (corruption, S3 ACL drift, mismatched hashes), not a benign 'fresh start.' 

Here I would make a more general suggestion of having asserts in place to ensure that initialization follows the branches we _expect_. I.e for such a "default to empty graph" decision we could have a required "hnsw-graph-is-empty=true" flag at the top level. Applies to other situations. Point is to prevent further state corruption when something unexpected happens during init. @bgillesp what do you think about this one as a general guideline?

iris-mpc-upgrade-hawk/tests/e2e_hawk.rs
```diff
@@ -0,0 +1,387 @@
...
+use crate::utils::{
+    genesis_runner,
+    irises::{self},
+    modifications::{ModificationInput, ModificationType},
+    mpc_node::{db_ops, MpcNodes},
+};
+
+#[test]
+#[serial]
+#[ignore = "requires external setup"]
```
mcalancea, line 68: CI does run `--include-ignored` in three scopes — `run-unit-tests.yaml:87` (`-p iris-mpc-cpu -- e2e`), `scripts/run-sanity-checks-genesis-e2e.sh` (`--test e2e_genesis`), `scripts/run-tests-hnsw-e2e.sh` (same) — but this test falls outside all three (wrong crate, wrong test binary). Add a `scripts/run-tests-hawk-e2e.sh` modelled on the genesis one + a CI workflow entry, otherwise the PR's most novel logic has no automated coverage despite the test infrastructure existing.

iris-mpc-upgrade-hawk/tests/e2e_hawk.rs
```diff
@@ -0,0 +1,387 @@
...
+                bail!("A server task exited unexpectedly: {:?}", exit_res);
+            }
+        }
+
+        // stop zombie threads
+        notify.notify_waiters();
+        // After synchronization, all parties should have 10 mutations (the union of all)
+        assert_hawk_mutations_len(&[10, 10, 10], &configs, line!()).await?;
+        // Assert all 3 parties have the same entries in their graph mutation databases
+        assert_hawk_mutations_equal(&configs).await?;
```
mcalancea, line 309: Only the asymmetric-backfill happy path is covered. At minimum add a mismatch-case test (two parties have different `Some` bytes for the same `modification_id` → `sync_graph_mutations` must `bail!` at `synchronize.rs:61-64`) before merge — it's the one *new* failure mode this PR introduces and the error message would benefit from being pinned by a test. The checkpoint+WAL-replay end-to-end and pre-WAL-checkpoint branches are also untested but can be follow-ups.

mcalancea (commented):
Nice, thanks! Nothing is a hard blocker, but the testing and CI items are the ones I'd most want addressed before merge.

Something not specific to this PR (probably should've flagged earlier):
 - We should have some "golden-file" wire-format tests for `GraphMutation`'s bincode encoding. There's a comment saying the enum should only be extended  in the final variant, but hard-coded tests would ensure this is not forgotten (and that we catch more subtle breaking of this law, like removing/editing existing variants, etc.)

iris-mpc-cpu/src/graph_checkpoint/synchronize.rs
sdwoodbury: plaintext/genesis doesn't need this function and has a different sync mechanism. using a generic is also unnecessary and in theory makes the build take longer.

iris-mpc/src/server/mod.rs
sdwoodbury: if a checkpoint with `None` for `graph_mutation_id` is not allowed then we can't use a graph built by genesis since `upload_and_sync_genesis_checkpoint()` intentionally passes `None`

sdwoodbury: added commit `4d4c74b1a1214abc606de95310c6f71413724ce3` to cover the case where there is no checkpoint but a non-empty WAL

iris-mpc/src/server/mod.rs
sdwoodbury: good question

iris-mpc-cpu/src/graph_checkpoint/synchronize.rs
sdwoodbury: perhaps `get_common_checkpoint` should only return `None` if all parties have no checkpoint.  otherwise, return an error. That would cover the error case raised here. 

as for a special flag to allow the default to empty graph, that seems like a big hassle unless it defaults to true.

sdwoodbury: added commit `26d7b9521589a9260c87bc548dde2e7423c323d1` to ensure that if at least one party has a checkpoint, and agreement isn't reached, the parties throw an error rather than start with an empty graph.

iris-mpc-upgrade-hawk/tests/e2e_hawk.rs
sdwoodbury: added as of commit `035eb4f140fc4aa1534e5d35852733c2f5160cf7`

mcalancea (approved):
Thank you for addressing the comments 🫡 ! Approved.

iris-mpc/src/server/mod.rs
sdwoodbury: per our conversation with TFH, this configuration option will be kept.

PR #2134, opened 2026-05-15
Adjust ordering of deletion request application to HNSW graphs

bgillesp:
Previously, deletions would be applied to graph data after all other graph updates were applied.  Since WAL entries are organized according to their index within a batch, this means that deletion mutations were applied in a different order to the in-memory graph than they would be upon replay from the WAL.  This PR restructures the functioning of `HawkHandle::handle_mutations` to bring deletion processing inside of the main `insert::insert` functionality.


sdwoodbury (approved):

mcalancea (approved):
Looks good!

PR #2135, opened 2026-05-16
Per-graph modification ids for HNSW mutations

bgillesp:
## Summary

Adds a per-graph monotonic `u64` sequence number to each `GraphMutation` group and to `GraphMem` itself. `GraphMem::insert_apply` becomes fallible with a strict-increase check, rejecting any mutation whose sequence number is not strictly greater than the current `last_update_seq_no`. Stamps sequence numbers at construction time in `insert::insert` via a small `UpdateSeqNoAllocator`. Also fixes a pre-existing write/read shape asymmetry in `hawk_graph_mutations`.

Foundation for the eventual stale-edge / per-edge-timestamp work; immediate value is the deterministic-replay safeguard.

## Notable changes

- `GraphMutation<V>` is now a record struct `{ seq_no: u64, ops: Vec<MutationOp<V>> }` (was a newtype around `Vec<MutationOp<V>>`).
- `GraphMem` gains `pub last_update_seq_no: u64` and a non-mutating `next_sequence_number(&self) -> u64` peek.
- `insert_apply(&mut self, &GraphMutation<V>) -> Result<()>` validates strict-increase and advances the counter on success; `insert_apply_all` is a slice helper that short-circuits.
- `SingleHawkMutation.plans: BothEyes<Vec<ConnectPlan>>` (was `BothEyes<Option<ConnectPlan>>`). Empty Vec means "no change on that side"; multi-group capacity is forward-compatible with the deferred per-slot-splits refactor.
- WAL replay loops in `test_utils.rs` and `db_sanity_check.rs` propagate `insert_apply` errors per group.
- GraphV0/V1/V2/V3 → `GraphMem` conversions set `last_update_seq_no: 0`; `GraphMem → GraphV3` silently drops the field.

## Hard cutover

- Existing `hawk_graph_mutations` rows do not deserialize under the new shape (the previous Option-wrapped write was already inconsistent with the read path).
- Existing graph checkpoints do not deserialize (no `#[serde(default)]` on the new field).
- Round-trips through `GraphV3` reset `last_update_seq_no` to 0 — the safeguard is reliable only within a single process lifetime until `GraphV4` lands. WAL has not run in production yet, so the cross-restart limitation is acceptable.

## Out of scope (follow-ups)

- `GraphV4` introduction and proper upgrade path.
- `insert::insert` / `insert_prepare_batch` refactor, after which per-slot splits land (separate \`RemoveNode\` for replacements, separate compaction occurrences).
- "Compaction at end of batch" rework.
- Per-edge timestamps and stale-edge identification.


iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -64,6 +64,10 @@ pub struct GraphMem<V: Ref + Display + FromStr + Ord> {
     /// subset of the nodes of the previous layer, and graph neighborhoods in
     /// each layer represent approximate nearest neighbors within that layer.
     pub layers: Vec<Layer<V>>,
+
+    /// The id of the most recently applied `GraphMutation`. `0` means no
+    /// mutation has been applied. Advanced by `insert_apply` on success.
+    pub last_modification_id: u64,
```
sdwoodbury: nit: i think this name conflicts with the `Modification` struct and could lead to confusion. `last_mutation_id` might help since there isn't yet anything that references mutation ids (`hawk_graph_mutations` has a relation with `modification_id` so no conflict there). this could still be confusing but less so than `last_modification_id`.

sdwoodbury (approved):

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
bgillesp: Yes, definitely makes sense.  Am looking at changing the nomenclature to "sequence number" for the graph updates/mutations counter values.

sdwoodbury (approved):

PR #2136, opened 2026-05-18
WAL-based graph checkpoint protocol (sidecar + restart)

mcalancea:
## Summary
- Adds a unified checkpoint protocol that periodically (sidecar daemon) or once-at-startup (Hawk restart) materializes the graph from a checkpoint everyone agrees on, deterministically replays WAL rows from `hawk_graph_mutations` up to an agreed freeze height, hash-agrees across parties, and either uploads a new checkpoint (`UploadAndRecord`) or swaps the graph into the live in-Hawk reference (`InstallAsServing`).
- New module `iris-mpc-cpu/src/checkpoint_protocol/` with six traits (`Materializer`, `TerminalAction`, `ConsensusTransport`, `MutationStore`, `GraphHasher`, `BaseSelector`) orchestrated by `run_cycle`'s five phases (base → height → materialize → hash → hash-consensus).
- New `wal-sidecar` binary at `iris-mpc-bins/bin/iris-mpc-cpu/sidecar.rs` driving the daemon loop over a `ControlChannel`-backed `RingConsensusTransport`.

Targets `feat/wal_feature_branch2` (depends on the WAL machinery from #2118 and the ampc-common `ControlChannel` bump from #2129, both landed there).

## Snapshot consensus
Phase 1 doesn't assume "latest is the right one". Each party advertises its recent-checkpoints window (10 deep, matching genesis's `find_common_checkpoint`); a pluggable `BaseSelector` picks the agreed base:

- **`StrictLatest`** — original strict-newest semantics. Any divergence → `Skipped(NoCommonBase)`.
- **`MostRecentCommon`** (default for sidecar + restart) — newest-first walk for a row present in every peer's list. Tolerates one party missing the latest row (e.g. crash between hash-consensus and the local DB insert) by falling back to the previous common ancestor.

Disagreement is now caller-retryable (`Skipped(NoCommonBase)`), not Fatal.

## Test plan
- [x] 39 unit tests pass: `cargo test -p iris-mpc-cpu --lib checkpoint_protocol` (hasher, selector, transport, materializer, terminal, store, mod, plus three-party tests over an in-memory ring).
- [x] Workspace builds + binary compiles: `cargo check --workspace --tests`.
- [x] Clippy clean: `cargo clippy --workspace --tests -- -D warnings`.
- [x] fmt clean.
- [ ] Integration test against localstack S3 — deferred (no existing localstack scaffolding in `iris-mpc-cpu`).
- [ ] End-to-end Hawk restart against a running cluster — deferred. `restart_from_checkpoint` is exposed as a library helper; the hawk_main startup wire-up is a follow-up.
- [ ] Sidecar binary deployment against the dev cluster — deferred. The binary compiles and has unit-level coverage of its components; no real-cluster runs yet.

## Design notes for review
- **`UploadAndRecord::finalize` hash-equality check.** Compares the consensus hash (`Blake3GraphHasher`) against the upload's tee'd BLAKE3 and Fatals on mismatch — defensive ratchet against future serializer drift between the two paths. Pinned at the byte level by `terminal::tests::consensus_hash_matches_upload_path_serialization`.
- **`PeerBehindBase` gate.** `freeze < base.graph_mutation_id` (peer hasn't ingested up to the base WAL position) surfaces as `Outcome::Skipped(PeerBehindBase)` and fires BEFORE materializer/finalizer — avoids regressing the WAL anchor on the next checkpoint.
- **Phase 1 nonce.** Uses a fixed `BASE_PHASE_NONCE` sentinel; the agreed-base-derived nonce can't be used yet since base agreement is what Phase 1 is doing. Phases 2-5 use `agreed_base.checkpoint_id` and retain full cross-cycle crossed-wire detection. Stale Phase 1 replies (from a previous cycle) can leak in but either contain a valid common ancestor or get skipped by the selector.
- **`last_indexed_iris_id`** in new checkpoint rows is carried forward from `base.last_indexed_iris_id`. Hawk Main has no per-mutation iris-id tracking; the field is observability-only on this path.
- **Test-seam dependency.** `GraphHasher` is a trait with a single production impl (`Blake3GraphHasher`). The trait exists so protocol tests can inject `ConstHasher` to drive specific hash-agreement / hash-disagreement scenarios without engineering real graph state — without the trait, those tests get significantly harder.


iris-mpc-cpu/src/checkpoint_protocol/mod.rs
```diff
@@ -0,0 +1,324 @@
+//! Unified checkpoint protocol.
+//!
+//! Shared by the sidecar binary, an in-Hawk background checkpointing task,
+//! and the Hawk startup/restart path. The protocol orchestrates five
+//! phases (base agreement → height agreement → materialize → hash → hash
+//! consensus) over pluggable traits. Daemon loops, sleep cadence, and
+//! process lifecycle live outside this module in each caller's binary.
+//!
+//! See `Checkpoint Protocol - Unified Design.md` for the design rationale.
```
sdwoodbury: this document isn't available. probably needn't mention it

iris-mpc-cpu/src/checkpoint_protocol/materializer.rs
```diff
@@ -0,0 +1,204 @@
...
+            )));
+        }
+
+        // Phase B: replay WAL rows in `(base.graph_mutation_id, freeze]`.
+        let lo = base.graph_mutation_id.unwrap_or(0);
+        let hi = freeze.0;
+        if hi > lo {
+            let stream = MutationStore::mutations_in_range(self.graph_store, lo, hi).await?;
+            apply_wal_stream(&mut graph, stream).await?;
+        }
```
sdwoodbury: should probably return an error if hi <= lo

sdwoodbury (commented):
this code is too good. the abstractions are too clean and everything is too well organized. 

I also appreciate that the abstractions allow for testing at the unit test level rather than relying on integration tests.

sdwoodbury (approved):

iris-mpc-cpu/src/checkpoint_protocol/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+            BASE_PHASE_NONCE,
+            cfg.peer_round_timeout,
+        )
+        .await?;
+    let agreed_base = match selector.pick(&my_recent, &peer_lists) {
+        Some(b) => b,
+        None => return Ok(Outcome::Skipped(SkipReason::NoCommonBase)),
+    };
+    // Phases 2-5 share a nonce derived from the agreed base — preserves the
+    // crossed-wire detection the original strict-equality design had.
```
bgillesp, line 240: Comment references a design that isn't available now.

iris-mpc-cpu/src/checkpoint_protocol/materializer.rs
```diff
@@ -0,0 +1,206 @@
...
+        &mut self,
+        base: CheckpointMeta,
+        freeze: FreezeHeight,
+    ) -> Result<Graph, CycleError> {
+        if base.graph_version != GraphFormat::Current.version() {
+            return Err(CycleError::Fatal(format!(
+                "unsupported checkpoint graph_version={} for {}/{}",
+                base.graph_version, self.bucket, base.s3_key,
+            )));
+        }
```
bgillesp, line 43: This means that the current implementation doesn't support upgrading serialized graph formats, correct?  We will need to do an upgrade from V3 to V4 or V5 for graphs that were already materialized in production genesis -- do we have a plan for how to handle this upgrade path?

bgillesp (approved):
PR is looking great overall!  A couple of high-level things to consider for after this is merged:

- As discussed previously, I think that `graph_mutation_id` may now be redundant with `last_indexed_modification_id` in structs and `genesis_graph_checkpoint` Postgres table -- common value represents "any WAL entries up to the following modification id have been applied, and only WAL entries following this id are relevant for future evolution".  (Distinction no longer needed because WAL uses modification table ids for pkey and sequencing.)  We should check this and simplify the schema/structs if true, to avoid confusion.

- We should probably think about snapshot retention and pruning -- at 90GB or more per graph snapshot, ongoing snapshots 6x per day add up quickly for S3 usage.  If we use an S3 retention policy for ongoing snapshot pruning, then the checkpoints Postgres table gets out of date, and also if the retention policy is purely time-based then there would be a risk of eliminating all snapshots if there is ever extended downtime for the sidecar, which would require a full-graph rebuild since these are the ground truth for graph state.  My preference would be to have some retention and pruning procedure in the sidecar -- perhaps something of the flavor "keep all of most recent 10 snapshots, make archival snapshot every 120 snapshots, and keep last 30 archival snapshots, and prune Postgres records to represent actually available snapshots".

- As noted in comment, it looks like the implementation doesn't currently support upgrades from older serialized graph checkpoint versions.  I believe this means that there's no way to bring the v3 graphs we have from genesis up to the v4 format we'll be merging in with the WAL feature branch.  Do we have a plan for how to handle this upgrade path?

I think all of these can be addressed in follow-up PRs -- this one looks very tidy. 🙏

iris-mpc-cpu/src/checkpoint_protocol/materializer.rs
sdwoodbury: in `graph_checkpoint/s3_client/mod.rs` `download_graph_checkpoint` calls `read_graph_pair` with a `GraphFormat` and handles the upgrade to the current format.

sdwoodbury:
>     * As discussed previously, I think that `graph_mutation_id` may now be redundant with `last_indexed_modification_id` in structs and `genesis_graph_checkpoint` Postgres table -- common value represents "any WAL entries up to the following modification id have been applied, and only WAL entries following this id are relevant for future evolution".  (Distinction no longer needed because WAL uses modification table ids for pkey and sequencing.)  We should check this and simplify the schema/structs if true, to avoid confusion.

@bgillesp I double checked this and if i remember correctly, `graph_mutation_id` was for checkpoints created by `iris-mpc-hawk` while `last_indexed_modification_id` was for checkpoints created by genesis. currently `graph_mutation_id` is not set anywhere yet but to ensure the ids are consistent, i had resolved to re-use the modification id because that field has a trigger to ensure the ids remain consistent across parties. the naming is confusing though.

if `iris-mpc-hawk` used `last_indexed_modification_id` then if we wanted to run genesis again we couldn't since that field would effectively be corrupted.

PR #2141, opened 2026-05-20
[POP-3877] Add converters for GraphV4

sdwoodbury:
GraphMem has been updated and the conversion functions now need to account for the extra `seq_no` field. 

Also the `download_graph_checkpoint` function now uses the `read_graph_pair` function to process checkpoints which are older than `GraphFormat::Current`.


bgillesp (approved):
Good to go!  I made one commit to remove the old GraphMem -> GraphV3 conversion since all new serialization should go through the most compatible type.  Otherwise looking good, thanks!

PR #2142, opened 2026-05-21
Refactor insert_prepare_batch into sequential apply + post-batch helpers

bgillesp:
Replaces `HnswSearcher::insert_prepare_batch` with a simpler pipeline:

- `insert::insert` now applies graph mutations inline as it iterates batch slots, runs a per-slot invalid-link prune over neighborhoods that mutation touched, and finishes with a single global compaction attributed to the last non-empty slot. The complex hand-tracked attribution logic (`query_idxs`, `nbhd_last_group`, `insert_to_group`) is gone.
- Two new `GraphMutation` accessors enumerate the candidate sets: `expanded_neighborhoods` (AddEdges only — compaction candidates) and `updated_neighborhoods` (AddEdges + RemoveEdges — prune candidates).
- Two new `HnswSearcher` helpers do the MPC-aware work: `compact_batch` (calls `store.compact_neighborhood_batch` on neighborhoods exceeding `M_limit`) and `prune_invalid_links` (filters stale references via `store.only_valid_vectors`).
- Reauth/identity-update slots now emit TWO separate `GraphMutation`s — `RemoveNode(old_id)` with the lower seq_no, then `AddNode/AddEdges` with the higher one. This is a deliberate "delete-then-add" change that sets up cleaner stale-edge filtering once neighborhood-versioning lands in a follow-up PR.
- Return type of `insert::insert` (and `HawkActor::insert`) changes from `VecRequests<Option<ConnectPlan>>` to `VecRequests<Vec<ConnectPlan>>`. The downstream `SingleHawkMutation.plans` field was already a `Vec<ConnectPlan>` — the boundary glue in `handle_mutations` switches from `push` to `extend`.
- `insert_prepare` and `insert_prepare_batch` are deleted along with the dedicated `test_insert_prepare_batch`. The singular `insert_from_search_results` path now inlines its mutation construction and reuses the same prune + compact helpers as the batched loop.

The TODO comments on `prune_invalid_links` and the per-slot prune step in `insert::insert` mark exactly the code that becomes dead once neighborhood-versioning makes stale-edge filtering implicit in `insert_apply`.

Follows #2135 (per-graph update sequence numbers).


sdwoodbury (approved):
this is awesome

mcalancea (approved):
Nice work to get this simplification! 

A few observations from me, but none blocking for this PR:
1. The 3 parallel VecRequests `plans, insert_ids, replace_ids` look awkward to me and may be better as a single VecRequest of a bundled struct and perhaps with some different encoding around the Options. May be a matter of taste
2. The mutation sequence number mechanism could leverage some type safety. Example approach drafted with Claude:

```rust
  pub struct UnstampedMutation<V> {
      pub ops: Vec<MutationOp<V>>,
  }

  pub struct StampedMutation<V> {
      seq_no: u64,                   // private — only constructible via GraphMem::apply
      pub ops: Vec<MutationOp<V>>,
  }

  impl<V> StampedMutation<V> {
      pub fn seq_no(&self) -> u64 { self.seq_no }
  }

  // iris-mpc-cpu/src/hnsw/graph/layered_graph.rs

  impl<V: Ref + ...> GraphMem<V> {
      /// Author a fresh mutation: mints seq_no, applies, returns stamped form
      /// for persistence / WAL emission. No monotonic check — by construction.
      pub fn apply(&mut self, m: UnstampedMutation<V>) -> Result<StampedMutation<V>> {
          let seq_no = self.last_update_seq_no + 1;
          self.apply_ops(&m.ops)?;
          self.last_update_seq_no = seq_no;
          Ok(StampedMutation { seq_no, ops: m.ops })
      }

      /// Apply an externally-stamped mutation (WAL replay, peer sync).
      /// Validates the seq_no is exactly the next-expected one.
      pub fn replay(&mut self, m: &StampedMutation<V>) -> Result<()> {
          let expected = self.last_update_seq_no + 1;
          if m.seq_no != expected {
              bail!("seq_no {} but expected {}", m.seq_no, expected);
          }
          self.apply_ops(&m.ops)?;
          self.last_update_seq_no = m.seq_no;
          Ok(())
      }

      fn apply_ops(&mut self, ops: &[MutationOp<V>]) -> Result<()> { /* current
  pass-1/pass-2 logic */ }
  }

  Call-site shape becomes:

  // insert::insert and insert_from_search_results both:
  let stamped = graph.apply(UnstampedMutation { ops })?;
  slot_outputs[idx].push(stamped);
  // No allocator, no graph.next_sequence_number(), no GraphMutation { seq_no: 0, ... }
  placeholders.

  For the disabled-persistence branch:
  // Carry UnstampedMutation through — never apply, never stamp.
  // Downstream consumer needs to know it's holding proposals, not applied state.
  let proposed = UnstampedMutation { ops: vec![MutationOp::RemoveNode { id: rid }] };
  slot_mutations.push(proposed);
  // Type signature on this path changes to Vec<UnstampedMutation>, distinct from
  // the live path's Vec<StampedMutation>. Forces the boundary to acknowledge the
  distinction.
```

3. It seems sane that the sidecar would always create an upload at batch cut-offs, i.e immediately after a WAL entry that contains all the compactions for some batch. Something to note while reviewing the consequent PRs.

PR #2148, opened 2026-05-22
bump ampc common to use modified http endpoint

sdwoodbury:
the modified http endpoint is in ampc-common. this would help potentially diagnose an issue with staging. 

ampc-common also modified the NetworkHandle trait, requiring an update to `sync_peers()`. 

the modified http endpoint requires the checkpoint sync code to be updated too. the response body is now returned.


bgillesp (approved):
I fixed the compilation errors in the `communication_stats.rs` binary -- otherwise good to go!

PR #2152, opened 2026-05-25
chore(POP-3883): Prevent node 0 locking after becoming unready

wojciechsromek:
Fix HNSW startup synchronization so a node does not begin heavy database loading until all peers have observed its health endpoint. This prevents node 0 from becoming unresponsive during DB load while nodes 1 and 2 are still waiting for a valid /health response.


leonanos8:
@codex review

leonanos8 (approved):

PR #2153, opened 2026-05-25
(bugfix)  free up the tokio runtime while loading the graph and irises

sdwoodbury:
graph deserialization could take a few minutes. This should not be done inside a regular tokio task. 

also run the iris loading on a separate tokio runtime since the default number of download tasks is 32, which is likely close to saturating the tokio runtime. 

restructuring the iris loading code to be aware of the cancellation token is a larger change and will be done as a follow up PR. 

also included is a bugfix where `record_counter`  is changed to a `&mut` so that the correct value is logged after calling `load_db_records_for_aurora`

### DIFF
--- a/iris-mpc-store/src/loader.rs
+++ b/iris-mpc-store/src/loader.rs
@@ -111,7 +111,11 @@ pub async fn load_iris_db(
         );
 
         let (tx, mut rx) = mpsc::channel::<S3StoredIris>(config.load_chunks_buffer_size);
-        tokio::spawn(async move {
+        let import_runtime = tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(s3_load_parallelism)
+            .thread_name("s3-importer")
+            .build()?;
+        import_runtime.spawn(async move {
             fetch_and_parse_chunks(
                 s3_arc,
                 s3_load_parallelism,
--- a/iris-mpc-store/src/loader.rs
+++ b/iris-mpc-store/src/loader.rs
@@ -3,18 +3,16 @@ use crate::{
     fetch_and_parse_chunks, last_snapshot_timestamp, DbStoredIris, S3Store, S3StoredIris, Store,
 };
 use ampc_server_utils::shutdown_handler::ShutdownHandler;
-use async_stream::try_stream;
 use aws_config::Region;
 use eyre::bail;
 use futures::stream::BoxStream;
-use futures::{Stream, StreamExt};
+use futures::StreamExt;
 use iris_mpc_common::config::Config;
 use iris_mpc_common::helpers::inmemory_store::InMemoryStore;
 use std::collections::HashSet;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
-use tokio::sync::mpsc::{self, Receiver};
-use tokio::task::JoinHandle;
+use tokio::sync::mpsc;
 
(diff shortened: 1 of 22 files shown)


iris-mpc-store/src/loader.rs
```diff
@@ -110,26 +113,86 @@ pub async fn load_iris_db(
...
+        }
+        let s3_stream = get_s3_stream(fetch_handle, rx);
+        tokio::pin!(s3_stream);
 
         let mut time_waiting_for_stream = Duration::from_secs(0);
         let mut time_loading_into_memory = Duration::from_secs(0);
         let mut load_summary_ts = Instant::now();
-        while let Some(iris) = rx.recv().await {
+        while let Some(iris) = s3_stream.next().await {
+            let iris = iris?;
```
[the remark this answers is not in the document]
sdwoodbury: disagree. `stream_irises_par` fetches irises that were created after `min_last_modified_time_at`. this comment suggests changing the scope of that to also include a fallback for s3_stream. that is not the intention of this PR.

iris-mpc-store/src/loader.rs
```diff
@@ -110,26 +114,85 @@ pub async fn load_iris_db(
...
+        ) -> impl Stream<Item = Result<S3StoredIris, eyre::Report>> {
+            try_stream! {
+                let mut fetch_done = false;
+                let mut task_error: Option<eyre::Report> = None;
+                tokio::pin!(join_handle);
+                loop {
+                    tokio::select! {
+                        result = &mut join_handle, if !fetch_done => {
+                            fetch_done = true;
+                            if  let Err(e) = result {
```
mcalancea: This is only the outer error (of the join handle) right? If the join succeeds but contains an inner error, it's not caught?

iris-mpc-store/src/loader.rs
```diff
@@ -110,26 +114,85 @@ pub async fn load_iris_db(
...
+
+        // Stream that yields S3StoredIris items from the channel while monitoring
+        // fetch_handle for errors.  Once fetch_handle finishes successfully we
+        // switch to draining the channel directly; if it exits with a JoinError
+        // (i.e. a panic inside the spawned task) we surface that as a stream error.
+        fn get_s3_stream(
+            join_handle: JoinHandle<eyre::Result<()>>,
+            mut rx: Receiver<S3StoredIris>,
+        ) -> impl Stream<Item = Result<S3StoredIris, eyre::Report>> {
+            try_stream! {
```
mcalancea: I've delved into this design a bit with Claude and we arrived at this alternative: https://github.com/worldcoin/iris-mpc/pull/2165/changes. 
- It's simpler in that it doesn't use try_stream!
- It's fail-fast because it does not defer the error; note that in `fetch_and_parse_chunks` each inidvidual handle keeps a copy of the send channel, so the receive channel in the stream does not close when one of them errors, but rather when all of them finish (an error in the join will cause the surviving handles to become detached, but living).
What do you think about it?

mcalancea (commented):
Left a couple of comments.

iris-mpc-store/src/loader.rs
sdwoodbury: this is a good fix. good point about the rx channels potentially remaining open. i merged it in.

mcalancea (approved):
Approved, thanks!

PR #2155, opened 2026-05-26
[POP-3802] Plaintext store for Deep Identifier

iliailia:
Mostly mirrors the existing `PlaintextStore`. The binaries generating plaintext graphs and stores will be updated in a follow-up PR.


bgillesp (approved):
Implementation is looking good!  I tweaked a few small things during review, but nothing major.  I'll integrate changes into #2171 before I do the review there.

PR #2169, opened 2026-05-27
[POP-3901] make load_iris_db cancellable

sdwoodbury:
Make `load_iris_db` respond to SIGTERM via the shutdown handler. 

Also redesign `fetch_and_parse_chunks` to allow waiting for the shutdown signal at the top level, so that it wouldn't get blocked by `sleep`.  using `stream::iter` also removes the head of line blocking, caused by waiting for the next handle in the `VecDeque`.

Failure modes:
1. SIGTERM → shutdown handler fires → `select!` in `load_iris_db` cancels `load_iris_db_internal`. Tasks spawned by `fetch_and_parse_chunks` also catch the signal and terminate.                                                                             
  2. Task error in `fetch_and_parse_chunks` → surfaces via `handle.await??` → `fetch_handle` resolves with `Err` → receiver loop returns it
  3. Shutdown before `RuntimeShutdownGuard` is created → safe because `fetch_and_parse_chunks` tasks also listen for shutdown directly, so `import_runtime` completes quickly regardless

### DIFF
--- a/iris-mpc-store/src/s3_importer.rs
+++ b/iris-mpc-store/src/s3_importer.rs
@@ -593,6 +593,7 @@ mod tests {
             tx,
             1,
             0,
+            Arc::new(ShutdownHandler::new(1)),
         )
         .await;
         let mut count = 0;
@@ -638,6 +639,7 @@ mod tests {
             tx,
             1,
             0,
+            Arc::new(ShutdownHandler::new(1)),
         )
         .await;
 
@@ -695,6 +697,7 @@ mod tests {
             tx,
             5,
             100,
+            Arc::new(ShutdownHandler::new(1)),
         )
         .await;
 
--- a/iris-mpc-store/src/s3_importer.rs
+++ b/iris-mpc-store/src/s3_importer.rs
@@ -5,9 +5,10 @@ use aws_sdk_s3::config::StalledStreamProtectionConfig;
 use aws_sdk_s3::{config::Builder as S3ConfigBuilder, Client as S3Client};
 use aws_sdk_s3::{primitives::ByteStream, Client};
 use eyre::{bail, eyre, Result};
+use futures::{stream, StreamExt};
 use iris_mpc_common::{vector_id::VectorId, IRIS_CODE_LENGTH, MASK_CODE_LENGTH};
-use std::{collections::VecDeque, mem, sync::Arc, time::Duration};
-use tokio::{io::AsyncReadExt, sync::mpsc::Sender, task};
+use std::{mem, sync::Arc, time::Duration};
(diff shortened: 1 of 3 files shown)


iris-mpc-store/src/s3_importer.rs
```diff
@@ -303,84 +306,83 @@ pub async fn fetch_and_parse_chunks(
     } else {
         last_snapshot_details.chunk_size as usize
     };
-    let mut handles: VecDeque<task::JoinHandle<Result<(), eyre::Error>>> =
-        VecDeque::with_capacity(concurrency);
 
-    for chunk in (1..=effective_last_serial_id).step_by(range_size) {
+    let chunk_iterator = (1_i64..=effective_last_serial_id).step_by(range_size);
+    let stream = stream::iter(chunk_iterator).map(|chunk| {
```
mcalancea, line 311: This still needs `task::spawn` around it, otherwise all the buffered futures will run on the same worker thread.

iris-mpc-store/src/s3_importer.rs
```diff
@@ -303,84 +306,83 @@ pub async fn fetch_and_parse_chunks(
...
-        }));
-    }
+        }
+    });
 
-    tracing::info!("All s3 import tasks are spawned. Waiting for them to finish");
-    // Wait for remaining handles
-    for handle in handles {
-        handle.await??;
+    let mut results = stream.buffer_unordered(concurrency);
```
mcalancea, line 336: Nice improvement!

iris-mpc-store/src/s3_importer.rs
```diff
@@ -708,4 +740,77 @@ mod tests {
...
+                tx,
+                20,     // plenty of retries — shutdown should interrupt first
+                10_000, // 10s initial backoff
+                shutdown,
+            ),
+        )
+        .await
+        .expect("cancellation should complete within 500ms of virtual time");
+
+        assert!(result.is_err(), "Expected Err on shutdown during backoff");
```
mcalancea: Tightening this check is reasonable:

`assert!(format!("{err:#}").contains("Shutdown"), "unexpected error: {err:#}");`

mcalancea (commented):
Thanks Stuart! Left a few comments.

iris-mpc-store/src/s3_importer.rs
sdwoodbury: agreed. at 100s of GB it needs the `task::spawn`.

mcalancea (approved):
LGTM. Parallelism restored via spawn-inside-map composition; test tightening covers the main case. Optional minor follow-up: apply the same `.contains("Shutdown")` check to the during_read test for consistency.

PR #2176, opened 2026-05-29
[POP-3915] use new sidecar functions for iris-mpc-hawk and genesis

sdwoodbury:
Consolidate checkpoint restart and memory-efficient graph loading

- Replaces load_graph_and_roll_forward with restart_from_checkpoint in server/mod.rs. Genesis did not get the update because Genesis does not save the WAL, should not be applying it, and has a different startup mechanism (rolling back the modifications and replaying irises from the GPU database). 
- Adds stream_download_and_deserialize_graph_pair — a streaming S3 download path that deserialises directly into GraphMem layer-by-layer, avoiding the peak memory spike of holding a full intermediate Graph alongside the destination graph. V4/Current and V3 use this single-copy path; older formats fall back to the unoptimized reader. download_graph_checkpoint is updated to use this new function.
- Also adds a pruning mode to the serialization layer and expands the streaming S3 integration tests to cover V3 and V4 graph round-trips.

### DIFF
--- a/iris-mpc-cpu/src/checkpoint_protocol/terminal.rs
+++ b/iris-mpc-cpu/src/checkpoint_protocol/terminal.rs
@@ -8,8 +8,8 @@ use crate::checkpoint_protocol::{
 };
 use crate::execution::hawk_main::{BothEyes, LEFT, RIGHT};
 use crate::graph_checkpoint::{
-    stream_serialize_and_upload_with, BlakeTeeWriter, DEFAULT_STREAMING_PARALLELISM,
-    DEFAULT_STREAMING_PART_SIZE,
+    cleanup_checkpoints, stream_serialize_and_upload_with, BlakeTeeWriter, GraphCheckpointState,
+    PruningMode, DEFAULT_STREAMING_PARALLELISM, DEFAULT_STREAMING_PART_SIZE,
 };
 use crate::hnsw::{
     graph::{graph_store::GraphPg, layered_graph::GraphMem},
@@ -29,6 +29,7 @@ pub struct UploadAndRecord<'a, V: VectorStore> {
     pub bucket: String,
     pub party_id: usize,
     pub is_archival: bool,
+    pub pruning_mode: PruningMode,
 }
 
 impl<'a, V: VectorStore + Send + Sync> UploadAndRecord<'a, V> {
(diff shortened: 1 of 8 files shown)


iris-mpc-cpu/tests/streaming_s3_integration.rs
```diff
@@ -205,3 +220,235 @@ async fn streaming_download_round_trip() -> Result<()> {
...
+/// Upload a `[GraphV3; 2]` to localstack S3, stream-download it via
+/// `stream_download_and_deserialize_graph_pair(…, GraphFormat::V3)`, and
+/// assert the result matches the reference `.into()` conversion.
+///
+/// Specifically verifies:
+/// - `entry_points` and `layers` (including recomputed `set_hash`) match
+///   the standard `GraphV3 → GraphMem` path.
+/// - `last_update_seq_no` is 0 for both graphs (V3 has no seq_no field).
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn v3_graph_pair_streams_to_graphmem() -> Result<()> {
```
mcalancea, line 271: These skip in CI — no `S3_TEST_ENDPOINT`/localstack in any workflow, so the V3/V4 round-trip coverage never runs on merge. Worth wiring a localstack service into `run-unit-tests.yaml` so the single-copy path is actually guarded.

iris-mpc-cpu/src/checkpoint_protocol/terminal.rs
```diff
@@ -127,6 +130,27 @@ impl<V: VectorStore + Send + Sync> TerminalAction for UploadAndRecord<'_, V> {
...
+        let graph_checkpoint = GraphCheckpointState {
+            s3_key,
+            last_indexed_iris_id: base.last_indexed_iris_id as _,
+            last_indexed_modification_id: freeze.0,
+            graph_mutation_id: Some(freeze.0),
+            blake3_hash: blake3_hash_hex,
+            graph_version: GraphFormat::Current.version(),
+            is_archival: self.is_archival,
+        };
+        if let Err(e) = cleanup_checkpoints(
```
mcalancea, line 142: `cleanup_checkpoints` deletes the S3 object before the DB row. On partial failure that leaves a dangling row in the advertised window, which a later restart could pick as base and fail to download. Safer to delete the row first.

mcalancea (approved):
Thanks Stuart!

iris-mpc-cpu/tests/streaming_s3_integration.rs
sdwoodbury: will add this in a follow up pr

PR #2178, opened 2026-06-02
Type-state for graph mutation sequence numbers

mcalancea:
## What

Adds compile-time type safety to the graph mutation sequence-number ("mutation id") logic, and removes the now-redundant `UpdateSeqNoAllocator`.

- **`UnstampedMutation<V> { ops }`** — a mutation's ops with *no* seq_no yet ("what to do", no position in the sequence). The neighborhood helpers (`updated_`/`expanded_neighborhoods`) move onto it, since they're computed pre-apply.
- **`GraphMem::apply_new(UnstampedMutation) -> Result<GraphMutation>`** — the sole in-process minter. It stamps with `next_sequence_number()` and applies in a single step, returning the stamped `GraphMutation` for collection/persistence.
- **`insert_apply` / `insert_apply_all` unchanged** — they remain the *apply-already-stamped* path for WAL/checkpoint replay (and test literals), with the strict-monotonicity check as the runtime backstop for externally-supplied `seq_no`s.

All live mint sites (`insert.rs` ×4, `searcher.rs` ×3) now build `UnstampedMutation` + `apply_new`.

## Why

Previously ordering was enforced *only at runtime*: `GraphMutation` was a plain struct anyone could build with any `seq_no`, and two construction patterns coexisted (per-mutation `next_sequence_number()` vs. a separate allocator minting a contiguous run). Nothing in the types prevented minting out of order, reusing an id, or applying a never-stamped mutation.

The key invariant now holds *structurally*: by **fusing stamp + apply** into `apply_new` (rather than exposing a standalone `stamp()` whose result a caller could sit on), a seq_no is never observable until it has already been assigned and applied — so `last_update_seq_no` can't lag the highest minted id and in-process mutations can't collide on a number.

`UpdateSeqNoAllocator` is deleted because every `mint()` was already immediately followed by `insert_apply()`, so the allocator's counter never diverged from the graph's — the graph is the single source of truth.

## Trust boundary (deliberate)

`GraphMutation` is still freely constructible with an arbitrary `seq_no` (needed for `Deserialize` on WAL/checkpoint replay). In-process minting is sealed behind `apply_new`; persisted/cross-process seq_nos can't be compile-time-trusted and are validated at runtime in `insert_apply`. A `SeqNo(u64)` newtype was considered and rejected — the only hand-written `seq_no` literals are tests and replay (both legitimate), so it'd be ceremony guarding no real failure mode.

## Tests

`cargo test -p iris-mpc-cpu` (mutation / layered_graph / insert / searcher) green; `cargo clippy --tests` + `cargo fmt` clean. No behavioral change — `apply_new` returns exactly what the allocator-minted path did.

### DIFF
--- a/iris-mpc-cpu/src/execution/hawk_main/insert.rs
+++ b/iris-mpc-cpu/src/execution/hawk_main/insert.rs
@@ -1,5 +1,8 @@
 use crate::hnsw::{
-    graph::{mutation::EdgeType, GraphMutation, MutationOp, UpdateEntryPoint},
+    graph::{
+        mutation::{EdgeType, UnstampedMutation},
+        MutationOp, UpdateEntryPoint,
+    },
     searcher::{ConnectPlanV, LayerMode},
     vector_store::VectorStoreMut,
     GraphMem, HnswSearcher, VectorStore,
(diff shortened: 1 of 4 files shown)


bgillesp (approved):
Looks good, thanks!

PR #2181, opened 2026-06-02
 [POP-3930] add option to run sidecar in the same process as the iris-mpc server

sdwoodbury:
this is to facilitate sidecar testing. although a separate service is desirable, it will take time before the dev cluster is set up for that. In the mean time, this will allow us to continue testing.

### DIFF
--- a/iris-mpc/src/server/mod.rs
+++ b/iris-mpc/src/server/mod.rs
@@ -3,9 +3,10 @@ use crate::services::processors::batch::receive_batch_stream;
 use crate::services::processors::job::process_job_result;
 use aws_sdk_s3::Client;
 use aws_sdk_sns::types::MessageAttributeValue;
-use iris_mpc_cpu::checkpoint_protocol::{restart_from_checkpoint, RestartOutcome};
+use iris_mpc_cpu::checkpoint_protocol::{restart_from_checkpoint, sidecar_main, RestartOutcome};
 use iris_mpc_cpu::graph_checkpoint::sync_graph_mutations;
 use iris_mpc_cpu::hnsw::GraphMem;
+use iris_mpc_cpu::network::mpc::{build_network_handle, NetworkHandleArgs};
 
 use crate::services::processors::modifications_sync::{
     send_last_modifications_to_sns, sync_modifications,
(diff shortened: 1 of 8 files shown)


iris-mpc/src/server/mod.rs
```diff
@@ -214,9 +216,55 @@ pub async fn server_main(config: Config) -> Result<()> {
...
+
+    // Extract what the sidecar needs from config/aws_clients before the async
+    // move block, since main_future already holds &config and &aws_clients.
+    let sidecar_party_id = config.party_id;
+    let sidecar_tls = config.tls.clone();
+    let sidecar_s3 = aws_clients.checkpoint_s3_client.clone();
+    let shutdown_ct = shutdown_handler.get_network_cancellation_token();
+    // the sidecar needs its own graph store and this function transforms config to the appropriate
+    // schema name when creating the pg client
+    let (_iris_store, graph_store) = prepare_stores(&config).await?;
```
mcalancea: _Claude_

This runs on **every** server start, before the sidecar-enabled (`config: Some`) check, so it opens a second RW Postgres pool against the same schema even when the sidecar is off — and the `?` means a failure here aborts `server_main` on the normal path. Suggest moving `prepare_stores` (and the network/S3/token setup) inside `sidecar_future`, after the `Some(sc_config)` guard, so nothing is allocated or can fail when the sidecar is disabled. (`_iris_store` is also built and immediately dropped.)

iris-mpc/src/server/mod.rs
```diff
@@ -214,9 +216,55 @@ pub async fn server_main(config: Config) -> Result<()> {
...
+            sc_config,
+            &graph_store,
+            &sidecar_s3,
+            &mut network_handle,
+            shutdown_ct.clone(),
+        )
+        .await
+    };
 
+    tokio::try_join!(main_future, sidecar_future)?;
```
mcalancea, line 291: _Claude_

`shutdown_ct` (the network cancellation token) is only fired inside `wait_for_pending_batches_completion()`, which `run_main_server_loop` calls **only on its `Ok` branch**. On the actor-error branch it returns `Ok(())` without cancelling the token, so `main_future` resolves but `sidecar_future` loops forever and `try_join!` hangs — `server_main` never exits. The happy path only works because batch-completion happens to fire the token. Also note `try_join!` short-circuits: a sidecar `CycleError::Fatal` would tear down the live request server. Suggest driving the sidecar off a token cancelled whenever `main_future` finishes (cancel in both branches, or `select!`/detached task), and treating sidecar failure as log-and-ignore rather than fatal.

mcalancea (approved):
Looks good! I pushed some small changes and left two comments.

PR #2185, opened 2026-06-04
Support deep-ID Int4 vectors in generate_ideal_neighborhoods

iliailia:
## Summary
- Extends the `generate_ideal_neighborhoods` binary (previously iris-only) to compute ideal top-k neighborhoods for deep-ID Int4 vectors.
- Adds a unified `EngineKind` CLI enum so a single `--engine-choice` selects either an iris engine (`NaiveFHD/MinFHD/NHD/MinNHD`) or `NaiveInt4Dot`; int4 mode loads an Int4 ndjson via `int4_vectors_from_ndjson` and runs `EngineInt4`, else the iris path runs as before.
- Introduces an `AnyEngine { Iris, Int4 }` dispatch wrapper so the chunk/append/timing loop stays single-copy.
- Records engine identity in the resume header and makes `iris_selection` optional (`None` for int4), closing a latent gap where a results file built with one distance could be silently resumed with another.
- Output line format is unchanged (`KNNResult<u32>`), so the sibling binary `generate_ideal_graph.rs` consumes the produced layer-0 file with no changes — verified by a new integration test against the consumer's exact reader (`read_knn_results_from_file`).

The neighborhoods binary uses the engine path (`NaiveKNN`), not `PlaintextDeepIDStore`, mirroring how the iris path uses `Engine` rather than the iris store; the store itself is untouched.

## Test Plan
- [x] `cargo test -p iris-mpc-cpu engine_kind_tests` — 3 pass (EngineKind helpers)
- [x] `cargo test -p iris-mpc-bins --bin generate_ideal_neighborhoods` — 2 pass (header serde round-trip, iris + int4)
- [x] `cargo test -p iris-mpc-cpu --test deep_id_neighborhoods_format` — 1 pass (int4 output consumable by graph reader; node ids contiguous `1..=N`)
- [x] `cargo clippy -p iris-mpc-cpu --all-targets -- -D warnings` and the binary — clean
- [x] `cargo fmt` — no diff

### DIFF
--- a/iris-mpc-bins/bin/iris-mpc-cpu/generate_ideal_neighborhoods.rs
+++ b/iris-mpc-bins/bin/iris-mpc-cpu/generate_ideal_neighborhoods.rs
@@ -5,10 +5,13 @@ use std::{
 };
 
 use clap::Parser;
-use iris_mpc_common::{iris_db::iris::IrisCode, IrisSerialId};
+use iris_mpc_common::{iris_db::iris::IrisCode, vector_id::SerialId};
 use iris_mpc_cpu::{
-    hawkers::ideal_knn_engines::{Engine, EngineChoice, KNNResult},
-    utils::serialization::iris_ndjson::{irises_from_ndjson_iter, IrisSelection},
+    hawkers::ideal_knn_engines::{Engine, EngineInt4, EngineKind, KNNResult},
+    utils::serialization::{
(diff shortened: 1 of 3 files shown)


bgillesp (approved):
Good to go, thanks so much!

bgillesp:
Force push required to squash history due to an errant unsigned merge commit.

PR #2188, opened 2026-06-09
[POP-3958] add integration tests for sidecar

sdwoodbury:
Similar to `iris-mpc-upgrade-hawk/tests/workflows`, exercise the various WAL scenarios:
- losing a checkpoint and using a previous one
- rolling forward hawk mutations
- loading a checkpoint made from a v3 graph
etc

wal_109 and wal_110 replace the `e2e_hawk` integration test. 

In a follow-up PR, the sidecar will be modified to allow using checkpoints where the graph format is V3. Until then, wal_test_100 is expected to fail.


iris-mpc-cpu/src/checkpoint_protocol/mod.rs
```diff
@@ -58,6 +58,26 @@ pub struct CheckpointMeta {
...
+    /// **Do not use `PartialEq` for cross-party comparisons.**
+    /// `checkpoint_id` is a DB-local auto-increment primary key; each party's
+    /// database independently assigns its own value for the same checkpoint, so
+    /// two `CheckpointMeta` objects that refer to the same snapshot will have
+    /// different `checkpoint_id` values and would incorrectly compare unequal.
+    ///
+    /// `blake3_hash` is the cryptographic content hash — it is always identical
+    /// across parties for the same checkpoint.  `last_indexed_modification_id`
+    /// and `graph_version` are included as a cheap sanity cross-check.
+    pub fn same_checkpoint(&self, other: &Self) -> bool {
```
mcalancea: The doc warns against `PartialEq` for cross-party comparison, but the derive remains — `==` still compiles and silently does the wrong thing. If there's no remaining same-party use, drop the derive; if there is, worth auditing the use sites.

Also: why is `graph_mutation_id` excluded from the sanity cross-check while `last_indexed_modification_id` is included?

iris-mpc/tests/utils/key_rotation.rs
```diff
@@ -0,0 +1,206 @@
...
+// ---------------------------------------------------------------------------
+// Key rotation
+// ---------------------------------------------------------------------------
+
+/// Generate a fresh ECDH key pair, upload the public key to S3, and store
+/// the private key in Secrets Manager.
+///
+/// Signature matches what the `key-manager` binary's `rotate` subcommand does
+/// so callers can use the same arguments as the CLI.
+pub async fn rotate_keys(
```
mcalancea, line 124: This manually mirrors the key-manager binary's `rotate`; if the binary changes scheme or naming, the two drift silently. Consider extracting the rotate logic into a shared fn (e.g. `iris-mpc-common`, next to `key_pair.rs` which holds the consume side) called by both. Non-blocking.

iris-mpc-cpu/tests/utils/runner.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/// Context passed to every lifecycle method.
+pub struct CpuTestContext {
+    pub configs: CpuConfigs,
+    pub env: TestEnvironment,
+    /// Pre-built S3 client pointed at the LocalStack instance for this environment.
+    /// Shared across all lifecycle phases to guarantee a consistent endpoint configuration.
+    pub s3_client: aws_sdk_s3::Client,
+    /// Test number (100–104) for log tagging and config selection.
```
mcalancea: Stale: 100–110 now.

iris-mpc-cpu/tests/workflows/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+macro_rules! run_hawk {
+    ($configs:expr, $shutdown:expr, $ctx:expr) => {{
+        use iris_mpc::server::server_main;
+        use std::sync::Arc;
+        use tokio::sync::Notify;
+        use tracing::{info_span, Instrument};
+
+        let mut join_set: tokio::task::JoinSet<eyre::Result<()>> = tokio::task::JoinSet::new();
+
+        // Arc<Notify> is used instead of CancellationToken because the latter
```
mcalancea: This comment is incorrect — `CancellationToken` is `Send + Sync` and runtime-agnostic; it can be cloned into each `spawn_blocking` closure and selected on directly.

The `Notify` substitution is also racy: `notify_waiters()` is edge-triggered (wakes only currently-parked waiters, stores no permit), so if cancellation fires before a party thread has entered its select, that party never observes shutdown and `stop_and_join!` hangs. Using the token directly removes both the race and this comment.

iris-mpc-cpu/tests/workflows/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+        join_set
+    }};
+}
+
+/// Spawn `sidecar_main` for all 3 parties concurrently in one-shot mode.
+///
+/// Uses `CpuNodeConfig.sidecar_port` so hawk_main and sidecar_main can
+/// co-exist in the same test process without port conflicts.
+#[macro_export]
+macro_rules! run_sidecar {
```
mcalancea: Could be a plain async fn — nothing here needs macro expansion, and `expect_sidecar_success` already is one. Same for `stop_and_join!`. Better type errors and IDE navigation. Non-blocking.

iris-mpc-cpu/tests/workflows/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+    }};
+}
+
+/// Cancel the shared `CancellationToken` and drain the `JoinSet`.
+///
+/// Returns the first task error encountered, or `Ok(())` if all tasks shut
+/// down cleanly.  Always drains the full set regardless.  Should be called
+/// even when the test is about to fail so that background tasks are cleaned up.
+#[macro_export]
+macro_rules! stop_and_join {
```
mcalancea: No timeout — if a server ignores cancellation (e.g. the `Notify` race above), this hangs CI with no diagnostic. `expect_sidecar_success` has a 60 s cap; same treatment here?

iris-mpc-cpu/tests/utils/cpu_node.rs
```diff
@@ -0,0 +1,853 @@
...
+    /// 1. Query `MAX(modification_id)` from `hawk_graph_mutations`.
+    /// 2. Read all `hawk_graph_mutations` rows up to that id.
+    /// 3. Replay mutations into `[GraphMem<IrisVectorId>; 2]` (left + right eye).
+    /// 4. Serialize with `bincode::serialize` — same wire format as the sidecar.
+    /// 5. Compute BLAKE3 hash of the bytes.
+    /// 6. Upload to S3 at key `"{party_id}/checkpoints/{label}_{last_modification_id}"`,
+    ///    where `label` is `"archival"` when `is_archival` is true, otherwise `"seed"`.
+    /// 7. Insert `genesis_graph_checkpoint` DB row inside a transaction.
+    /// 8. Return the newly inserted `GraphCheckpointRow`.
+    async fn make_checkpoint_with_options(
```
mcalancea: ~80 lines duplicated with `make_checkpoint_v3` (replay → serialize → hash → upload → insert), differing only in serializer and version. Note the literal `4` passed as `graph_version` below, vs `GraphFormat::V3.version()` in the v3 twin — if Current bumps to v5 this silently keeps minting "v4" rows. Suggest one fn taking a `GraphFormat`, using `GraphFormat::Current.version()`.

iris-mpc-cpu/tests/utils/wal_builder.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    pub async fn build(&mut self, nodes: &CpuNodes) -> eyre::Result<()> {
+        self.insert_mutations_all(nodes).await?;
+        self.seed_modifications_all(nodes).await?;
+        self.processed = self.entries.len();
+        Ok(())
+    }
+
+    pub async fn build_single(
```
mcalancea: Watermark trap: the single `processed` watermark gates both tables and is bumped even on partial writes — `build_single(node, true, false)` makes those ids unseedable via this API afterwards (wal_110 works around it by calling `insert_mutations` directly). Relatedly, `set_persisted` after `build()` is a silent no-op (watermark + `ON CONFLICT DO NOTHING`). Per-table watermarks, or a doc comment stating the must-precede-first-build constraint.

iris-mpc/tests/workflows/wal_106.rs
```diff
@@ -0,0 +1,143 @@
...
+        builder.build(nodes).await?;
+
+        // Phase 2: sidecar runs. Wait until each party's checkpoint count exceeds its individual baseline.
+        {
+            let baselines = nodes.checkpoint_counts().await?;
+
+            let shutdown = CancellationToken::new();
+            let sidecar_set = run_sidecar!(ctx.configs, shutdown.clone(), ctx);
+
+            let wait_res = timeout(Duration::from_secs(120), async {
```
mcalancea, line 87: With `one_shot: true`, joining the set already implies counts are final — this 120 s poll looks redundant, and it runs *before* the join, so an instantly-failing sidecar stalls 2 minutes before the error surfaces. Join first, then check counts (or drop the loop).

iris-mpc-cpu/tests/workflows/wal_109.rs
```diff
@@ -0,0 +1,148 @@
...
+    async fn exec_assert(&mut self, _ctx: &CpuTestContext) -> eyre::Result<()> {
+        let nodes = self.nodes.as_ref().unwrap();
+
+        let post = WalAssertions::new()
+            .assert_wal_row_count(3 * MIN_MUTATIONS_PER_SIDECAR_CYCLE)
+            .assert_checkpoint_count(2)
+            .assert_latest_checkpoint_mod_id(3 * MIN_MUTATIONS_PER_SIDECAR_CYCLE as i64);
+        nodes.apply_uniform_assertions(&post).await?;
+
+        nodes.assert_checkpoint_hashes_agree().await
```
mcalancea: `assert_consensus_and_reference()` as in wal_105/106? Agreement alone passes if sync corrupts all three parties identically.

scripts/run-tests-hawk-e2e.sh
```diff
@@ -8,18 +8,9 @@ function _main()
...
-    cargo run --release -p iris-mpc-bins --bin key-manager --  --region us-east-1 --endpoint-url http://localstack:4566 --node-id 1 --env dev rotate --public-key-bucket-name wf-dev-public-keys
-    cargo run --release -p iris-mpc-bins --bin key-manager --  --region us-east-1 --endpoint-url http://localstack:4566 --node-id 2 --env dev rotate --public-key-bucket-name wf-dev-public-keys
-
-    cargo run --release -p iris-mpc-bins --bin key-manager --  --region us-east-1 --endpoint-url http://localstack:4566 --node-id 0 --env dev rotate --public-key-bucket-name wf-dev-public-keys
-    cargo run --release -p iris-mpc-bins --bin key-manager --  --region us-east-1 --endpoint-url http://localstack:4566 --node-id 1 --env dev rotate --public-key-bucket-name wf-dev-public-keys
-    cargo run --release -p iris-mpc-bins --bin key-manager --  --region us-east-1 --endpoint-url http://localstack:4566 --node-id 2 --env dev rotate --public-key-bucket-name wf-dev-public-keys
-
-    cargo test --release --test e2e_hawk -- --include-ignored
+    cargo test --release -p iris-mpc-cpu --test e2e_wal -- --include-ignored --skip test_wal_100
```
mcalancea: The re-enable for wal_100 lives here, in a different file from the test, with no ticket reference at either site — nothing forces removal when v3 selection lands. Add the POP ticket ref here and at the test.

iris-mpc-cpu/Cargo.toml
```diff
@@ -78,6 +78,12 @@ ampc-server-utils.workspace = true
 criterion = { version = "0.5.1", features = ["async_tokio"] }
 rayon.workspace = true
 num_cpus = "1.16.0"
+aws-config.workspace = true
+aws-sdk-secretsmanager.workspace = true
+serial_test.workspace = true
+sodiumoxide.workspace = true
+iris-mpc = { path = "../iris-mpc" }
```
mcalancea: This dev-dep cycle means every iris-mpc-cpu edit rebuilds the whole server crate before cpu tests link (and cycles choke some tooling). Why do these tests live here rather than in `iris-mpc` — it's `server_main` under test, and that's where e2e_hawk's successor would naturally sit?

mcalancea (approved):

iris-mpc-cpu/Cargo.toml
sdwoodbury: that's a good point

iris-mpc-cpu/src/checkpoint_protocol/mod.rs
sdwoodbury: i'm not sure why `graph_mutation_id` was left out. it might have been added after `CheckpointMeta` was created. i wound up implementing `Eq` and `PartialEq` correctly and removing the special function. and including `graph_mutation_id` in the check.

scripts/run-tests-hawk-e2e.sh
sdwoodbury: https://linear.app/worldcoin/issue/POP-3996/make-sidecar-allow-checkpoints-with-graphformatv3

PR #2213, opened 2026-06-17
[POP-4028] Concretize GraphMem and related code

sdwoodbury:
As an intermediate step to `GraphV5`, which will use half the space (an enormous savings at scale), concretize `GraphMJem`. In this PR, `GraphMem` becomes concrete over `VectorId` and in the future it will be changed to `SerialId`.

As part of this effort, also remove support for unsorted neighborhoods and "single" entrypoints. 

The last 2 commits to remove support for unsorted neighborhoods and "single" entrypoints were reviewed already when being merged into `sw/concretize_graph_mem`. Changes to `main` necessitated cherry-picking onto a new branch.

### DIFF
--- a/iris-mpc-cpu/src/hawkers/ideal_knn_engines.rs
+++ b/iris-mpc-cpu/src/hawkers/ideal_knn_engines.rs
@@ -43,14 +43,41 @@ impl KNNResult {
     }
 }
 
+/// The on-disk serialization format for [`KNNResult`]: plain `u32` serial IDs.
+///
+/// This is the stable file format used by both the generator and the reader.
+/// `IrisVectorId` is intentionally not used here so that the file format stays
+/// compact and backward-compatible (no `{id, version}` objects).
+#[derive(Serialize, Deserialize)]
+pub struct KNNResultU32 {
+    pub node: u32,
+    pub neighbors: Vec<u32>,
+}
+
+impl From<&KNNResult> for KNNResultU32 {
+    fn from(r: &KNNResult) -> Self {
+        KNNResultU32 {
+            node: r.node.serial_id(),
+            neighbors: r.neighbors.iter().map(|v| v.serial_id()).collect(),
+        }
+    }
+}
+
+impl From<KNNResultU32> for KNNResult {
+    fn from(r: KNNResultU32) -> Self {
+        KNNResult {
+            node: IrisVectorId::from_serial_id(r.node),
+            neighbors: r
+                .neighbors
+                .into_iter()
+                .map(IrisVectorId::from_serial_id)
+                .collect(),
+        }
+    }
+}
+
 /// Reads a `Vec<KNNResult>` from a file, skipping the first line (header).
 pub fn read_knn_results_from_file(path: PathBuf) -> std::io::Result<Vec<KNNResult>> {
-    #[derive(Deserialize)]
-    struct KNNResultU32 {
-        node: u32,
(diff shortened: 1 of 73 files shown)


iris-mpc-cpu/src/hnsw/searcher.rs
```diff
@@ -403,89 +369,63 @@ impl HnswSearcher {
...
-                    .unzip();
-                let ep_vectors = store.only_valid_vectors(ep_vectors).await;
-                metrics::gauge!("entry_points_count").set(ep_vectors.len() as f64);
-
-                // TODO when updating entry points, should check for invalid vectors and remove
+        // Get all valid entry points
+        let (ep_vectors, ep_layers): (Vec<_>, Vec<_>) = graph
+            .entry_points
+            .iter()
+            .map(|ep| (ep.point, ep.layer))
```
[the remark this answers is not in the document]
sdwoodbury: this matches the previous behavior.

iris-mpc-cpu/src/hawkers/ideal_knn_engines.rs
```diff
@@ -20,33 +20,37 @@ use crate::hawkers::{
 };
 
 #[derive(Serialize, Deserialize, Clone, Default)]
-pub struct KNNResult<V> {
-    pub node: V,
-    pub neighbors: Vec<V>,
+pub struct KNNResult {
+    pub node: IrisVectorId,
+    pub neighbors: Vec<IrisVectorId>,
 }
```
bgillesp: I think we should probably revert the change here, keeping the signature `KNNResult<V>`.  Output from the `generate_ideal_neighbors` binary reflects the contents of KNNResult, so the new version produces a big serialized JSON string per id, compared with just the concise `u32` provided by the original output.  The nearest neighbor data files produced are already quite large without this change.

I guess that the generic here primarily is used to distinguish between the unversioned `SerialId` and versioned `VectorId` choices, so it isn't really the same distinction as the graph generic.

sdwoodbury: agreed. already in progress.

iris-mpc-cpu/src/hnsw/searcher.rs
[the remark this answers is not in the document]
sdwoodbury: added a function to `VectorStore` to address this.

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -575,40 +560,32 @@ where
...
+            let k = searcher.params.get_M_max(i + 1).min(n - 1);
+
+            let mut engine = NaiveKNN::<K>::init(knn_proto.clone(), vectors, k, 1);
+            let results = engine
+                .compute_chunk(n)
+                .into_iter()
+                // remap from engine 1-based indices to original vector ids
+                .map(|result| result.map(|i| vector_ids[(i as usize) - 1]))
+                .collect::<Vec<_>>();
+            Layer::from_knn_results(results, n)
```
bgillesp: Is there a reason to inline this function call instead of calling into the `Layer::ideal_from_data` implementation?

sdwoodbury: good question. i'll look into changing it back

bgillesp (approved):
Looking good!  Thanks for the couple of revisions.

PR #2220, opened 2026-06-18
Add per-layer ef_search config for HNSW search

mcalancea:
## What

Adds an optional `hnsw_param_ef_search_layers` config field that overrides the per-layer search `ef` (indexed from layer 0 upward). This enables `ef > 1` on upper layers (e.g. layer 1) in the live Hawk Main search path, which was previously hardcoded to greedy `ef = 1` above layer 0.

## Why

The HNSW traversal already dispatches greedy/std/batched search per layer based on the `ef` value — only the config plumbing to set upper-layer `ef` was missing. The accuracy binary already supports per-layer `ef` via its own config; this brings the equivalent knob to the live path.

## Behavior

- **Unset (default):** existing scalar `hnsw_param_ef_search` applies unchanged — layer 0 gets the scalar, layers 1+ stay greedy (`ef=1`). Fully backward-compatible.
- **Set:** the vec fully replaces the `ef_search` array (including layer 0). Padding reuses the last value, so `[256, 8]` → `[256, 8, 8, 8, 8]` and `[256, 1]` keeps upper layers greedy.
- Affects **search only** (`ef_search`), not insertion-time search (`ef_constr_search`).
- Participates in the cross-party `CommonConfig` hash, so all 3 parties must set the same value or the startup sanity check fails (search must be deterministic across parties).

## Usage

\`\`\`toml
hnsw_param_ef_search_layers = [256, 8]   # layer 0 = 256, layers 1+ = 8
\`\`\`

### DIFF
--- a/iris-mpc-bins/bin/iris-mpc-cpu/sidecar.rs
+++ b/iris-mpc-bins/bin/iris-mpc-cpu/sidecar.rs
@@ -122,6 +122,7 @@ fn hawk_args_from(args: &SidecarArgs, tls: Option<TlsConfig>) -> HawkArgs {
         hnsw_param_ef_constr: 0,
         hnsw_param_m: 0,
         hnsw_param_ef_search: 0,
+        hnsw_param_ef_search_layers_override: None,
         hnsw_param_ef_supermatch: 0,
         hnsw_param_ef_saturation_margin: 0,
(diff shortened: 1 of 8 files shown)


bgillesp (approved):
Looks good, thanks!

PR #2222, opened 2026-06-19
HNSW: add a configurable floor on layer-search target batch size

mcalancea:
## What

`layer_search_batched_v2` previously treated the configured layer-search batch size as a **hard override** — when set, it replaced the adaptive insertion-rate estimator with a constant `target_batch_size` for every iteration.

This changes it to a **lower bound (floor)**. The estimator still computes its target from the unchanged `ef / 2` base:

```rust
let target_batch_size = (insertion_batch_size * ins_rate_denom / INS_RATE_NUM)
    .max(min_batch_size.unwrap_or(0));
```

So each iteration opens at least `min_batch_size` neighbors; once the adaptive estimate climbs past the floor it takes over and scales as before. The ceiling stays `ef/2 * 1000` (the existing `MAX_INS_RATE_DENOM` cap) — the floor does not raise it.

## Rename

The knob is a minimum, not a fixed value, so it's renamed to match:

| Kind | Old | New |
|------|-----|-----|
| `HnswSearcher` field | `fixed_layer_search_batch_size` | `min_layer_search_batch_size` |
| Config field | `hnsw_fixed_layer_search_batch_size` | `hnsw_min_layer_search_batch_size` |
| Env var | `SMPC__HNSW_FIXED_LAYER_SEARCH_BATCH_SIZE` | `SMPC__HNSW_MIN_LAYER_SEARCH_BATCH_SIZE` |
| CLI flag | `--hnsw-fixed-layer-search-batch-size` | `--hnsw-min-layer-search-batch-size` |

## Deploy impact

`deploy/dev/common-values-ampc-hnsw.yaml` is updated to the new env var name (value `8000`). **dev** now opens at least 8000 neighbors per iteration, with the adaptive estimator taking over above that. **stage/prod/e2e** don't set the knob → unchanged.

> ⚠️ The env var rename must land on the deploy branch in lockstep — the old `SMPC__HNSW_FIXED_*` name will no longer be read.

## Cross-party config

`hnsw_min_layer_search_batch_size` is promoted from a per-party tuning knob into `CommonConfig`, so the startup sanity-check hash now enforces that all 3 parties agree on the value (a per-party mismatch changes search behavior and is treated as a misconfiguration). dev sets it identically via the shared `common-values-ampc-hnsw.yaml`; stage/prod/e2e leave it unset (`None`) → still consistent.


iris-mpc-common/src/config/mod.rs
```diff
@@ -767,7 +768,7 @@ impl From<Config> for CommonConfig {
             hnsw_param_ef_supermatch,
             hnsw_param_ef_saturation_margin,
             hnsw_layer_density,
-            hnsw_fixed_layer_search_batch_size: _, // per-party tuning knob
+            hnsw_min_layer_search_batch_size: _, // per-party tuning knob
```
bgillesp: This should be added to common config, not dropped -- tuning can't be per-party as it impacts the MPC coordinated search algorithm.

mcalancea: For sure, it's just I remember editing all 3 config files for M, ef and such for the longest time 🤔. I think somebody did a clean-up to enforce commonality more strictly recently. Will move this to common as well 👍 .

bgillesp (dismissed):

wojciechsromek (approved):

PR #2229, opened 2026-06-23
Buffered restore for hawk-main checkpoint materialize

mcalancea:
## Problem

The hawk-main WAL restart path deserializes the graph checkpoint via a single-threaded streaming reader. On a prod-scale checkpoint it's much slower than the pre-WAL parallel download and wedged startup for over an hour.

## Fix

Restore the buffered restore path for the restart materialize:
- `materializer.rs` — parallel `download_graph` → blake3 verify → `read_graph_pair` from memory, run on a blocking thread (off the async runtime).
- `From<graph_v{3,4}::Layer>` — pre-size the link map (`Layer::with_capacity`) to avoid rehashing.
- `read_graph_pair` — convert the two eyes concurrently (`rayon::join`); the bincode decode stays serial, the per-eye conversion is the dominant independent cost.

`set_hash` is recomputed from the links on load (not trusted from the checkpoint): older genesis checkpoints (e.g. the current prod base, written by genesis v0.31.11) carry a `set_hash` from a prior fold algorithm, so the stored value would diverge from what the running node computes and feeds into the cross-party `StateChecksum`.

Compatible with V3 (prod) and V4 (dev). Peak memory (the streaming path's only benefit) is irrelevant on the restart hosts.

## Testing

Unit tests: write→read round trip preserves the recomputed checksum; the V3 conversion ignores a deliberately-wrong stored `set_hash`. fmt/clippy/doc clean. Independent review: SHIP. Verified on the dev cluster.

### DIFF
--- a/iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
+++ b/iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
@@ -635,6 +635,15 @@ impl Layer {
         }
     }
 
+    /// Empty layer with the link map pre-sized for `n` nodes, avoiding repeated
+    /// rehashing when bulk-loading a known-size layer.
+    pub fn with_capacity(n: usize) -> Self {
+        Layer {
+            links: HashMap::with_capacity(n),
+            set_hash: SetHash::default(),
+        }
+    }
+
     pub fn get_links(&self, from: &VectorId) -> Option<&[VectorId]> {
         self.links.get(from).map(|v| v.as_slice())
     }
--- a/iris-mpc-cpu/src/utils/serialization/graph.rs
+++ b/iris-mpc-cpu/src/utils/serialization/graph.rs
@@ -260,11 +260,21 @@ fn read_pair<R: std::io::Read + ?Sized, G: for<'a> Deserialize<'a>>(
     Ok(data)
 }
 
+/// Convert a decoded wire-format pair into `GraphMem`s, running the two eyes'
+/// (independent) conversions concurrently. The preceding bincode decode is serial.
+fn convert_pair_parallel<G>(pair: [G; 2]) -> [GraphMem; 2]
+where
+    G: Into<GraphMem> + Send,
+{
+    let [left, right] = pair;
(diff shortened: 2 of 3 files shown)


carlomazzaferro (dismissed):

iris-mpc-cpu/src/utils/serialization/graph.rs
bgillesp, line 331: Should we update the logic for these versions to use `convert_pair_parallel` as well?

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -635,6 +635,15 @@ impl Layer {
         }
     }
 
+    /// Empty layer with the link map pre-sized for `n` nodes, avoiding repeated
+    /// rehashing when bulk-loading a known-size layer.
```
bgillesp, line 639: How does pre-sizing the HashMap prevent re-hashing?  The hashing we're talking about is logical in terms of the inserted neighborhoods, so I don't see how internal allocation within the HashMap struct impacts the resulting value.

iris-mpc-cpu/src/utils/serialization/graph.rs
```diff
@@ -536,7 +540,10 @@ impl From<graph_v3::EntryPoint> for layered_graph::EntryPoint {
 
 impl From<graph_v3::Layer> for Layer {
     fn from(value: graph_v3::Layer) -> Self {
-        let mut layer = Layer::new();
+        // Recompute set_hash via set_links rather than trusting the stored
+        // value: older checkpoints carry a set_hash from a prior fold algorithm.
+        // Pre-size the map so the bulk insert doesn't rehash.
```
bgillesp, line 545: "bulk insert doesn't rehash" -- as above, not clear what this is talking about.

bgillesp (commented):
Just a few small comments, otherwise looks good.

bgillesp (approved):
Approving now for time constraint -- can address comments later.

PR #2230, opened 2026-06-23
db-sanity-check: per-bucket degree stats, single-source hop BFS, SCC counts

mcalancea:
## What

Adds structural-navigability diagnostics to `db-sanity-check` for validating HNSW graph topology — reachability, connectivity, and degree distribution — **from graph topology alone** (no vectors, no MPC). All new passes are flag-gated and off by default; `--bucket-size <N>` is the one new required arg.

## Added

- **Per-bucket degree** (`degree_by_bucket.csv`): out- and in-degree min/avg/median/max per serial-id bucket. In-degree is version-strict (valid).
- **Neighbor-serial matrix** (`neighbor_serial_matrix.csv`): per-(eye, layer) bucket→bucket directed-edge counts, `valid` (live→live) + `raw` (live→any). Sparse; surfaces the serial-locality of edges.
- **`--bfs-hops`** (`hops_by_bucket.csv`): single-source layered BFS (top layer→0, Dial's bucket queue) over the live subgraph; per-bucket reachable/unreachable counts + hop min/avg/median/max.
- **`--scc`**: per-layer strongly-connected-component count + largest, via forward-only iterative Tarjan (no reverse adjacency, bounded memory).
- **`--probe-serials <list>`** (`probe_report.{txt,json}`): per-(eye, serial) dossier — reachability + hop, raw vs valid in-neighbor lists, out-neighbors, SCC membership, graph version, pending modifications, and a verdict (REACHABLE / ORPHAN / CUT-OFF / ABSENT).

## Correctness

A serial can have several versions present; only the **live (max-version)** node carries valid edges, and a valid edge requires **both endpoints live**. All per-node state is array-indexed by `VectorId::index()`; reachability/SCC/in-degree are version-strict throughout.

## Validation

Exercised end-to-end against the genesis e2e checkpoints (parties 0/1/2), including a multi-layer graph (M=4) and a multi-version node case — verdicts, layered descent, SCC, and the matrix all check out. Build/fmt/clippy clean. Scale (~17M) performance is reasoned, not measured.


iris-mpc-bins/bin/iris-mpc-cpu/db_sanity_check.rs
```diff
@@ -226,6 +243,404 @@ struct DegreeHistEntry {
...
+    (min, avg, median, max, total)
+}
+
+/// Per-layer adjacency (slices into the graph) and live-key arrays, both indexed
+/// by `VectorId::index()`, plus the array length `n`.
+type LayerArrays<'a> = (Vec<Vec<&'a [VectorId]>>, Vec<Vec<VectorId>>, usize);
+
+/// Fill length-`n` adjacency + live-key arrays (indexed by `VectorId::index()`)
+/// from one layer's links. A serial may have several versions present; only the
+/// most recent (highest `version_id`) is live and carries valid edges, so each
```
bgillesp: In principle "highest version_id is live" should be correct, but might be good to double-check this against the irises table -- both that the highest version represented in the graph is the version in the irises table, and that the versions match between cpu and gpu iris tables (acknowledging that cpu and gpu versions may diverge in a small number of cases since the gpu table is live).

iris-mpc-bins/bin/iris-mpc-cpu/db_sanity_check.rs
```diff
@@ -226,6 +243,404 @@ struct DegreeHistEntry {
...
+        }
+    }
+    (comp_of, sizes)
+}
+
+/// Single-source layered BFS hop/reachability stats, bucketed by serial ID.
+///
+/// Uses one entry point: the top layer (~N/M^2 nodes, M ≥ that) is near-complete,
+/// so a single source reaches all of it within ~1 hop — equivalent to seeding from
+/// the whole entry-point set the search picks from.
```
bgillesp: Not sure if this is conveying things properly as stated -- based on the description, it suggests that layer 2 appears as a nearly-complete graph layer, but in the actual implementation layer 2 does not exist as a graph layer.  Also, N/M^2 is a correct size for the list of entry points only when we haven't overridden the layer density parameter, which we have set to a lower value for production deployment.

My inclination here might be to restructure `layered_bfs` to take a list of source nodes to form the start of the traversal (I think just insert all of them in `reached` and set distances to `0` at the start), then include all entry points in the list for this check.  This then leans on the heuristic that "nearest entry point to a query, which is concretely the entry point used to traverse the graph in HNSNW, is the source of a shortest path".  What do you think about this option as opposed to the current single-entry-point metric?

iris-mpc-bins/bin/iris-mpc-cpu/db_sanity_check.rs
```diff
@@ -236,6 +651,288 @@ impl Stats {
...
+}
+
+/// In-depth per-(eye, serial) reachability dossier for a probed node.
+#[derive(Serialize)]
+struct ProbeReport {
+    eye: String,
+    serial: u32,
+    exists_in_graph: bool,
+    in_irises_table: bool,
+    graph_version: Option<i16>,
```
bgillesp: Might be worth doing a check of the graph version against CPU and GPU postgres versions.

iris-mpc-bins/bin/iris-mpc-cpu/db_sanity_check.rs
```diff
@@ -236,6 +651,288 @@ impl Stats {
...
+            let source_live = si < n && key[0][si] == *node;
+            let reachable = source_live && dist[si] != u32::MAX;
+            for nb in neighbors {
+                let target = nb.serial_id();
+                if !probe_set.contains(&target) {
+                    continue;
+                }
+                let ti = nb.index() as usize;
+                let valid = source_live && ti < n && key[0][ti] == *nb;
+                let pn = ProbeNeighbor {
```
bgillesp, line 833: Would be good to aggregate information on target version as well -- distribution of which version of the probe id is linked to.  Could we add it as an additional field in ProbeNeighbor?

iris-mpc-bins/bin/iris-mpc-cpu/db_sanity_check.rs
```diff
@@ -226,6 +243,404 @@ struct DegreeHistEntry {
...
+        return;
+    }
+    let (adj, key, n) = build_layer_arrays(graph);
+    let dist = match pick_source(graph, &key, n) {
+        Some(src) => layered_bfs(&adj, &key, n, src),
+        None => vec![u32::MAX; n],
+    };
+
+    // Aggregate per bucket: hops over reachable nodes, plus reachable/unreachable split.
+    let num_buckets = (n as u32 / bucket_size) as usize + 1;
```
bgillesp: Current indexing on buckets results in slighly unintuitive bucket distribution -- e.g. only the first `bucket_size - 1` ids go in the first bucket, and an evenly divided graph produces an extra stray bucket (e.g. 1M entries and `bucket_size` 50k produces 21 buckets, last one of which has size 1).

Implemented update to switch to uniformly 0-based indexing for bucket selection in #2245 if it looks good to you.

bgillesp (commented):
Looking good overall, just a few smaller thoughts on some of the details!

iris-mpc-bins/bin/iris-mpc-cpu/db_sanity_check.rs
mcalancea: Yes, Claude seems to find his own justifications for what I ask it to do sometimes. I try to trim these, but some get by. I think initializing the source set with all the entry-points is sensible, I've done it.

bgillesp (approved):
Nice!  Looking good to me now -- thanks for the updates!

PR #2232, opened 2026-06-24
Graph v5: serial-id-only HNSW graph

mcalancea:
Migrates the in-memory HNSW graph (`GraphMem`/`Layer`) to store **`SerialId` (u32)** in neighbor lists instead of full **`VectorId`** (`{id, version}`, 8B after alignment). Edge lists dominate graph memory, so this roughly halves edge storage (~32 GB/eye vs ~64 GB at prod scale).

## Design

**Version resolved at read time.** Edges are version-free serials. `VectorStore::serials_to_vector_ids` resolves each serial to its current `VectorId` (registry for aby3, storage for plaintext) when the searcher expands a neighborhood, so an edge always points at the live version. A serial not currently live resolves to `None` and is dropped — mirroring the old `only_valid_vectors`.

**Staleness via sequence-number clocks.** Each `Neighborhood` carries a `seq_no`; `GraphMem` tracks `last_update_seq_no` and a per-node content clock (bumped only on a node's own (re-)insertion, cleared on removal). An edge is stale when its neighbor's content clock exceeds the neighborhood's certified `seq_no` (`is_active`). One discipline, three anchors:
- at read — `get_active_links` masks stale/removed edges during traversal;
- at write — filter-on-bump drops them before re-stamping a touched neighborhood (`edit_neighborhood`, the sole edit path);
- at apply — each record's edge references are resolved against the record's `as_of` (below).

This replaces main's runtime `prune_invalid_links` / `only_valid_vectors`. `RemoveNode` is a pure node removal (no implicit backlink mutation); dangling backlinks are handled by the read mask + filter-on-bump.

**Consensus.** `GraphMem::checksum` folds both clocks — per-neighborhood `seq_no` via a `(node, seq_no, sorted-neighbors)` set-hash, and the content clock via a maintained `SetHash` — so parties that would skip different edges cannot agree on a stale checksum. Neighbor lists are kept canonical (sorted+deduped) and hashed once per neighborhood.

## WAL: intent records, resolved at every apply

Graph changes are `GraphMutation { seq_no, as_of, ops }` records: `ops` carry the minting intent verbatim, and `as_of` is the graph seq the ops' edge references were identified against (search and neighborhood-ranking APIs return it alongside their results; `InsertPlanV` carries it to the mint).

There is one apply path (`apply_new` = stamp + `insert_apply`). Every apply — first or replayed — resolves the record's references against the current graph at the record's `as_of`: a reference whose target was removed or re-inserted after identification is void and dropped; an op whose `base` is void is dropped whole; `AddEdges` self-references are always dropped (a self-edge is never meaningful — in a minting record it can only be a stale echo of a search against the node's replaced content). This closes the window where an insert identified before an intra-batch reauth would re-point at — and freshness-certify — the target's new content. Intra-batch bootstrap links (added when a searched bottom layer comes back small) are identified at apply time, so they mint as their own follow-up record at the current seq.

Replay reaches the mint's state because resolution is a deterministic function of the replayed history. This makes the resolution predicate part of replay semantics: a `GraphMutationFormat` version pins the wire layout and the abstract mutation semantics, and semantics changes deploy behind a checkpoint barrier, with the format bump as the loud-failure tripwire for stray old segments. Wire: `GraphMutationV1` (bare-serial edge ops + `as_of`); V0 is rejected — the WAL resets at cutover.

## Executable spec

`hnsw/graph/model.rs` (`cfg(test)`) implements the abstract graph in plain set arithmetic — versioned nodes, entry points, eager incident-edge removal, and independent reference resolution from its own node-insertion times. A refinement test drives `GraphMem` and the model with generated record streams (inserts, reauths, deletions, edge touches, compaction-shaped removals, and drifted intents minted with a stale `as_of` after an intervening reauth/deletion; 8 seeds × 300 steps), asserting observation equality after every record plus replay-to-identical-state per stream. Pinned regression tests cover the reauth-window, compaction-window, void-base, and reauth-self-reference cases.

## Serialization / migration

- New `GraphV5` wire format + `GraphFormat::V5` (now `Current`); V0–V5 read paths; streaming reader routes through the derived bincode path.
- **Genesis is the sole legacy migrator.** It prunes a V3/V4 base at read (`read_graph_pair_pruned`): drops edges whose stored version ≠ the iris store's current version (`stream_iris_ids`) and edges to deleted serials (S3 deletion list), then rewrites as V5. `download_graph_checkpoint` refuses V2/V3/V4 for every other consumer; the materializer requires a native V5 checkpoint.

## Cutover

Checkpoint checksums change, so this needs an **all-parties-at-once** cutover: run genesis on all parties from the v3/v4 base → clean v5, restart hawk onto v5 (keep v4 for rollback), `db-sanity-check` after.

## Follow-ups (tracked, not blocking)

- Batch serial→vector resolution once per `open_nodes_batch` (currently per-node) — perf.
- Unify or pin the `GraphMem`/`GraphV5` serializers (byte-identical by convention today) — fragility.
- Extend the graph differ to compare `seq_no` + content clock (currently topology-only) — diagnostics.
- `db_sanity_check`: report active (not raw) degree and support inspecting legacy checkpoints — diagnostics.
- Drop self-edges in the legacy prune (`legacy_prune_to_mem!`) so version-matching self-loops in a V3/V4 base are swept at migration.
- Enforce WAL seq contiguity at replay (`seq_no == last + 1`) so a gap fails loud instead of silently changing what later records resolve against.


iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -224,14 +377,25 @@ impl GraphMem {
                         self.layers.resize(*height, Layer::new());
                     }
                     for layer_idx in 0..*height {
-                        self.layers[layer_idx].insert_node(id, Vec::new());
+                        self.layers[layer_idx].create_node(sid, tick);
+                    }
+                    if let Some(old) = self.node_init_seq_no.insert(sid, seq_no) {
+                        self.node_init_hash
```
bgillesp: Potential adjustment here: nontrivial logic (e.g. update to entry point) is encoded in "remove node" logic, so force updates to remove existing node before adding new one?

iris-mpc-cpu/src/hnsw/graph/mutation.rs
```diff
@@ -1,22 +1,28 @@
...
+/// The only way to turn one into a stamped [`GraphMutation`] is
+/// [`crate::hnsw::GraphMem::apply_new`], which assigns the sequence number and
+/// applies it in a single step.
 #[derive(Clone, Default, Debug, PartialEq, Eq)]
 pub struct UnstampedMutation {
+    /// Sequence number of the graph state the edge references in `ops` were
+    /// identified against; becomes `GraphMutation::as_of`. Must come from the
+    /// graph read that produced the references — a later seq would re-certify
+    /// references against content they never saw.
+    pub as_of: u64,
```
bgillesp, line 25: I think putting `as_of` at the `GraphMutation` level works in the current design, but conceptually the timestamp goes with an individual set of edges, which could be sourced from different graph timestamps, so ideally would be identified at the `AddEdges` and `RemoveEdges` level.  (Perhaps even a separate struct type `NeighborSet` which has an associated timestamp?)

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -297,80 +482,161 @@ impl GraphMem {
...
+    ///
+    /// `AddEdges` additionally drops self-references (`z == base`): in a
+    /// minting record they are stale echoes of a search against the node's
+    /// replaced content, so the own-`AddNode` carve-out must not re-admit
+    /// them. Dropped silently (routine on reauth). `RemoveEdges` keeps
+    /// self-references, so removing an existing self-edge stays expressible.
+    ///
+    /// Node ops pass through untouched. Deterministic in the graph state, so
+    /// replay resolves each record exactly as its mint did.
+    fn resolve_ops(&self, as_of: u64, ops: &[MutationOp]) -> Vec<MutationOp> {
```
bgillesp, line 574: In our WAL interface, I think that the base nodes for the edge ops probably should be interpreted as "current node with specified serial" rather than "node with the specified serial at some timestamp, which could have become stale".  This also goes with the idea of packaging the timestamp with the neighbor data of individual `MutationOp`s; then resolution to the current timestamp is just a local operation filtering the edges over the neighborhood, and doesn't depend on larger structure of the full `GraphMutation`.

sdwoodbury (dismissed):
Approved. Nice work adding `as_of` to address the delete + re-insert scenario. 

 I have some minor requests that could be addressed in this PR, a follow up, or one to merge into this one. 

Background about `seq_no`:
one u64 seq_no wears three names:
  - seq_no (raw on Neighborhood, NodeInit, last_update_seq_no, as_of)
  - Tick (newtype, only at the edit boundary: ~378/520/980/1006, mutation.rs:181/187)
  - "content clock" / content (comment+param name for node_init/NodeInit, e.g. the is_active(content: ...) param at ~107)
  
 Requests:
- rename `Tick` to `StampedSeq`.  (edit_links(), edit_neighborhood(), etc would have their parameter list updated accordingly). 
- within `layered_graph.rs` rename `content` to `node_init` and drop "content clock" from the comments.

Also, in the follow work it would be helpful to trim the doc-comments. Several are long where a line would do, and lean on invented metaphors ("content clock", "certificate", "three anchors"). 

Edit: `StampedSeq` could potentially be something else, like `WalHeight`.

iris-mpc-cpu/src/utils/serialization/graph.rs
```diff
@@ -411,41 +533,45 @@ impl From<graph_v0::Layer> for Layer {
     fn from(value: graph_v0::Layer) -> Self {
         let mut layer = Layer::new();
         for (point_id, edges) in value.links.into_iter() {
-            layer.set_links(
-                point_id.into(),
-                edges.0.into_iter().map(|x| x.0.into()).collect(),
-            );
+            layer.set_links_trusted(point_id.0, edges.0.into_iter().map(|x| x.0 .0).collect(), 0);
```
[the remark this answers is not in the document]
sdwoodbury: but all checks passed

sdwoodbury (approved):

PR #2240, opened 2026-06-26
Quiet & clarify ampc-hnsw startup and batch-processing logs

mcalancea:
Cuts prod log noise during startup and steady-state batch processing; clarifies a few confusing lines.

**Startup**
- `Skip loading s3 retried item` → debug (one WARN/record for a normal chunked-load skip; ~5k/party).
- Load progress logged every 500k records, not 100k.
- Drop duplicate `installed graph from checkpoint` (protocol layer already logs it).

**Batch processing**
- Idle poll chatter → debug (`No messages to poll`; poll-cycle lines when `messages_to_poll==0`; info kept for real batches).
- `Started processing share` (per share) → one `Started processing N shares`; per-share trace-ids at debug.
- `Inserted IN_PROGRESS modification` / `Setting skip_persistence` (per request) → debug, replaced by per-batch summaries in `receive_batch`.
- `Setting mirror attack to true` → `…mirror-attack detection to <bool>`.
- `BATCH_SUMMARY` now carries stage timings (`receive_ms`/`compute_ms`/`persist_ms`); `RESULT_SUMMARY` carries only counts; both tagged with `batch_id`+`hash`. Conflated `total_ms` removed.
- Redundant per-batch detail lines → debug.

Companion: worldcoin/ampc-common#118.

### DIFF
--- a/iris-mpc-cpu/src/execution/hawk_main.rs
+++ b/iris-mpc-cpu/src/execution/hawk_main.rs
@@ -1525,7 +1525,7 @@ impl HawkResult {
             .map(|&d| matches!(d, ReauthUpdate(_)))
             .collect_vec();
 
-        tracing::info!(
+        tracing::debug!(
             "Reauths: {:?}, Unique insert: {:?}, Unique no-match (incl. skip): {:?}",
             successful_reauths,
             unique_insert,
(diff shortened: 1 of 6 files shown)


bgillesp (approved):
Nice improvements here -- I was having a similar experience looking at the logs recently, so I think some refinement of the logging logic is very welcome.

PR #2243, opened 2026-06-27
Collapse graph deserialization + stream/buffer the sidecar checkpoint restore and upload (folds #2244)

mcalancea:
- Collapse graph deserialization onto the derived path; drop the hand-rolled streaming decoders.
- Fold in #2244's streaming-restore speedups (32-way parallel ranged download, decode buffer over the `SyncIoBridge`).
- Repoint the checkpoint `Materializer` onto the streaming decoder. Scoped via a `CheckpointDownload` mode: the **sidecar** cycle streams (peak bounded to download window + graph); **hawk `restart_from_checkpoint` stays `Buffered`** — restore behavior unchanged.
- Buffer the streaming **upload** serializer (`drive_upload`): `bincode::serialize_into` wrote straight into the `SyncIoBridge`, so every small write blocked across the async pipe individually — the same unbuffered slowness #2244 fixed only on the read side. A 1 MiB `BufWriter` (`GRAPH_ENCODE_BUFFER`) coalesces writes; serialized bytes unchanged.

Net effect (dev A/B, ~625 MB checkpoint, 3-party consensus held):

| | before | after |
|---|---|---|
| download + deserialize | ~32 s | ~1.25 s |
| serialize + upload | ~72 s | ~4.8 s |
| streamed-restore peak RSS | ~2.1 GB (buffer-all) | ~1.4 GB (resident graph + fixed ~512 MB window) |
| full cycle compute | ~78 s | ~8.8 s |

At prod scale (~93 GB pair) the upload fix is the larger win — the upload is what *writes* the checkpoint (genesis + checkpointer), and unbuffered that would be hours.

Supersedes #2244.


sdwoodbury (approved):
great work!

PR #2244, opened 2026-06-27
Speed up streaming checkpoint restore

mcalancea:
## What

Make the streaming graph-checkpoint restore path fast. It's used by **Genesis startup**, db-sanity-check, and the sidecar (hawk-main restart already uses the buffered path from #2229). At prod scale (17.4M nodes/eye, M=320, 93.6 GB pair) it took >20 min.

## Profile

Local harness, N=500K nodes/eye, degree 475 (35× → prod, all costs linear):

| stage | local | prod est |
|---|---|---|
| buffered `read_graph_pair` (decode+From) | 2.6s | ~91s |
| streaming via `SyncIoBridge`, **no BufReader** (today) | 34.8s | **~1211s (~20min)** |
| streaming via `SyncIoBridge`, **1 MB BufReader** | 3.0s | ~105s |
| `set_hash` recompute alone (inside From) | 0.55s | ~19s |
| `set_hash` parallelized (rayon) | 0.056s | ~1.9s |

The slowness was **not** the network: the old 19 MB/s was back-pressure from the unbuffered bridge. bincode issues millions of tiny reads, each blocking across the duplex individually.

## Changes

- **BufReader** between the bincode decoder and `SyncIoBridge` (~11×).
- **Bounded parallel-range download**: `range_stream` keeps `DEFAULT_DOWNLOAD_PARALLELISM=32` range GETs in flight via `buffered()` (emitted in offset order), 16 MB ranges → ~512 MB bounded window. Mirrors the buffered path's 32-way fan-out (~1.2 GB/s) but streams in order into the pipe, so the full object never coexists in RAM with the deserialized graph.
- **Parallel `set_hash`**: `Layer::from_links` builds the map then folds the checksum with rayon. Wrapping-add is associative + commutative, so per-chunk partials sum to the **bit-identical** checksum (the two pinned `set_hash` tests pass). `From<graph_v{3,4}::Layer>` use it; this also speeds the buffered hawk-main restart path.

## Expected

Prod restore: >20 min → **download (~78s) overlapped with decode (~80s) ≈ ~80–90s wall**, at bounded ~512 MB instead of holding all 93 GB in RAM. New floor is the serial bincode decode (~65s).

## Validation

- cpu lib builds; serialization + layered_graph + state_check tests green; fmt + clippy clean.
- ⚠️ **Download throughput (the parallel fan-out) is not validatable locally (no S3) — needs a dev/prod-scale run** to confirm it saturates.

### DIFF
--- a/iris-mpc-cpu/src/execution/hawk_main/state_check.rs
+++ b/iris-mpc-cpu/src/execution/hawk_main/state_check.rs
@@ -97,6 +97,13 @@ impl SetHash {
         self.accumulator
     }
 
+    /// Reconstruct from a precomputed accumulator, e.g. the wrapping-sum of
+    /// per-chunk checksums from a parallel fold. Sound because the fold is
+    /// associative and commutative.
+    pub(crate) fn from_accumulator(accumulator: u64) -> Self {
+        Self { accumulator }
+    }
+
     pub(crate) fn hash(value: impl Hash) -> u64 {
         let mut hasher = SipHasher13::default();
         value.hash(&mut hasher);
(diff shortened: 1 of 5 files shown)


mcalancea:
Folded into #2243, which now carries the decoder collapse, this PR's streaming speedups (32-way parallel ranged download + decode buffer), and a sidecar materializer repoint onto the streaming path. Closing as superseded.

PR #2247, opened 2026-07-01
POP-3905: generic retention-reaper CronJob

carlomazzaferro:
## What
A single generic, config-driven retention CronJob — **`retention-reaper`** — for append-only Postgres tables. Runs a bounded, batched, guarded `DELETE`.

Pairs with the **ampc-common** PR (adds `created_at` + index) — land that first.

## The binary (`iris-mpc-bins/bin/retention-reaper`)
Per `RETENTION_JOBS` entry `{table, ts_column, retention, guard, batch_size}`: batched `ctid` delete of rows older than the window (bounded `statement_timeout`, non-zero exit on failure), emitting `retention.{rows_deleted, oldest_retained_seconds, dead_tuple_ratio, last_success, run.duration}`. Identifiers validated + quoted; `guard` is trusted deploy config. One binary serves anon_stats (guard `processed = TRUE`), `modifications`/POP-3931 (terminal-state + watermark guard), and future tables. clippy `-D warnings` clean, unit tests pass.

## Deploy (stage only for now)
`Dockerfile.retention-reaper` + GHCR CI + per-party `common-cronjob` values for **both clusters** (smpcv2 + ampc-hnsw × 3 parties), **stage only** in this PR. Prod manifests land in a follow-up once the monitors are validated in staging. Monitors (`deploy/monitors/…`) refocused on **bloat** (`dead_tuple_ratio`) — the DELETE-approach's real risk.

## ⚠️ Gate (before prod)
ampc-common migration landed + deployed; monitors validated firing in staging; then add prod manifests + enable per party (canary first). Follow-ups: least-privilege retention DB role; batch-size / autovacuum tuning verified at volume.

POP-3905


iris-mpc-bins/bin/retention-reaper/main.rs
```diff
@@ -0,0 +1,369 @@
...
+        serde_json::from_str(&raw).wrap_err("RETENTION_JOBS is not valid JSON")?;
+    if jobs.is_empty() {
+        bail!("RETENTION_JOBS is empty — nothing to do");
+    }
+    Ok(jobs)
+}
+
+/// Postgres identifier guard: unqualified, `[A-Za-z_][A-Za-z0-9_]*`. Rejects anything that
+/// could break out of the interpolated position (dots, quotes, whitespace, etc.).
+fn validate_identifier(ident: &str) -> Result<()> {
```
gryaele, line 314: Could alternatively use https://docs.rs/pg_escape/latest/pg_escape/ to avoid SQL injection attacks

iris-mpc-bins/bin/retention-reaper/main.rs
```diff
@@ -0,0 +1,369 @@
...
+    statement_timeout: Duration,
+    party: &str,
+) -> Result<i64> {
+    let guard = job
+        .guard
+        .as_deref()
+        .map(|g| format!(" AND ({g})"))
+        .unwrap_or_default();
+    // retention is bound as a parameter; identifiers are validated + quoted; batch is i64.
+    let delete_sql = format!(
```
gryaele, line 195: Is it for the anon stats tables?

gryaele (approved):

iris-mpc-bins/bin/retention-reaper/main.rs
carlomazzaferro: primarily, yes. But for the modifications table as well

iris-mpc-bins/bin/retention-reaper/main.rs
carlomazzaferro: fair, but we control the input here and we have an allow-list. should be fine since the range of uses is quite small

PR #2248, opened 2026-07-01
Reduce WAL sidecar flakiness: bound peer-mesh connect, entry barrier, generous round timeout

mcalancea:
## Why

Two flakiness classes in the WAL sidecar (`ampc-hnsw-graph-persistence-job`, daily CronJob, one-shot 3-party ring), both observed on dev:

1. **Wedge:** if a peer never becomes reachable, the underlying dial (`connect_loop` in ampc-common) retries `os error 110` with **no deadline**, and it runs inside `control_channel()` *before* the exchange, so `--peer-round-timeout-secs` never applies. A wedged fire spins until `activeDeadlineSeconds` (observed as a ~13h job).
2. **Desync:** parties reach the first timed exchange at different times (pod scheduling / startup skew); the fast party trips `peer_round_timeout`, exits transient, and the next retry is the next day's fire.

## What

Sidecar-focused, minimal:

- **`make_connections_timeout`** (default **300s**, `--make-connections-timeout-secs`) — wraps `control_channel()` in `sidecar_cycle` with a wall-clock bound. On elapse → `CycleError::Transient` → the one-shot fire exits, and the next scheduled fire retries. Bounds the wedge to 5 min instead of hours. The hawk restart path is intentionally left unbounded — it forms the mesh reliably, and there a spurious timeout would be worse than a rare wait.
- **Entry `barrier()` in `run_cycle`** — the untimed ring rendezvous (already used after materialize and before finalize) now also runs before Phase 1, so `peer_round_timeout` only has to cover in-cycle round latency, not cross-party startup skew. Mesh formation is bounded upstream by the connect timeout, so the untimed barrier can't reintroduce a wedge. Covers both the sidecar and the hawk restart path.
- **Sidecar `peer_round_timeout` default 10s → 120s** (dev deploy values 30 → 120). The sidecar is a daily one-shot and each round's window also contains per-party DB queries; a tight timeout only buys spurious desyncs.

## Validated on dev

Two rounds on the dev graph-persistence-job:

- **Connect-timeout commit** (`1ce5fcd9` test image): 5× consecutive fires → 15/15 clean `NotEnoughMutations` skips across all 3 parties.
- **Full branch** (`df36bd65` test image, barrier + 120s round timeout): 9× consecutive fires, one enrollment batch sent between fires → **27/27 clean party outcomes, zero transients/desyncs/wedges**. 7 runs finalized (heights 69→75, 3-party hash consensus, ~148s each incl. S3 materialize+upload); 2 runs correctly skipped `available: 0` (idle WAL — one batch failed client-side, unrelated). One skip run absorbed ~140s of party startup skew inside mesh formation — the exact regime that previously desynced the ring.

The connect timeout itself was not exercised (no peer was ever unreachable, so nothing wedged — a true wedge is hard to force on demand); its value is confirmed wired, not stress-proven.

## Observability (logs, no code)

No StatsD recorder in the sidecar (short-lived one-shot). Signal rides on existing logs; recommended Datadog monitors on `service:ampc-hnsw-graph-persistence-job env:prod`:

- **Failure rate:** log-based metric on `outcome="transient"`/`"fatal"` (or `"sidecar exited with fatal error"`) — alert on sustained ≥1/day.
- **Wedge / no-op:** alert if zero `"sidecar one-shot complete"` lines in ~25h (a wedged fire logs no completion line).
- **Upload failure:** a finalizing party logs `running terminal action` → `Streaming upload complete: key=…` → `outcome="finalized"`. Alert if `running terminal action` appears without a matching `Streaming upload complete` (S3 upload failed) — distinct from the DB-record failure strings (`insert_genesis_graph_checkpoint`/`commit tx`).

Logs ship as ANSI text, so these are substring matches, not facets.

## Test

`cargo fmt`, `clippy`, `checkpoint_protocol` tests (48) green. Config round-trip test extended for the new field.

## Deploy note

Prod deploy values live in the infra repo; code defaults (300s connect, 120s round) apply everywhere even without arg changes. (The dev deploy for testing was applied on `sw/deploy_main` and will be reverted separately.)


sdwoodbury (dismissed):
I was going back and forth about whether `barrier()` is needed and concluded that `control_channel()` can indeed have one peer return first. and even though at that point all nodes are up, the other the peers could be delayed by at most the max-value of the random reconnect delay, which is currently 3 seconds.

and in Genesis, it calls `control_channel().sync()`. This PR matches existing usage elsewhere. 

good catch!

sdwoodbury (approved):
approving after fixing the cargo-deny and merging with main.

PR #2252, opened 2026-07-02
feat(retention-reaper): dry-run mode + bump ampc-common to e35cb4b

carlomazzaferro:
- Bump the 4 ampc-* deps 34170ed -> e35cb4b (picks up ampc-common #121, the anon_stats created_at migration the reaper's ts_column depends on).
- Add RETENTION__DRY_RUN (bool, default false): when set, each job runs a read-only COUNT over the exact delete predicate (ts < now() - retention [AND guard]) and logs how many rows it WOULD delete, deleting nothing. Distinct metric retention.dry_run.rows_would_delete. Lets the reaper be validated in stage before it removes data. Live path unchanged when unset.


carlomazzaferro:
@codex review

eaypek-tfh (approved):

PR #2256, opened 2026-07-05
graph_v5: resolve mutation intents at stamping (linearization alternative)

mcalancea:
Alternative boundary for reference resolution, against the base branch's resolve-at-every-apply (#2232); one of the two lands. Shared with the base: `as_of` provenance through the search/ranking APIs, the resolution predicate itself (void refs, void base, dropped `AddEdges` self-references), the executable spec, and the pinned regression tests.

**What differs — resolution happens once, at stamping.** `UnstampedMutation { as_of, ops }` is linearized inside `apply_new` before the first apply; the stamped `GraphMutation` carries only the surviving ops, and `as_of` is not persisted (wire records carry no `as_of` field). Replay applies records as recorded, so replayed segments are independent of the resolution predicate: predicate evolution never reinterprets history. The trade: records don't carry their identification context, and linearization must be atomic with the first apply — enforced structurally (`linearize` is private to `apply_new`, same exclusive borrow).

**Oracle.** Because the model consumes resolved records, the refinement test alone cannot observe a broken mint-side filter; the generator's drift arm therefore additionally pins that minted records never reference a serial whose content changed after the intent's `as_of` (mutation-tested: a neutered filter fails both this pin and the pinned unit tests).

348 cpu lib tests green; workspace clippy + fmt clean.

### DIFF
--- a/iris-mpc-cpu/src/execution/hawk_main/insert.rs
+++ b/iris-mpc-cpu/src/execution/hawk_main/insert.rs
@@ -35,6 +35,9 @@ pub struct InsertPlanV<V: VectorStore> {
     pub query: V::QueryRef,
     pub links: Vec<Vec<VectorId>>,
     pub update_ep: UpdateEntryPoint,
+    /// Graph state `links` was identified against (see
+    /// `UnstampedMutation::as_of`); comes from the search that produced them.
+    pub as_of: u64,
 }
 
 // Manual implementation of Clone for InsertPlanV, since derive(Clone) does not
@@ -45,6 +48,7 @@ impl<V: VectorStore> Clone for InsertPlanV<V> {
             query: self.query.clone(),
             links: self.links.clone(),
             update_ep: self.update_ep.clone(),
+            as_of: self.as_of,
         }
     }
 }
@@ -65,8 +69,8 @@ impl<V: VectorStore> Clone for InsertPlanV<V> {
 ///
 /// Returns a parallel pair of `VecRequests`:
 /// - the per-slot `Vec<ConnectPlanV>` carrying the graph mutations to persist (a slot
-///   may produce 0 to 3 mutations from the per-slot steps, and the last non-empty slot may
-///   additionally carry the batch's single global compaction mutation);
+///   may produce several mutations from the per-slot steps, and the last non-empty slot
+///   may additionally carry the batch's single global compaction mutation);
 /// - the per-slot `Option<VectorId>` identifying the newly inserted vector, or `None`
 ///   for pure deletions and no-op slots.
 pub async fn insert<V: VectorStoreMut>(
@@ -96,7 +100,7 @@ pub async fn insert<V: VectorStoreMut>(
     let insert_plans = join_plans(plans);
     validate_ep_updates(&insert_plans, searcher.max_graph_layer)?;
 
-    let mut intra_batch_inserted = vec![];
+    let mut intra_batch_inserted: Vec<VectorId> = vec![];
     let m = searcher.params.get_M(0);
 
     let mut slot_outputs: Vec<Vec<ConnectPlanV>> = vec![vec![]; insert_plans.len()];
@@ -109,6 +113,7 @@ pub async fn insert<V: VectorStoreMut>(
         // (a) Delete first: own GraphMutation with the lower seq_no.
         if let Some(rid) = replace_id {
             let mutation = graph.apply_new(UnstampedMutation {
+                as_of: graph.last_update_seq_no,
                 ops: vec![MutationOp::RemoveNode { id: *rid }],
             })?;
             slot_outputs[idx].push(mutation);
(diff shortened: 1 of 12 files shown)


iris-mpc-cpu/src/execution/hawk_main/insert.rs
bgillesp, line 163: I like this reformulation -- it keeps the early bootstrapping logic separate from the standard operation.

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -576,32 +569,42 @@ impl GraphMem {
...
-    /// Node ops pass through untouched. The resolved ops all satisfy the
-    /// spec preconditions (`model.rs`) at apply time. Deterministic in the
-    /// graph state, so replay resolves each record exactly as its mint did.
-    fn resolve_ops(&self, as_of: u64, ops: &[MutationOp]) -> Vec<MutationOp> {
+    /// `AddEdges` additionally drops self-references (`z == base`): a
+    /// self-edge is never meaningful, and in a minting mutation it can only
+    /// be a stale echo of a search that ran against the node's replaced
+    /// content — the carve-out must not re-admit it. Dropped silently
+    /// (routine on reauth, not a drift signal). `RemoveEdges` keeps
+    /// self-references, so removing an existing self-edge stays expressible.
```
bgillesp, line 596: This broadly of makes sense from an application-logic perspective, but I don't like the extra special case since it extends the semantics of an "AddEdges" operation in a non-obvious way.  If loops are being added to the graph, this is probably a bug that should be surfaced elsewhere.  (In terms of operation logic, this generally is not a problem for Reauths since RemoveNode occurs in a separate `GraphMutation` than the subsequent `AddNode` and `AddEdges`, so any reference to the existing node gets nuked by the updated sequence number on the node.)

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -612,50 +615,59 @@ impl GraphMem {
         let fresh = |z: &SerialId| minted.contains(z) || is_active(&self.node_init, *z, as_of);
 
         let mut dropped = 0usize;
-        // The surviving refs of one op half, or `None` if the op is void.
-        let mut filter_refs = |base: &SerialId, neighbors: &[SerialId]| {
+        // The surviving refs of one op, or `None` if the op is void.
+        let mut filter_refs = |base: &SerialId, neighbors: Vec<SerialId>, keep_self: bool| {
             if !fresh(base) {
```
bgillesp, line 620: I think this may be a logic error for Reauth-type requests.  The `as_of` timestamp for the mutation is set at the time of search, but in the meantime a mutation with a `RemoveNode` mutation op is applied to the serial to invalidate the old version, followed by a separate mutation with an `AddNode` representing the updated version, and a corresponding `AddEdges`.  In this case, the `AddEdges` will use the outdated `as_of` from the original searchto interpret the base serial, which will result in the op being dropped by this code.  I may be being a bit of a broken record here, but I think that the correct semantics of our operations are that "base node serials represent the node as of the current state, while edge sets are historical with associated staleness".

bgillesp: Edit: On my second reading, I see that the `minted` field addresses this particular use case, but it's really addressing this by bringing application logic into the core update logic of the data structure, which seems like a bad design pattern -- and the logic is unintuitive to follow here (at least I missed it on a first pass).  What if the caller decides that they want to move the `AddNode` to a separate mutation for some purpose?  Is there a disciplined reason for why that should be treated differently than the current `GraphMutation` configuration?  My opinion is that anything in the core graph mutation logic that needs to read the relative structure of multiple `MutationOp`s in a mutation in order to process things is going to be fragile, and probably is pushing too much logic into the core.

iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
```diff
@@ -612,50 +615,59 @@ impl GraphMem {
...
+                    layer,
+                    edge_type,
                 }),
             };
             resolved.extend(resolved_op);
         }
         if dropped > 0 {
-            metrics::counter!("graph_resolution_dropped_refs").increment(dropped as u64);
+            metrics::counter!("graph_linearization_dropped_refs").increment(dropped as u64);
             warn!(
```
bgillesp, line 669: It might not make sense to put this at a `warn` level of logging (or to log it at all, potentially?) -- dropping stale neighbor refs should be considered normal behavior, since for instance a Reauth node could be identified as a nearest neighbor of an intra-batch search.

bgillesp (commented):
The logic looks correct to me as written, but I would strongly recommend two concrete revisions.

- Linearization should definitely not treat base nodes as versioned by `seq_no`.  Right now there's tailored application logic to look for `AddNode` in the same mutation as `AddEdges` to identify when the base node shouldn't be considered stale because it was being added in the same mutation, but this is fragile and application-dependent, and doesn't fit naturally into a low-level data structure model.  This recommendation I consider high priority because I believe the current functionality could be a foot-gun for future developers due to the ad-hoc nature of the fix needed to support Reauth and similar request types.

- If we really need the linearization step to modify mutations prior to application and writing to the WAL, then the stamping step should be separated from application to the graph.  In other words, replace `apply_new` with `finalize_mutation` or something like, which does the stamping and the linearization, but does not apply it.  Then `insert_apply` is used uniformly to apply the finalized mutation to the graph.  This separation into an extra step with an informative name warns the caller that something more than adding a timestamp might be happening to the mutation prior to application to the graph.

An additional recommendation I'll put down as simply a rejoinder since I raised it previously: I think that inclusion of the linearization step is sub-optimal here and the design would be cleaner with preserving `as_of` in the WAL.  This kind of a preprocessing transformation could in principle make sense if the operation is expensive to execute (i.e. to cache the intermediate result if it requires MPC, or some DB-scale computation).  But in the current design, I don't see the advantage to implementing this transformation at all.  The `as_of` field truly only applies to the edge sets of the `AddEdges` and `RemoveEdges` mutation op variants, and this data can just be carried along for the ride into the WAL and applied inline via a lean `filter_refs` operation.  As currently implemented, the fact that we are pushing to avoid edge timestamp data getting into the WAL means that we can't put `seq_no` in the edge update mutations, unless you want a fully separate type to represent "processed mutation operations".  This could be done, but it emphasizes that the intermediate representation "post-linearization" adds nontrivial complexity to the design.

PR #2258, opened 2026-07-06
fix: bound empty batch poll wait with opt-in abort (POP-4051)

carlomazzaferro:
## Problem

[POP-4051](https://linear.app/worldcoin/issue/POP-4051/iris-mpc-batch-loop-can-deadlock-silently-liveness-probe-cant-detect): `poll_exact_messages` waits for the cross-party-agreed message target with **no deadline** and only `debug`-level retry logs. A party whose queue never yields the agreed count stalls silently — pods 1/1 Ready, zero log output at info level. This wedged all three stage parties for 2.5h on 2026-06-23 and is the mechanism behind the recurring stage enrollment stalls / orb-tools e2e flake (root-cause analysis on the ticket).

## Change

- **`batch_poll_abort_after_secs` config, default `0` = disabled** — exactly today's behavior when unset. Prod is untouched.
- When enabled and an **empty** poll attempt exceeds the deadline (`msg_counter == 0` — nothing consumed/deleted this attempt), the attempt aborts, the peer-visible sync target (`/batch-sync-state`) is reset, and the loop re-enters batch sync after a jittered 1–3s sleep. Reconvergence with peers that latched the old target rides the existing indefinitely-retrying batch-entries SHA sync.
- **Partial polls never abort**: once any message is processed it is already deleted from SQS and exists only in this attempt's in-memory `batch_query` — aborting would strand it. Past the deadline these log `warn` + `batch_poll_partial_stuck` instead.
- Observability: progress log every 30s during waits (the silence is what let 06-23 hide), `batch_poll_duration` histogram, `batch_poll_aborted` counter, >5 aborts on one batch escalates to `error!` (alertable).
- Stage-only rollout: `SMPC__BATCH_POLL_ABORT_AFTER_SECS: "60"` in stage values for **iris-mpc and ampc-hnsw** (both consume this receive loop).

## Protocol-safety invariants (please verify in review)

1. **Abort ⇒ zero SQS side effects this attempt.** The deadline check lives only in the two empty-receive branches; `msg_counter` increments post-process+delete, and receive-then-fail paths take the existing `Err`/teardown route — so `msg_counter == 0` there proves no message was received in this attempt.
2. **Peer-visible target reset is reconvergent.** Peers recompute `min()` fresh each sync round; a peer already polling the old target completes and waits (loudly) in the batch-entries SHA sync until the aborting party re-forms the identical batch.
3. **No new divergence modes.** In the terminal case (a party's queue never receives the message), today's unbounded wait would also consume the next-arriving message and diverge identically — this change makes that case loud and bounded instead of silent.

`CommonConfig` carries the new field so `check_common_config` **forces the deadline to be identical across parties** (asymmetric deadlines would invite abort oscillation). `#[serde(default)]` keeps an old peer's serialized state parseable on an upgraded party during rolling skew (same pattern as `hnsw_param_ef_search_layers_override`).

## Alternative considered

Metric-first rollout (ship `batch_poll_duration` alone, observe stage distribution, then enable the abort). Compressed into one PR because the abort is config-gated default-off and stage enablement is a one-line values revert; happy to split if preferred.

## Verification

- Runtime abort/reconvergence is not unit-testable without a 3-party SQS harness; first stage deploy is the live probe — `batch_poll_aborted` will show it firing during the known stall episodes.

## Follow-ups (tracked on the ticket, not in this PR)

- Alert on `receive_batch_duration` tail + `batch_poll_aborted` rate (the metric existed; the alert is the gap).
- Loop-aware liveness (batch-id advancement heartbeat).
- Revisit the 10s `batch_sync_polling_timeout_secs` + Err-teardown coupling (restarts healthy peers when one party stalls).


iris-mpc-common/src/config/mod.rs
```diff
@@ -893,6 +907,19 @@ impl From<Config> for CommonConfig {
             batch_polling_timeout_secs,
             sqs_long_poll_wait_time,
             batch_sync_polling_timeout_secs,
+            batch_poll_abort_after_secs,
         }
     }
 }
+
+#[cfg(test)]
```
wojciechsromek: Do we need this test?

iris-mpc/src/services/processors/batch.rs
```diff
@@ -308,8 +341,33 @@ impl<'a> BatchProcessor<'a> {
 
             // Poll the determined number of messages
             if messages_to_poll > 0 {
```
gryaele, line 343: Is messages_to_poll coming from the other parties? I.e. this party is somehow faster than the others? Would the next sync show messages_to_poll = 0 for all the parties? 

The self-unblock behaviour we see on stage: does it self-unblocks when the next request comes in 10m triggered by the e2e tests?

gryaele (dismissed):

iris-mpc/src/services/processors/batch.rs
carlomazzaferro: messages_to_poll = min() across all three parties' self-reported counts (each from its own queue's ApproximateNumberOfMessages, served via /batch-sync-state). So not "this party is faster" — it reported ≥1 at count-read, everyone agreed on the target, then its copy stopped being receivable (and the reasons here can be many - Agent theorizes: fanout lag, visibility-hidden, or approximate overshoot).

After the abort the next sync is effectively 0 via the min(): the aborting party resets its served state and keeps reporting 0 while its copy is invisible; peers that already consumed theirs are parked in the entries-SHA retry loop. When the copy reappears → agree 1 → identical batch → SHA converges. #2259 demonstrates that full cycle locally.

Self-unblock: yes. poll_exact_messages counts any received message toward the target, so the next e2e request (~10 min) ends the wait — and on 06-23 the scheduler was broken (POP-4070), nothing arrived, hence the "permanent" wedge. That's also why the abort only fires at msg_counter == 0: unblocking on an unrelated message can form a divergent batch, and pre-consumption is the only point where bailing is safe.

iris-mpc-common/src/config/mod.rs
carlomazzaferro: will remove 👍

carlomazzaferro:
## Stage verification: the fix caught and bounded a real stall within minutes of deploy

Deployed this branch (`d02eea8b2` images) to all six stage clusters (`smpcv2-{0,1,2}-stage` + `ampc-hnsw-{0,1,2}-stage`) via Argo at ~14:10Z today, with `SMPC__BATCH_POLL_ABORT_AFTER_SECS=60` from the stage values in this PR. Within 7 minutes, live stage traffic (e2e cadence + sporadic manual enrollments) produced every behavior this PR ships, in one window:

**[Datadog — stage logs, 14:00–14:30Z](https://app.datadoghq.com/logs?from_ts=1783346400000&live=false&query=env%3Astage+%28%22aborting+empty+batch+poll%22+OR+%22Batch+poll+still+waiting%22%29&stream_sort=desc&to_ts=1783348200000)** (SMPC org; query: `env:stage ("aborting empty batch poll" OR "Batch poll still waiting")`)

1. **The bounded abort, on the exact 06-23 wedge shape — all three parties at once.** Batch 13 (iris-mpc): all three parties agreed `target=1`, none could receive. All three logged `Batch poll still waiting after 30s with 0 out of 1` at 14:16:33, then all three aborted in lockstep at 14:17:02–03: `aborting empty batch poll attempt after 60s waiting for target 1; re-entering batch sync`. On main this is an unbounded silent wait — precisely the state the whole fleet wedged in for 2.5h on 06-23. With the fix it cost 60 seconds and left a warn + counter.
2. **Reconvergence works under real traffic.** Enrollments kept completing through and after the abort (manual `simulate-signup`: `SIGNUP_UNIQUE_SUCCESSFUL` at 26s/31s; first two post-deploy e2e runs: success/success). No entries-SHA wedge, no teardown, no restarts.
3. **Progress visibility replaces the silence.** Batches 11 (iris-mpc) and 18 (ampc-hnsw — confirming the fix applies to both services via the shared path) logged the 30s wait line and resolved naturally before the deadline — waits that were previously invisible at info level.
4. **The partial-batch guard behaves.** Batch 4 logged `still waiting after 30s with 1 out of 2 messages processed` and correctly did NOT abort (messages already consumed from SQS — abort would strand them).

Together with the deterministic local repro in #2259, this covers the fix end-to-end: stall reproduced (local), abort + reconvergence proven (local + stage), all-party wedge shape bounded (stage, real traffic). Follow-ups tracked on POP-4051: monitor on `batch_poll_aborted` rate / `receive_batch_duration` tail, loop-aware liveness.

iris-mpc/src/services/processors/batch.rs
```diff
@@ -394,7 +457,34 @@ impl<'a> BatchProcessor<'a> {
...
                     "Stopping batch receive during polling exact messages due to shutdown signal..."
                 );
-                return Ok(()); // Exit if shutdown is signaled
+                return Ok(PollOutcome::Completed); // Exit if shutdown is signaled
+            }
+
+            if last_progress_log.elapsed() >= BATCH_POLL_PROGRESS_LOG_INTERVAL {
+                let elapsed_secs = poll_start.elapsed().as_secs();
+                if self.config.batch_poll_abort_after_secs > 0
+                    && poll_start.elapsed()
```
dkales: nit: can we not use the above elapsed secs here? (I guess this way is more accurate since you actually compare nanos as well)

iris-mpc/src/services/processors/batch.rs
```diff
@@ -419,6 +509,26 @@ impl<'a> BatchProcessor<'a> {
                         self.msg_counter,
                         num_to_poll
                     );
+                    // The abort deadline check must stay only in empty-receive branches.
+                    // With msg_counter == 0, no message was processed and deleted in this
+                    // attempt; process_message increments only after successful processing
+                    // and delete, while receive/process failures return Err and tear down.
+                    // Never abort with msg_counter > 0 because those deleted messages exist
+                    // only in this processor's in-memory batch_query.
```
dkales, line 517: IIRC, this was also fixed by Philipp in the perpetual resharing PR. Maybe this should just be a backport to main, so deleted messages are not lost in this way? (Also the system is just less resilient this way, as it could die of other reasons as well, without a graceful shutdown).

dkales (dismissed):

iris-mpc/src/services/processors/batch.rs
carlomazzaferro: 777ab7d56,  kept the `Duration` comparison

iris-mpc/src/services/processors/batch.rs
carlomazzaferro: as discussed per slack: those fixes are included, waiting on the results from this PR before re-designing anything

dkales (approved):

PR #2259, opened 2026-07-06
test: local 3-party repro harness proving the POP-4051 stall and the bounded-abort fix

carlomazzaferro:
Stacked on #2258. This PR adds a deterministic local reproduction of the silent batch-poll stall ([POP-4051](https://linear.app/worldcoin/issue/POP-4051/iris-mpc-batch-loop-can-deadlock-silently-liveness-probe-cant-detect)) and records the evidence that #2258's abort semantics work end-to-end on the real 3-party protocol — so review of the fix doesn't have to rest on code reasoning alone.

## What it proves (observed locally, image built from #2258's branch)

**1. Baseline — fix disabled (default): the stall reproduces on the first attempt.**
Party 0 enters `poll_exact_messages` with `target=1` latched and an empty queue and goes fully silent — its only log lines during the stall are healthcheck 200s. This is the exact signature of the 2026-06-23 stage wedge (all parties 1/1 Ready, logs quiet). The peer meanwhile shows the loud half:

```
Batch ID: 5. Finished polling SQS. Processed 1 messages for this batch attempt (target: 1).
Batch sync entries attempt 1 failed after 23.2s: Timeout waiting for party
http://hawk_participant_0:3000/batch-sync-entries to reach batch hash 97a05d20
(last observed: peer batch_hash b6da345d differs from own 97a05d20). Retrying...
```

Releasing the stolen queue copy (`ChangeMessageVisibility → 0`) recovers the batch within ~2s — the freezes-then-continues behavior seen on stage.

**2. Fix — `SMPC__BATCH_POLL_ABORT_AFTER_SECS=10` via the compose overlay: the wait is bounded and reconvergence works.**

```
Batch ID: 1. aborting empty batch poll attempt after 11s waiting for target 1; re-entering batch sync
...release stolen copy...
Batch ID: 1. Finished polling SQS. Processed 1 messages for this batch attempt (target: 1).
```

After the abort, party 0 resets its peer-visible target, re-agrees, consumes the released copy, forms the **identical** batch, the entries-SHA sync converges, and the batch completes. That exercises all three invariants #2258 asks reviewers to verify: empty-abort-only, peer-visible reset, SHA-sync reconvergence.

**3. Incidental: the peer-teardown coupling, observed live.** An earlier harness iteration paused one party >10s (`batch_sync_polling_timeout_secs`); both *healthy* peers exited (`Server exited normally`) via the batch-sync `Err` path. That's the follow-up item on the ticket demonstrated in miniature — one stalled party takes down its healthy peers.

## How the repro works (why it's deterministic)

Racing an external queue-steal against the party's own `ReceiveMessage` loses (localstack delivers in ~13ms). Instead:

1. `docker pause` party 1 — party 0's next sync round now blocks retrying peer-state fetches (a seconds-wide window).
2. Send one uniqueness request via the `client` bin (**unique `--rng-seed` per send** — the FIFO topic/queues use content-based deduplication and silently drop identical payloads for 5 min).
3. Wait for party 0 to log its latched `messages_to_poll=1` (it's now blocked in peer-sync, so the steal can't lose a race).
4. Steal party 0's queue copy with an external `ReceiveMessage` (600s visibility).
5. Unpause party 1 within 8s (under the 10s sync timeout — see finding 3 for what happens otherwise).

Party 0 completes the sync round with the target latched and an empty queue → the stall, every time.

## Run it

```bash
docker build -f Dockerfile.dev.hawk -t hawk-server-local-build:latest .

# Baseline (expect STALL_REPRODUCED, then RECOVERED after release):
docker compose -f docker-compose.hawk.yaml up -d
./scripts/tools/pop4051-repro.sh

# Fix validation (expect ABORT_OBSERVED, then RECOVERED):
ABORT_SECS=10 docker compose -f docker-compose.hawk.yaml -f docker-compose.pop4051-repro.yaml up -d
EXPECT_ABORT=1 ./scripts/tools/pop4051-repro.sh
```

Manual harness, not wired into CI. Happy to fold it into a CI job later if we want regression coverage, but that's scope for a follow-up.

### DIFF
--- a/docker-compose.pop4051-repro.yaml
+++ b/docker-compose.pop4051-repro.yaml
@@ -1,0 +1,13 @@
+# POP-4051 repro overlay: enables the bounded batch-poll abort on all parties.
+# Baseline stall repro: omit this file (or ABORT_SECS=0).
+# Fix validation:      ABORT_SECS=10 docker compose -f docker-compose.hawk.yaml -f docker-compose.pop4051-repro.yaml up
+services:
+  hawk_participant_0:
+    environment:
+      SMPC__BATCH_POLL_ABORT_AFTER_SECS: "${ABORT_SECS:-0}"
+  hawk_participant_1:
+    environment:
+      SMPC__BATCH_POLL_ABORT_AFTER_SECS: "${ABORT_SECS:-0}"
+  hawk_participant_2:
+    environment:
+      SMPC__BATCH_POLL_ABORT_AFTER_SECS: "${ABORT_SECS:-0}"
--- a/scripts/tools/pop4051-repro.sh
+++ b/scripts/tools/pop4051-repro.sh
@@ -1,0 +1,186 @@
+#!/usr/bin/env bash
+# POP-4051 local repro: force one party into the unbounded poll_exact_messages
+# wait (the silent batch-loop stall) against the local 3-party hawk stack.
+#
+# Mechanism (deterministic, no log-races):
+#   1. All parties idle, queues empty.
+#   2. Pause party 1. Party 0's next sync round will block retrying peer-state
+#      fetches against the paused party — a seconds-wide window.
+#   3. Send one uniqueness request (unique --rng-seed to defeat FIFO
+#      content-based deduplication). Copies land on all three queues.
+#   4. Party 0 reads its own count (1) and latches messages_to_poll=1 in its
+#      peer-visible sticky state, then blocks on party 1's sync endpoint.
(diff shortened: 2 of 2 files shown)


carlomazzaferro:
Closing: this evidence PR documented the local repro of the unbounded-poll stall and validated the #2258 abort fix, which regressed on stage and was superseded by the DB-backed ingest redesign (#2264, merged; released in v0.32.6). The repro methodology lives on in the PR description and in scripts/tools/db-ingest-poison-e2e.sh on main.

PR #2262, opened 2026-07-07
feat: modifications retention via guarded scheduled DELETE (POP-3931)

carlomazzaferro:
## POP-3931 — `modifications` retention via guarded scheduled DELETE (30d age)

Implements [POP-3931](https://linear.app/worldcoin/issue/POP-3931) by reusing the POP-3905 retention-reaper (#2247): a migration + deploy config, plus two small operational hardenings to the binary. **Deploys disabled** (`RETENTION__ENABLED` unset → no-op); enabling is a separate, gated step (see rollout below).

### Scope

`modifications` grows unbounded in **both** systems — the GPU MPCV2 server and the hawk server are the same shared server binary (`iris-mpc-hawk` → `server_main`; writes via `insert_modification`, `services/processors/batch.rs:1327`), each into its own DB, and nothing deletes either copy in steady state. This PR's deploy config reaps the **HNSW (hawk) DB copy** first; the **MPCV2 copy's instance follows shortly after** (same binary and guard — values + a cluster-apps entry).

### Migration

- `20260707090000` — `ADD COLUMN created_at TIMESTAMPTZ NOT NULL DEFAULT now()`: metadata-only on PG11+ (empirically proven on PG16.14 for the identical anon_stats migration in ampc-common#121 — relfilenode unchanged). Existing rows backfill to migration time → nothing is deletable until 30d later (conservative; expect `rows_deleted=0` for the first 30 days).
- The index is a **plain transactional `CREATE INDEX`** — deliberately not `CONCURRENTLY`: sqlx serializes migrators via a session advisory lock, and `CONCURRENTLY` waits on every other snapshot-holding transaction, so any concurrent migration of one database (parallel e2e tests, multi-pod boots) deadlocks (`40P01` — caught by cpu-hawk-e2e in this PR's own CI, run 28879180378). The plain build briefly SHARE-locks writers at deploy; acceptable at current `modifications` volume.
- PK / `UNIQUE(id)` untouched — `id` is the globally-unique cross-party key; the ticket documents why partitioning is ruled out.
- The shared migrations dir applies to both DBs' schemas, so the MPCV2 copy's 30-day age clock starts ticking at merge, ahead of its follow-up instance.

### Guard anatomy (deploy config, stage values)

The binary always applies `created_at < now() - '30 days'::interval`; the configured guard ANDs on top:

```
status = 'COMPLETED' AND persisted            -- terminal, fully-persisted rows only
AND id < (SELECT COALESCE(MIN(id), 0) FROM   -- always retain the newest 10k rows
          (SELECT id FROM modifications ORDER BY id DESC LIMIT 10000) newest)
```

The newest-10k floor is same-DB and free, and it covers the two real cross-cutting concerns: the cross-party startup sync compares the newest `max_modifications_lookback` rows by id (`ORDER BY id DESC LIMIT`, default `(100+64)*2 = 328` — ~30× margin), and the `MAX(id)+1` id-assignment trigger must never observe an emptied table (id reuse). It also makes deletion on a very quiet party (fewer than 10k rows in total) a structural no-op, which is the right failure direction.

### Binary changes (operational hardening only, generic)

- `batch_pause_ms` (default 100) between full delete batches — the DELETE otherwise competes with `modifications`' `LOCK TABLE ... IN EXCLUSIVE MODE` id-assignment trigger and could starve the hot write path for the whole drain.
- Health probes (`oldest_retained_seconds`, `dead_tuple_ratio`) now run on the statement_timeout-bounded connection with a client-side timeout — a hung probe under `concurrencyPolicy: Forbid` would silently block every subsequent run — and the bloat probe is pinned to `current_schema()` (same relname can exist in several schemas of one DB).

### Deployment (stage-only, disabled)

Runs on the **ampc-hnsw clusters, ns `ampc-hnsw`** against `DATABASE_AURORA_HNSW_URL`, with the hawk service's own pod SG (in-region). Schedule 04:00 UTC (offset from the anon-stats reaper). Registered via wf-cluster-apps (companion PR — **merge this PR first**, the bootstrap's valueFiles must exist on main).

### Rollout / enable path (not in this PR)

1. Merge → build image → bump the values pin (currently the POP-3905 `d9a3d333…` build).
2. `RETENTION__ENABLED=true` + `RETENTION__DRY_RUN=true` on stage → verify would-delete counts + health gauges per party.
3. Flip `DRY_RUN` off, one party canary, then the rest. Prod after the stage gate, as with POP-3905.
4. **Follow-up (shortly after): the MPCV2 copy's instance** — same guard, values + cluster-apps entry (either on the hnsw clusters via `DATABASE_AURORA_MPCV2_URL`, or on the smpcv2 clusters next to the iris-mpc ns secret — deploy-locality call for that PR).

### Verification

- clippy `-D warnings` clean; 5/5 unit tests.
- Deploy values validated end-to-end: YAML folded scalar → JSON → the actual binary's config loader with the exact per-cluster env (parses, kill switch honored, exit 0).
- Migration lock behavior + probe boundedness reviewed cross-model (GPT-5.5); bounded probes + `current_schema()` pin incorporated. Its `CONCURRENTLY` suggestion was tried and **empirically refuted by CI** (advisory-lock deadlock above) — reverted to the transactional build.

Follow-ups: MPCV2-copy instance (above); update POP-3931 acceptance criteria (watermark predicate retired); `activeDeadlineSeconds` backstop on the CronJob (chart support to confirm); consider renaming `deploy/monitors/anon-stats-retention-monitors.md` now that it spans both tickets.


carlomazzaferro:
@codex review

migrations/20260707090000_modifications_add_created_at.up.sql
```diff
@@ -0,0 +1,20 @@
...
+-- database (parallel e2e tests; multi-pod boots) that pairing deadlocks (40P01, proven
+-- in cpu-hawk-e2e CI). The plain build takes a SHARE lock for its duration, briefly
+-- pausing writers — acceptable at current modifications volume, and the column ADD
+-- itself is instant.
+--
+-- Deliberately does NOT touch the PK/UNIQUE(id) — `id` is a globally-unique
+-- cross-party key (sync grouping key, ON CONFLICT target); see POP-3931 for why
+-- partitioning on created_at is ruled out.
+ALTER TABLE modifications ADD COLUMN created_at TIMESTAMPTZ NOT NULL DEFAULT now();
+CREATE INDEX IF NOT EXISTS idx_modifications_created_at ON modifications (created_at);
```
eaypek-tfh, line 20: small q: do we need `created_at` column as we have have the modification `id` column which is a serially incrementing number?

eaypek-tfh (approved):

migrations/20260707090000_modifications_add_created_at.up.sql
carlomazzaferro: answered in slack - consistency of interfaces

PR #2264, opened 2026-07-07
poc: DB-backed ingest + batch formation — the POP-4051 overhaul

carlomazzaferro:
Redesign (Daniel's sketch): decouple SQS consumption from batch formation entirely, moving request durability into Postgres. **Direction + design review — HNSW/stage-scoped, not a merge for prod.**

## Why this direction

The 2026-07-07 stage wedge (batch 92) came from the #2258 abort re-agreeing a poll target while a peer held partial progress — a bug in the *mitigation*, possible only because formation and SQS consumption are entangled (parties count approximate queue depths, commit to targets, and mutate queue state while agreeing). **In this design that bug class is inexpressible**: there is no mid-batch target to re-agree and no partially-consumed queue state — every request is durable in Postgres before formation looks at it. The original POP-4051 stall (unbounded waits on phantom `ApproximateNumberOfMessages` targets) also disappears: formation counts `SELECT count(*)` over rows it already holds and never waits on SQS.

## Architecture

1. **Ingest task** (independent tokio loop, registered on the shared `TaskMonitor` so a panic tears the process down loudly instead of silently starving the fleet): SQS receive → `INSERT INTO ingested_requests` (**PK = SNS FIFO sequence number**, `ON CONFLICT DO NOTHING`) → delete from SQS. At-least-once delivery becomes exactly-once ingestion; the DB row is then the only copy. Bounded backoff+jitter on transient errors.
2. **Batch formation** never touches SQS: own sync-state = count of pending DB rows; the existing min()-agreement over the existing `BatchSyncState`; batch = first N pending rows `ORDER BY sequence_number`, claimed, fed through the **existing** message-processing path (reauth/deletion/reset/recovery inherit automatically).
3. **Two-phase claim/persist**: formation *claims* rows (`consumed_batch_id`, an opaque non-NULL marker); the batch results transaction — the same tx that commits irises/graph/modifications — *persist-marks* exactly those sequence numbers atomically. A row is "done" iff its batch committed, by construction. (`consumed_batch_id` is deliberately not a stable id: batch ids restart per boot and advance under prefetch — proven live when an interim batch-id-keyed persist-mark marked zero rows. Correlation is by sequence number.)

## Sequence numbers

SNS assigns one per message, identical across all three parties' queues (single `MessageGroupId` — verified against every iris-mpc-bound publisher in signup-service: `sns_dispatcher.go:275`, `recovery/update/processor.go:197`, `reset/update/processor.go:212`). So each party's pending set is a **contiguous prefix of one global order** → "first N by sequence number" produces **identical batches across parties by construction**; the entries-SHA sync degrades to defense-in-depth. Stored fixed-width zero-padded TEXT (lexicographic order = numeric). Production guardrail owed: enforce/monitor the single-group invariant (a future publisher with a different group id reintroduces holes).

## Failure handling (built + locally proven)

- **Boot recovery + fleet-frontier skip-ahead** — closes the asymmetric-commit divergence (C2). If one party's results tx fails while peers commit, a purely-local "release my unpersisted claims" decision would re-form rows the peers already persisted → permanent divergence (locally indistinguishable from a symmetric crash — only the peers know). Fix, the DB analog of the old SQS startup queue-trim: each party's persisted set is a contiguous prefix, so one number (`max_persisted_sequence_number`) fully describes its progress; parties exchange it in `SyncState` (serde-default, rolling-deploy safe). Boot marks rows ≤ fleet frontier persisted (their effects arrive via the existing modifications roll-forward), then releases the rest (symmetric case). Manufactured that exact wedge state locally → heals in one boot.
- **Poison quarantine, two layers** — a message that can't be processed must not wedge the FIFO group head forever. Envelope-parse/normalize failures are caught at **ingest** (delete + `db_ingest_poison_messages`, never insert); type/payload-parse failures at **formation** (`is_content_poison`, exhaustive match — a new error variant is a compile-time classification decision; persist-marked, never re-formed). Only content-determined failures (identical bytes on all parties ⇒ symmetric) quarantine; SQS/S3/DB errors stay fatal.
- **Dual-consumer guard** — pending predicate is `consumed_batch_id IS NULL AND persisted_at IS NULL`, closing the rolling-deploy pod-overlap window where a new pod's boot recovery could resurrect a row an old pod's in-flight tx persist-marks.
- **Startup validations (fail-fast)** — the flag refuses to start with `disable_persistence` (claims would re-form forever), with `enable_modifications_sync` off (the frontier skip-ahead would reconcile bookkeeping whose *data* never arrives — silent divergence instead of a loud wedge), with out-of-range `db_ingest_sqs_wait_secs`, and **on the GPU binary** (no ingest wiring there — flag-on would silently starve the fleet). The startup queue-trim + head-peek are skipped under the flag (the head-peek hard-errors on a corrupt message = boot crashloop).
- **Shutdown-safe ingest task** — `TaskMonitor::check_tasks()` panics on any finished task (even `Ok`), so the ingest loop parks on `pending()` after the shutdown signal instead of returning; graceful drain verified live (SIGTERM → exit 0, no panic). The claim CAS also carries the `persisted_at IS NULL` guard, so the loser of a dual-consumer race fails loud instead of re-claiming a persisted row.

## Verified

- **Unit/db** — `fmt` · `clippy --workspace --all-features --all-targets -D warnings` · `cargo test --lib` (15: sizing rules + legacy/db parity, backoff, seq normalize/order/reject, poison classification exhaustive, CommonConfig equality) · `db_dependent` (idempotency, two-phase claim/release/persist, mixed-state frontier skip-ahead, double-claim compare-and-swap).
- **Local 3-party stack (HNSW)** — happy path + load (identical row-set MD5 ×3), pause-disruption, `kill -9` chaos ×6, 45s DB outage (zero message loss), manufactured asymmetric-commit (C2 heals in one boot), and a repeatable poison harness (`scripts/tools/db-ingest-poison-e2e.sh`: both quarantine layers, all-poison batch still persist-marks, deletion through the stored-body path, batch identity held — asserted on the authoritative DB end-state).

## Scope / still owed before prod

**Fleet scope**
- **HNSW fleet only.** The GPU main shares the formation code but has no ingest wiring (no ingest task, no boot recovery, no persist-mark hook in its results path). A startup bail now refuses the flag on the GPU binary; the wiring itself is the mirror-image follow-up.
- **Two-flag rollout split** (`db_ingest_enabled` → soak → `db_backed_formation`) for a drain-safe switchover; one flag controls both today (coordinated all-party flip, no canary — the flag is in the CommonConfig hash so mixed modes can't run).
- **Kill-switch rollback is lossy** (flag ON→OFF strands pending DB rows) — runbook: drain to zero pending before flipping off.

**Robustness backstops**
- **DLQ-equivalent backstop** for content failures *outside* the enumerated poison set (a deterministic panic in row processing still crashloops); attempts-counter on claim-release, quarantine above threshold. Related: `db_ingest_claims_released` re-releasing the *same* rows across consecutive boots is that loop's signature — needs a monitor.
- **Envelope-strictness blast radius**: a subscription misconfig (raw message delivery toggled, attribute-less publish) makes *every* message fail the envelope parse → quarantine-deleted with only a counter, fleet green. Needs a rate alert on `db_ingest_poison_messages` (and consider dead-lettering bodies to a table/S3 instead of delete-only).
- **Single-writer enforcement**: the claim CAS + pending-predicate guards make the dual-consumer race (rolling-deploy pod overlap) fail loud; a Postgres advisory lock would degrade it to a clean wait instead of teardown noise.
- **Queue redrive policy**: party request queues must NOT have DLQ redrive under this flag — per-party redrive punches per-party holes in the ingested sequence → divergence. Needs an infra check + statement in the deploy docs.
- **Single-group invariant enforcement**: publisher-side documentation/CI check that everything stays in one `MessageGroupId`, plus runtime monotonic-arrival detection at ingest (out-of-order arrival = multi-group alarm).

**Recovery envelope**
- **Frontier exchange rides boot-snapshot SyncStates** — sound only under the full-fleet-restart invariant (unready-gate + heartbeat teardown). If solo restarts are ever allowed, the frontier must come from a live DB read in the sync endpoint (documented at the skip-ahead site).
- **Modifications-lookback window (~328) is recovery's outer bound**: a party down long enough for peers to advance past the window can't backfill data — genesis-restore territory (existing main limitation; the skip-ahead now also depends on it, hence the `enable_modifications_sync` bail).
- **Boot-time ingest-skew visibility**: the old queue trim doubled as a cross-party frontier check; the persisted-frontier exchange replaces the persist half, but an *ingest* gap still only surfaces at entries-SHA time. Cheap: log/gauge own max-ingested sequence at boot.

**Operations**
- **`ingested_requests` retention** (unbounded growth; boot recovery + skip-ahead scans scale with the unreaped table; + late-redelivery landmine if reaped inside SQS's 14-day window — any reaper must reason against message retention).
- **Migration discipline**: `20260616000002` has been edited in place while unmerged/undeployed (fine — no environment holds its checksum); once ANY environment applies it, further changes must ship as new migrations or boots crashloop on sqlx checksum mismatch.
- **Parser-skew across binary versions** — a request-schema change during a rolling deploy makes new-binary parties quarantine a row old-binary parties include → wedge; candidate mitigation: schema-version in the CommonConfig hash.
- Metrics → Datadog monitors (`db_ingest_*` counters emitted; dashboards/alerts owed: poison rate, claims-released trend, persist-mark mismatch, pending-rows gauge).
- Naming: `persisted_at` → `finalized_at` rename offered in review (it now has three writers: results tx, frontier skip-ahead, poison quarantine) — pending decision.

Observability companion #2265 (batch-poll wait progress + duration metric) merged to main.

### DIFF
--- a/iris-mpc-bins/bin/iris-mpc/server.rs
+++ b/iris-mpc-bins/bin/iris-mpc/server.rs
@@ -265,6 +265,8 @@ async fn server_main(config: Config) -> Result<()> {
         next_sns_sequence_num: next_sns_seq_number_future.await?,
         common_config: CommonConfig::from(config.clone()),
         graph_mutation_bytes: vec![],
+        // GPU path has no DB-backed ingest wiring.
+        max_persisted_sequence_number: None,
     };
 
     tracing::info!("Sync state: {:?}", my_state);
--- a/iris-mpc-bins/bin/iris-mpc/server.rs
+++ b/iris-mpc-bins/bin/iris-mpc/server.rs
@@ -460,6 +460,7 @@ async fn server_main(config: Config) -> Result<()> {
     let _result_sender_abort = background_tasks.spawn(async move {
         while let Some(ServerJobResult {
             merged_results,
+            sqs_sequence_numbers: _,
             request_ids,
             request_types,
             metadata,
--- a/iris-mpc-bins/bin/iris-mpc/server.rs
+++ b/iris-mpc-bins/bin/iris-mpc/server.rs
@@ -138,6 +138,15 @@ async fn main() -> Result<()> {
 }
 
 async fn server_main(config: Config) -> Result<()> {
+    if config.db_backed_ingest {
+        bail!(
+            "db_backed_ingest=true is not supported by the GPU server: it has no ingest task, \
+             boot recovery, or persist-mark hook — formation would read an empty/stale \
+             ingested_requests table and silently starve the fleet. Wire the GPU path first \
+             (see PR #2264 production gaps) or unset SMPC__DB_BACKED_INGEST."
+        );
+    }
+
     let shutdown_handler = Arc::new(ShutdownHandler::new(
         config.shutdown_last_results_sync_timeout_secs,
     ));
--- a/iris-mpc-bins/bin/iris-mpc/server.rs
+++ b/iris-mpc-bins/bin/iris-mpc/server.rs
@@ -21,7 +21,7 @@ use eyre::{bail, eyre, Context, Report, Result};
 use futures::{stream::BoxStream, StreamExt};
 use iris_mpc::services::aws::clients::AwsClients;
 use iris_mpc::services::init::initialize_chacha_seeds;
-use iris_mpc::services::processors::batch::receive_batch_stream;
+use iris_mpc::services::processors::batch::{receive_batch_stream, spawn_db_backed_ingest_task};
 use iris_mpc::services::processors::get_iris_shares_parse_task;
 use iris_mpc::services::processors::modifications_sync::{
     send_last_modifications_to_sns, sync_modifications,
@@ -138,12 +138,26 @@ async fn main() -> Result<()> {
 }
 
 async fn server_main(config: Config) -> Result<()> {
-    if config.db_backed_ingest {
+    if config.db_backed_ingest && config.disable_persistence {
         bail!(
-            "db_backed_ingest=true is not supported by the GPU server: it has no ingest task, \
-             boot recovery, or persist-mark hook — formation would read an empty/stale \
-             ingested_requests table and silently starve the fleet. Wire the GPU path first \
-             (see PR #2264 production gaps) or unset SMPC__DB_BACKED_INGEST."
+            "db_backed_ingest=true is incompatible with disable_persistence=true: ingested request claims would never be marked persisted and would be re-formed on every restart"
+        );
+    }
+
+    if config.db_backed_ingest && !(0..=20).contains(&config.db_ingest_sqs_wait_secs) {
+        bail!(
+            "db_ingest_sqs_wait_secs must be within SQS's accepted 0..=20 range, got {}: every receive would fail and ingest would idle with warn-only logs",
+            config.db_ingest_sqs_wait_secs
+        );
+    }
+
+    if config.db_backed_ingest && !config.enable_modifications_sync {
+        bail!(
+            "db_backed_ingest=true requires enable_modifications_sync=true: the boot-time \
+             fleet-frontier skip-ahead marks rows persisted whose EFFECTS only reach this \
+             party via the modifications roll-forward. With sync disabled, an asymmetric-commit \
+             recovery would reconcile bookkeeping without the data — silent cross-party data \
+             divergence instead of a loud wedge."
         );
     }
 
@@ -173,11 +187,21 @@ async fn server_main(config: Config) -> Result<()> {
 
     tracing::info!("Initialising AWS services");
     let aws_clients = AwsClients::new(&config.clone()).await?;
-    let next_sns_seq_number_future = get_next_sns_seq_num(
-        &aws_clients.sqs_client,
-        &config.requests_queue_url,
-        config.sqs_sync_long_poll_seconds,
-    );
+    // Under DB-backed ingest the queue head is party-local state (see the
+    // hawk main): the peek's only consumer, the startup queue trim, is skipped,
+    // and the peek itself hard-errors on a corrupt head message.
+    let next_sns_seq_number_future = async {
+        if config.db_backed_ingest {
+            Ok(None)
+        } else {
+            get_next_sns_seq_num(
+                &aws_clients.sqs_client,
+                &config.requests_queue_url,
+                config.sqs_sync_long_poll_seconds,
+            )
+            .await
+        }
+    };
 
     let shares_encryption_key_pair = match SharesEncryptionKeyPairs::from_storage(
         aws_clients.secrets_manager_client.clone(),
@@ -274,8 +298,11 @@ async fn server_main(config: Config) -> Result<()> {
         next_sns_sequence_num: next_sns_seq_number_future.await?,
         common_config: CommonConfig::from(config.clone()),
         graph_mutation_bytes: vec![],
-        // GPU path has no DB-backed ingest wiring.
-        max_persisted_sequence_number: None,
+        max_persisted_sequence_number: if config.db_backed_ingest {
+            store.max_persisted_sequence_number().await?
+        } else {
+            None
+        },
     };
 
     tracing::info!("Sync state: {:?}", my_state);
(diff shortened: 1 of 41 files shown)


iris-mpc-common/src/config/mod.rs
```diff
@@ -733,6 +761,12 @@ pub struct CommonConfig {
     batch_polling_timeout_secs: i32,
     sqs_long_poll_wait_time: usize,
     batch_sync_polling_timeout_secs: u64,
+    // serde(default) keeps version skew deserializable: an old peer's serialized
+    // state (no such key) must still parse on an upgraded party. Value equality is
+    // then enforced by check_common_config — parties must not run in mixed
+    // ingestion modes.
+    #[serde(default = "default_db_backed_ingest")]
+    db_backed_ingest: bool,
```
dkales: is this intended (mixing node versions) or is this just a safeguard claude added?

iris-mpc-store/src/lib.rs
```diff
@@ -486,6 +502,89 @@ WHERE id = $1;
         Ok(id.0.unwrap_or(0) as usize)
     }
 
+    pub async fn insert_ingested_request(
```
dkales, line 512: could use some docstring on what the Result<bool> means

iris-mpc/src/services/processors/batch.rs
```diff
@@ -449,6 +612,80 @@ impl<'a> BatchProcessor<'a> {
         Ok(())
     }
 
+    async fn process_pending_ingested_requests(
+        &mut self,
+        num_to_poll: u32,
+    ) -> Result<(), ReceiveRequestError> {
+        let current_batch_id = self.current_batch_id_atomic.load(Ordering::SeqCst);
```
dkales, line 749: This is not globally unique since it is reset on startup, but it is not used for anything anyways other than it being not NULL.

carlomazzaferro: Yep and that's key. The final shape: `consumed_batch_id` is just an opaque non-NULL claim marker; persist-marking is keyed by the claimed sequence numbers (1266a3191) and boot recovery releases unpersisted claims (003ab13c3). Comment added at the claim site in f47cb3f79. Happy to rename to `claimed_batch_id` if the name misleads.

iris-mpc-store/src/lib.rs
carlomazzaferro: Docstring added in f47cb3f79: `Ok(true)` = new row inserted; `Ok(false)` = sequence number already present (duplicate delivery — safe to delete from SQS either way); `Err` = DB failure, message is NOT deleted and gets re-received.

iris-mpc-common/src/config/mod.rs
carlomazzaferro: yeah, removed in f47cb3f79. I think my claude ran into the issue of InvalidCommonConfig when deploying once and wanted to protect against it (which never happens in practices)

iris-mpc-store/src/lib.rs
```diff
@@ -486,6 +502,189 @@ WHERE id = $1;
...
+    // recovery can release a claim that an old pod's in-flight results tx then
+    // persist-marks — without the guard such a row would be re-formed and
+    // re-processed despite its results being committed.
+    pub async fn count_pending_ingested_requests(&self) -> Result<i64> {
+        let count: (i64,) = sqlx::query_as(
+            r#"
+            SELECT COUNT(*)
+            FROM ingested_requests
+            WHERE consumed_batch_id IS NULL
+              AND persisted_at IS NULL
```
gryaele, line 547: When is persisted_at field updated? It's not on insert into db, right?

iris-mpc-bins/bin/iris-mpc/server.rs
```diff
@@ -265,6 +265,8 @@ async fn server_main(config: Config) -> Result<()> {
         next_sns_sequence_num: next_sns_seq_number_future.await?,
         common_config: CommonConfig::from(config.clone()),
         graph_mutation_bytes: vec![],
+        // GPU path has no DB-backed ingest wiring.
```
dkales: Shouldn't it also get it though?

iris-mpc-store/src/lib.rs
```diff
@@ -486,6 +502,189 @@ WHERE id = $1;
...
+        if sequence_numbers.is_empty() {
+            return Ok(());
+        }
+
+        let result = sqlx::query(
+            r#"
+            UPDATE ingested_requests
+            SET consumed_batch_id = $1
+            WHERE sequence_number = ANY($2::text[])
+              AND consumed_batch_id IS NULL
```
gryaele, line 586: Is AND consumed_batch_id IS NULL necessary here? Is it possible for the request to be consumed twice?

iris-mpc-store/src/lib.rs
carlomazzaferro: yeah never on insert (NULL = pending). 

Otherwiswe:
- normally inside the batch results tx (atomic with irises/graph/modifications: 'persisted' and 'results committed' are the same event)
- at boot via the skip-ahead for (modifications sync).

iris-mpc-bins/bin/iris-mpc/server.rs
carlomazzaferro: yeah, for now since I was testing what could be tested locally (i.e. not GPU) I left the wiring for GPU out of the question to decrease review surface

iris-mpc/src/server/mod.rs
```diff
@@ -154,7 +167,63 @@ pub async fn server_main(config: Config) -> Result<()> {
...
+        // them again. Releasing them instead (the naive local decision) would
+        // make this party re-form rows the peers moved past — a permanent
+        // batch divergence. Locally, "claimed but unpersisted from a crashed
+        // symmetric batch" and "claimed but unpersisted because only MY commit
+        // failed" are indistinguishable; only the peers' frontier tells them
+        // apart.
+        let fleet_frontier = sync_result.max_persisted_sequence_number();
+        if let Some(frontier) = &fleet_frontier {
+            let marked = iris_store
+                .mark_ingested_requests_persisted_up_to(frontier)
```
gryaele, line 204: Persisted timestamp is used here to skip the left-over messages. I thought persisted_at indicated that the timestamp when the request was processed. Would it make sense to use a separate field/flag for these ignored messages?

gryaele: Would marked be > 0 if iris share was persisted in 1 party but not in another?

iris-mpc-store/src/lib.rs
carlomazzaferro: it is just a loud failure, not something to be expected. It makes is so that the claim for a specific row is a compare-and-swap op. 

A second claimer (e.g. the wild case in which a pod coming up tries to claim this row) matches 0 rows on the overlap and trips the rows_affected check and errors instead of silently failing

iris-mpc/src/server/mod.rs
carlomazzaferro: the separate field question I think is more of an observability one: `db_ingest_frontier_skip_ahead` and `db_ingest_poison_rows` metrics already exist. maybe something more honest would be renaming the field to `finalized_at` or `completed_at`.  in the end, `persisted_at IS NOT NULL` means "terminal, never form this row again", and every pending-predicate checks that condition

carlomazzaferro: about marked > 0, yes, correct. to note: once that happens, the iris/graph data itself isn't skipped, it arrives via the modifications roll-forward on startup. this just reconciles the booking to what happened to on the other nodes

gryaele (dismissed):

gryaele (approved):

PR #2266, opened 2026-07-08
genesis: version-join delta with uniform graph repair (replaces modifications replay)

mcalancea:
## What

Replaces the genesis delta phase: the modifications-replay mode is removed, and the delta now reconciles the HNSW graph and HNSW iris store to the source iris store by `(serial → version)` comparison against a base S3 checkpoint ("version-join"), with uniform graph repair.

Under the schema invariant that any content change bumps `version_id` (`increment_version_id_trigger`), version comparison subsumes the modifications log: every replay-worthy modification surfaces as a version mismatch, while unlogged writes, multi-version lineages, and cursor corruption — invisible to log replay — surface the same way. The modifications log remains a telemetry input (the delta logs a join-vs-log comparison with set checksums).

## How

**Classification** (`version_join.rs`, pure): for each serial up to the checkpoint height, any disagreement puts it in the repair set — graph key missing / behind / ahead / multiple keys, HNSW row missing, HNSW row version ≠ source. Per-class reasons are retained for logging/metrics only. Plans are computed per eye and unioned (eyes can hold asymmetric graph damage).

**Deletion tombstones** are detected by content comparison against the party's deterministic dummy shares (both eyes); tombstones are removed without reinsert, and never-indexed tombstones are dropped.

**Cross-party consensus**: before any MPC work, parties exchange repair sets over the genesis coordination server and take the union — each party's HNSW rows are party-local, and the union makes the interactive MPC job list identical everywhere (a replay on a locally-clean party is a refresh). The tombstone sets must agree exactly (set-hash exchange, bail on mismatch — source stores disagreeing on content is not repairable here).

**Repair** (per serial, ascending, sequential): remove every existing graph key for the serial (per-eye key lists, version-sorted), then MPC search + reinsert at the current source version. Removing an entry point is allowed; the searcher falls back to a temporary entry point.

**Durability ordering**: no HNSW row write happens until the post-delta graph checkpoint upload succeeds; then missing rows are INSERTed and repaired rows UPDATEd (explicit-version writes, row-count-verified). A crash before the checkpoint leaves every row untouched, so a rerun re-derives the identical plan — the divergence signals that triggered repair are never erased before the graph repair is durable.

**Setup changes**: `--delta-mode` is gone; `--base-checkpoint-hash` optionally pins the base (resolvable from anywhere in local checkpoint history; all parties must pin the same hash, enforced by sync-config equality). A pinned reset trims the HNSW schema to the checkpoint (iris tail, cursors, WAL, modifications table) and drops newer checkpoint rows by row id. Without a common checkpoint, the run requires a fresh store (`last_indexed_id == 0`) and skips the delta. The cross-party sync config now carries only globally-agreed fields. The end-of-delta modification cursor and checkpoint metadata are stamped with the source's max persisted+completed modification id, read before the iris pools load.

**Layout**: genesis processing is split from a single `mod.rs` into per-phase modules (`setup`, `snapshot`, `indexation`, `delta`); `mod.rs` keeps the `exec` orchestrator and shared types.

## Testing

- `version_join` unit tests: classification, tombstone split, per-eye union, row-write derivation.
- e2e `genesis_110`: builds a base graph with a baked multi-version lineage, injects one serial per divergence class (version bump with/without log row, log row without bump, rejected modification, row version drift, row deletion, tombstone, delete-then-reinsert, missing rows, stale tail, junk WAL, garbage cursors — including damage applied to a single party only), then asserts full reconciliation: byte-wise store equality, exact graph keys per serial, empty WAL/modifications tables, correct cursors, post-delta checkpoint present, and a second run is quiescent (checkpoint hash unchanged).
- Existing genesis workflows (100–109) updated where they exercised the removed mode; the plaintext reference model mirrors the repair semantics so model-vs-MPC graph-equality assertions remain.


iris-mpc-cpu/src/genesis/version_join.rs
bgillesp, line 1: A couple of thoughts on the version join logic.

1. In principle version updates from the source DB could be from deletions.  In the GPU release, I believe deletion is tracked by updating the iris to an incremented version with iris shares replaced by dummy values.  One might double-check the source iris shares prior to doing a replay to see if the shares are equal to the dummy shares for the party in question, and if so, only delete from the graph, don't re-insert.

2. My feeling is that it would be good to at least allow (perhaps via an additional flag) a more aggressive version of repairing the graph which also fixes some additional of the anomalous cases.  Specifically: a) when a source serial is not present in the graph, we could insert it; and b) if the graph has a max version higher than the source version, we could insert the correct source version anyway.  For b), replay of the lower source version should be justified for both V5 and V4 graphs for different reasons.  For V5, since the format enforces just a single node per serial, overwriting with the older version brings the graph data in line with the source version.  For V4, it's possible that the older version is still present in the graph as an unpruned node.  If this is the case, then it's probably still worth re-inserting the source version into the graph even if it's already present, since subsequent updates to the graph that led to the current state may not have been able to link to the older node version after the version was updated in the registry.

iris-mpc-upgrade-hawk/src/genesis/graph_checkpoint.rs
```diff
@@ -87,3 +89,57 @@ pub async fn maybe_rollback_iris_db(
     }
     Ok(())
 }
+
+/// Reset all HNSW-schema state to `graph_checkpoint` for version-join mode:
+/// trim the iris tail beyond the checkpoint, restore the indexed cursors, and
+/// clear the WAL and modifications table. When a base checkpoint was pinned,
+/// also drop checkpoint rows that post-date it so an abandoned lineage cannot
+/// win the next run's latest-common selection. Single transaction: a crash
```
bgillesp: Perhaps this pruning of later checkpoint rows should be unconditional?  The functionality does a full DB rewind, so later checkpoints, even if they were skipped by the consensus mechanism, likely would not be valid with the updated state.

bgillesp (commented):
Looking quite good!  I haven't gone through everything, but I left a few comments on some parts of the high-level logic.

sdwoodbury (dismissed):
approved. 

If agreed, let's merge [2289](https://github.com/worldcoin/iris-mpc/pull/2289) into here too, or as a follow up.

sdwoodbury (approved):

PR #2268, opened 2026-07-08
[POP-4099] seed extended searches

sdwoodbury:
No need to do a full second search for supermatchers. Starting with the `edges` passed to `classify_and_extend()`, create a `SortedNeighborhood` and use it to search layer 0 via `search_layer()`.


mcalancea (commented):
I'm wondering whether it's safer to include the layer 1 search results together with the seed, so as to allow the search to branch out at a higher level too (i.e not risk the seeds to localize it too much). 

Did we manage to test the effects of this change on dev?

PR #2274, opened 2026-07-10
fix(store): gate version_id writes behind an explicit-version flag

mcalancea:
The `irises` `BEFORE UPDATE` trigger auto-incremented `version_id` on any content change, overriding a `version_id` set by the `UPDATE` itself — so `update_iris_with_version_id` could not write an authoritative version.

The trigger now:
- honors a hand-set `version_id` verbatim (any value, including a lower one) when the transaction enables `app.explicit_version_id`;
- rejects a hand-set `version_id` otherwise, instead of silently overwriting it;
- auto-increments on content change when `version_id` is untouched — unchanged behavior, so the `update_iris` path is unaffected.

`ExplicitVersion` enables the flag for a transaction and is required by the version-writing APIs (`update_iris_with_version_id`, `increment_iris_version`), so the flag is a compile-time requirement rather than a convention; the trigger is the runtime backstop for raw SQL. The migration replaces the trigger function in place (bound by name).

Tests cover: auto-increment when untouched, verbatim explicit version (differing, equal, and lower), and rejection of a hand-set `version_id` without the flag.

### DIFF
--- a/iris-mpc-upgrade-hawk/src/genesis/mod.rs
+++ b/iris-mpc-upgrade-hawk/src/genesis/mod.rs
@@ -1292,8 +1292,7 @@ async fn get_results_thread(
                                     right_code: &right_iris.code.coefs,
                                     right_mask: &right_iris.mask.coefs,
                                 };
-                    // We should ensure that the vector_id_to_persist is matching the inserted serial id.
-                    // The handle bypasses the auto-increment trigger so the replayed version is written verbatim.
+                    // Replay the modification's version verbatim; the handle bypasses the auto-increment trigger.
                     {
                         let mut ev = ExplicitVersion::enable(&mut graph_tx.tx).await?;
                         hnsw_iris_store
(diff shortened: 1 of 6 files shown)


sdwoodbury (dismissed):

iris-mpc-upgrade-hawk/src/genesis/mod.rs
```diff
@@ -1285,20 +1285,24 @@ async fn get_results_thread(
...
+                            .update_iris_with_version_id(
+                                &mut ev,
+                                vector_id_to_persist.version_id(),
+                                &iris_data,
+                            )
+                            .await?;
+                    }
 
                     let mut db_tx = graph_tx.tx;
                     set_last_indexed_modification_id(&mut db_tx, modification_id).await?;
```
carlomazzaferro, line 1307: This updates a different table, but the `ExplicitVersion::enable` flag is still set after the block closes, right? since we're in the same transaction.

It is not a problem as is but somewhat confusing since we sent the ev in the block, but that ev config stays one after the block.

maybe some comment about this could avoid future confusion

mcalancea: Good point. I renamed it and moved it closer to the start of the transaction with an appropriate comment, so that it becomes clear it's just a token to be consumed and not a scope guard or anything.

carlomazzaferro (approved):

PR #2299, opened 2026-07-23
detect when an extended search affected the match result

sdwoodbury:
Add metrics to detect whether a supermatcher extended search changed a request's match outcome. For requests whose search was extended, we retain the initial search (named `pre_extension` in the code) result alongside the extended one and compare the two decisions.

  New counters:
  - `supermatcher_extended_search_found_new_match` — extended search matched, original alone would not have.
  - `supermatcher_extended_search_lost_match` — original would have matched, extended did not.
 
note that 3 `info!` logs were consolidated. 

also note that `missing_vector_ids` now includes the `pre-extension` one-sided matches, which results in extra MPC comparisons (only when an extended search occurs).


iris-mpc-cpu/src/execution/hawk_main/matching.rs
```diff
@@ -284,22 +403,11 @@ impl BatchStep3 {
...
-                                Some(decision) => decision.is_mutation(),
-                                // The request we matched with is after us in the batch, so we are not blocked by it.
-                                None => false,
-                            }
-                        }
-                    });
+                    let (is_match, bsm) = uniqueness_is_match(request.select(filter), &decisions);
+                    because_supermatch = bsm;
+
+                    record_extension_metrics(request, filter, &decisions);
```
mcalancea, line 445: `decisions()` runs twice per batch on the same `BatchStep3`: once in `handle_mutations` (`hawk_main.rs:1939`) and once in `HawkResult::job_result` (`hawk_main.rs:1460`), both unconditional. So each counter increments twice per affected request, per party. `supermatcher_rejections` below has the same issue, but for a new metric it'd be nice to emit from a single place — e.g. have `decisions()` return the effect and let one caller emit it.

Related: a rate against `supermatcher_extended_searches` won't be meaningful, since that one is incremented per rotation task (3 rotations x 2 eyes x 2 orientations per request) while these are per request. If you want a ratio, a per-request denominator emitted from the same place would give matching units — one increment when `has_pre_extension()` is true.

iris-mpc-cpu/src/execution/hawk_main/matching.rs
```diff
@@ -255,6 +314,66 @@ pub const DECISION_FILTER: Filter = Filter {
...
+    match (pre_match, extended_match) {
+        (false, true) => {
+            metrics::counter!("supermatcher_extended_search_found_new_match").increment(1);
+        }
+        (true, false) => {
+            metrics::counter!("supermatcher_extended_search_lost_match").increment(1);
+        }
+        _ => {}
+    }
+}
```
mcalancea, line 411: Excluding `Supermatch` on both sides makes this measure "the extended search surfaced a new both-eye neighbour match", not "the decision changed". The decision-level flip stays invisible — in particular the case where the pre-extension results were saturated at the match threshold (`matches.len() + margin >= ef`) but the extended results are not, at the larger `ef_supermatch`: supermatch-rejected before, `UniqueInsert` after.

Both are useful, and they are different questions. The data for the second is already plumbed: `pre_extension` carries the original `matches.saturated`, which reaches the pre `ResolvedJoin`, and `select_pre` already yields `MatchId::Supermatch` from it. Running the same comparison without the `not_supermatch` filter would give the decision-level A/B for free.

iris-mpc-cpu/src/execution/hawk_main/matching.rs
```diff
@@ -145,14 +207,26 @@ impl Step1 {
...
         let other_side = 1 - side;
-        let anti_join = &self.anti_join[other_side];
+        let matched_one_side = &self.join.matched_one_side[other_side];
+        // Include the pre-extension one-sided matches so the pre-extension outcome
+        // can be resolved from the same MPC results (its ids are a subset of
+        // `join`'s, but include them explicitly to be safe).
+        let pre_matched_one_side = self
+            .pre_join
+            .iter()
+            .flat_map(|j| j.matched_one_side[other_side].iter());
```
mcalancea: The subset claim doesn't hold, and this chain is load-bearing rather than defensive. A vector that is one-sided pre-extension can be both-sided in the extended join, so it lands in `join.matched_both` and the extended join never requests a comparison for it. Leave it out of this list and `SearchJoin::resolve` finds no `missing_is_match` entry for the pre join, silently drops the id, and `found_new_match` fires even though the id would have matched before extension too.

The actual invariant is that the pre join's one-sided ids are contained in the extended join's one-sided ids plus its both-sided ids, and only the one-sided part gets resolved. Worth stating that here, so the chain doesn't get optimized away later.

bgillesp: Agreed on this.  One step further, the pre-join and extended results don't even technically have to have a subset relationship per side, since they are implemented via independent searches, so the approximate nature of the searches could result in non-comparable result sets.  I'll push an updated comment in a bit.

iris-mpc-cpu/src/execution/hawk_main/matching.rs
bgillesp: I'm looking into decoupling metrics emissions of `BatchStep3` from the underlying `decisions` call in a follow-up PR.  For now, adding a counter `supermatcher_extended_search_requests` in `record_extension_metrics` after the `has_pre_extension()` short-circuit to provide the denominator you mentioned.

iris-mpc-cpu/src/execution/hawk_main/matching.rs
bgillesp: Good observation I think.  I'm adding additional comparison metrics here targeting the decision process as well.  I think that for comparing the decision outcomes, it is sensible to exclude `Supermatch` outcomes from the pre-extension results.  The point which determines the outcome is "being saturated at the supermatcher level", and this isn't something that can reasonably be ascertained at the base search `ef` -- so the "default" decision without extension is "was a match found" without "were the matches saturated".  (In a case where the results were extended, most of the time an extended result set is saturated, producing a `Supermatch` flag for the set.  But not always: extension occurs when the anon stats results are saturated, which is measured a more lenient threshold.  However, base results saturation and extension are frequently correlated.)

So the final result metrics will compare "did the decision change" (pre-result against post-result-with-supermatch) and "did the extended search find a new match" (pre-result against post-result-without-supermatch).

bgillesp (approved):
I think this is fine to merge for now -- I'm going to take a pass at doing a bit more refactoring and documentation in the `matching` module to untangle some of the complex logic involved, in a follow-up PR.

PR #2305, opened 2026-07-28
chore: bump ampc-common to 9a3d6a4 — startup-wedge fix (#130) for GPU/HNSW/anon-stats binaries

carlomazzaferro:
Delivers [ampc-common #130](https://github.com/worldcoin/ampc-common/pull/130) to every coordination consumer in this repo (GPU server, HNSW server, anon-stats-server, genesis): **in-process retry of the startup unready barrier + graceful shutdown on peer UUID change** instead of the swallowed heartbeat panic. This is the fix for the skewed-restart wedge class (HNSW stage 2026-07-09, prod DI anon-stats 2026-07-27 — locally reproduced and verified self-healing <1 min with both halves).

Crossed ampc-common commits (bcc6addb → 9a3d6a4):
- `9a3d6a4` #130 — the startup fix (this PR's payload)
- `c47f0b5` #127 — face anon-stats op support (additive)
- `d42a62d` #128 — created_at + concurrent-index migration on `anon_stats_2d_di` — **instant here** (this side's table is empty); applies at anon-stats-server boot
(#122 handshake hardening was already on board.)

**Stage rollout:** after CI builds, stage image pins (values) need their bump to the new build — follow-up PR once images exist. Mixed-version note: full self-heal benefit arrives once all 3 parties of a fleet run the new build; restart fleets together if churn appears during transition.

cargo check workspace clean. Related: POP-4162 (visibility-barrier adoption for the non-HNSW binaries).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,7 +78,7 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 [[package]]
 name = "ampc-actor-utils"
 version = "0.1.0"
-source = "git+https://github.com/worldcoin/ampc-common.git?rev=bcc6addbaab6297db7699a8c739d9056b0bdcbe#bcc6addbaab6297db7699a8c739d9056b0bdcbed"
+source = "git+https://github.com/worldcoin/ampc-common.git?rev=9a3d6a4717910b684c055e3ca83e63867dfa9dba#9a3d6a4717910b684c055e3ca83e63867dfa9dba"
 dependencies = [
  "aes-prng",
  "ampc-secret-sharing",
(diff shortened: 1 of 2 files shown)


gryaele (approved):

carlomazzaferro:
CPU Hawk e2e red on this branch is NOT the bump — bisected to the anon-stats migrations deadlocking under the e2e's shared-DB topology (advisory lock + CREATE INDEX CONCURRENTLY); #130 is exonerated (c47f0b5 without it already fails). Root-cause + fix in **#2310** (per-party test databases). Merge order: land #2310 → rebase/rerun this → green. Verified: #2310's fix + this bump ran e2e green 5/5 (ex-#2309, run 30349896607).

PR #2306, opened 2026-07-28
[DO NOT MERGE] ci control: CPU Hawk E2E on #2305's base

carlomazzaferro:
Empty-commit control for the deterministic wal-test migration deadlock seen on #2305 (both attempts): runs the identical e2e on base 06673c8bd WITHOUT the ampc-common bump. Green here → the bump is implicated and #2305 stays blocked pending root-cause; red here → main is already broken and #2305 is exonerated. Will be closed after the run either way.


carlomazzaferro:
Bisect complete — closing throwaway control. Result folded into #2305 analysis: the wal migration deadlock is triggered by ANY new anon-stats migration applied concurrently by the 3 e2e parties on their shared Postgres (base green; d42a62d/#128 red; c47f0b5/#128+#127 red; 9a3d6a4/+#130 red → #130 not the cause).

PR #2308, opened 2026-07-28
[DO NOT MERGE] ci bisect: ampc-common d42a62d (#128 only)

carlomazzaferro:
Third bisect point. base green, c47f0b5(#128+#127) red, 9a3d6a4(+#130) red. This pins d42a62d = #128 (anon_stats_2d_di created_at migration) ALONE. #128 only adds SQL files under ampc-common anon_stats_migrations/ — inert to iris-mpc (hawk runs sqlx::migrate!("./../migrations")), so if this is RED the cause is timing/rebuild-exposed not #128 content; if GREEN the culprit is #127. DO NOT MERGE.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,10 +86,10 @@ tokio-util = "0.7.15"
 toml = { version = "0.8.23", features = ["preserve_order"] }
 uuid = { version = "1", features = ["v4"] }
 iris-mpc-cpu = { path = "./iris-mpc-cpu" }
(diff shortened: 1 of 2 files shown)


carlomazzaferro:
Bisect complete — closing throwaway control. Result folded into #2305 analysis: the wal migration deadlock is triggered by ANY new anon-stats migration applied concurrently by the 3 e2e parties on their shared Postgres (base green; d42a62d/#128 red; c47f0b5/#128+#127 red; 9a3d6a4/+#130 red → #130 not the cause).

PR #2314, opened 2026-07-29
update dev cluster deployment workflow to convert old graph types to v5

sdwoodbury:
This is needed because after the v5 graph update, `iris-mpc-hawk` will reject old checkpoints and require genesis to update them. Old graphs can be converted using the `graph-utils` binary, allowing testing to continue with old graphs.


bgillesp (approved):
Looks fine to me!

PR #2317, opened 2026-07-30
genesis: remove the unused post-run snapshot phase

mcalancea:
Genesis had an optional phase 3 that created an RDS cluster snapshot after indexation. The `--perform-snapshot` flag has only ever been passed as `false`, so the phase is unreachable in practice; this removes the flag, the phase, and the RDS client that existed solely to serve it, which drops the `aws-sdk-rds` dependency from the genesis path.

Compatibility: any deployment config that still passes `--perform-snapshot` must drop the flag before adopting an image with this change, since the argument parser rejects unknown flags.


carlomazzaferro (approved):

carlomazzaferro (approved):

sdwoodbury:
one of the commits wasn't signed so I checked out a new branch, squash merged (cherry-pick failed), diffed with the original branch (they were the same), and then force pushed.

PR #2318, opened 2026-07-30
feat(startup): run the mutual-visibility barrier on every ampc consumer (POP-4162)

carlomazzaferro:
> [ampc-common #132](https://github.com/worldcoin/ampc-common/pull/132) is **merged**; the pin here points at its squash SHA `3faee93` and the branch is rebased on current main. Ready for review.

## Problem

[#2152](https://github.com/worldcoin/iris-mpc/pull/2152) (POP-3883) added the startup mutual-visibility barrier — hold until every peer's `verified_peers` contains every current UUID — because node 0 was progressing into DB load while peers still waited on its health endpoint. It only ever ran in the HNSW server. The GPU server, the anon-stats server and the genesis coordinator call `wait_for_others_unready` alone, so all three are still exposed to that same initial-deploy wedge.

## Change

All four sites call ampc-common's new `wait_for_startup_barriers` (unready → visibility):

| Site | Before | After |
| --- | --- | --- |
| `iris-mpc/src/server/mod.rs` (HNSW) | both calls, explicit | one call, identical semantics |
| `iris-mpc-bins/bin/iris-mpc/server.rs` (GPU) | unready only | both |
| `iris-mpc-bins/bin/iris-mpc-anon-stats-server/main.rs` | unready only | both |
| `iris-mpc-upgrade-hawk/src/genesis/setup.rs` | unready only | both |

HNSW is a pure collapse — same two functions, same arguments, same error types, no double-wait. Genesis ordering was checked specifically: nothing interleaves between the unready barrier and heartbeat init, `/health` is served by the coordination server started before the barrier, and the node is still unready during phase 2 (`set_node_ready` comes much later), so peers' phase 1 passes trivially. No circularity.

## Crossed upstream changes

The pin bump (9a3d6a4 → #132) crosses two commits with call-site impact:

**#129** (heartbeat + `wait_for_others_ready` hardening) — both now validate peer UUIDs against the startup-verified set and fail fast on an unknown one. Each site snapshots `verified_peers` after the barriers and passes it down, matching the pattern in [deep-identifier-ampc#97](https://github.com/worldcoin/deep-identifier-ampc/pull/97). Snapshotting after *both* phases rather than after unready alone makes the set strictly more complete.

**#131** — `BucketStatistics2D::new` gained the two mirror-match arguments; `iris-mpc-common/tests/statistics.rs` passes `None, None`.

## Test harness fix worth a look

`iris-mpc/tests/utils/wait_conditions.rs` no longer calls `wait_for_others_ready`. Post-#129 that function `ensure!`s every peer UUID is in the caller's verified set — correct for a cluster member, wrong for this harness, which never runs the startup handshake and therefore holds an empty set. Every peer would look like a restarted node and the CPU Hawk e2e would fail on the first poll. The harness now polls `/health` for `is_ready` itself, without the restart-detection invariant (it isn't a fleet member, so that invariant isn't its to enforce). `configs.rs` and `wait_conditions.rs` also pick up the new `startup_visibility_barrier_disabled` field, since `ServerCoordinationConfig` has no `Default` and both literals are exhaustive.

## Testing

`cargo check --workspace --all-targets` clean; `cargo fmt --all` applied. CI runs the CPU Hawk e2e, which is the real exercise of both the barrier and the harness change. The 3-party local-stack acceptance (skewed cold start + targeted single-party kill, the #130 repro) runs on the deep-identifier stack before any of these leave draft.

## Startup budget

Worst-case barrier wall-clock is ≈4× `startup_sync_timeout_secs` (default 300s → 1200s) — see #132's description for the arithmetic. Every iris-mpc `startupProbe` budget should be checked against that before this rolls to prod, though no values change is included here.

Ticket: [POP-4162](https://linear.app/worldcoin/issue/POP-4162)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,7 +78,7 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 [[package]]
 name = "ampc-actor-utils"
 version = "0.1.0"
-source = "git+https://github.com/worldcoin/ampc-common.git?rev=9a3d6a4717910b684c055e3ca83e63867dfa9dba#9a3d6a4717910b684c055e3ca83e63867dfa9dba"
+source = "git+https://github.com/worldcoin/ampc-common.git?rev=ebc6c704f691951e24e554603d4399f3b27b6beb#ebc6c704f691951e24e554603d4399f3b27b6beb"
 dependencies = [
  "aes-prng",
  "ampc-secret-sharing",
@@ -111,7 +111,7 @@ dependencies = [
 [[package]]
 name = "ampc-anon-stats"
 version = "0.1.0"
-source = "git+https://github.com/worldcoin/ampc-common.git?rev=9a3d6a4717910b684c055e3ca83e63867dfa9dba#9a3d6a4717910b684c055e3ca83e63867dfa9dba"
+source = "git+https://github.com/worldcoin/ampc-common.git?rev=ebc6c704f691951e24e554603d4399f3b27b6beb#ebc6c704f691951e24e554603d4399f3b27b6beb"
 dependencies = [
  "ampc-actor-utils",
  "ampc-secret-sharing",
@@ -137,7 +137,7 @@ dependencies = [
 [[package]]
 name = "ampc-secret-sharing"
 version = "0.1.0"
-source = "git+https://github.com/worldcoin/ampc-common.git?rev=9a3d6a4717910b684c055e3ca83e63867dfa9dba#9a3d6a4717910b684c055e3ca83e63867dfa9dba"
+source = "git+https://github.com/worldcoin/ampc-common.git?rev=ebc6c704f691951e24e554603d4399f3b27b6beb#ebc6c704f691951e24e554603d4399f3b27b6beb"
 dependencies = [
  "aes-prng",
  "base64",
@@ -153,7 +153,7 @@ dependencies = [
 [[package]]
 name = "ampc-server-utils"
 version = "0.1.0"
-source = "git+https://github.com/worldcoin/ampc-common.git?rev=9a3d6a4717910b684c055e3ca83e63867dfa9dba#9a3d6a4717910b684c055e3ca83e63867dfa9dba"
+source = "git+https://github.com/worldcoin/ampc-common.git?rev=ebc6c704f691951e24e554603d4399f3b27b6beb#ebc6c704f691951e24e554603d4399f3b27b6beb"
 dependencies = [
  "aws-sdk-secretsmanager",
  "aws-sdk-sqs",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,7 +78,7 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 [[package]]
 name = "ampc-actor-utils"
 version = "0.1.0"
-source = "git+https://github.com/worldcoin/ampc-common.git?rev=ebc6c704f691951e24e554603d4399f3b27b6beb#ebc6c704f691951e24e554603d4399f3b27b6beb"
+source = "git+https://github.com/worldcoin/ampc-common.git?rev=3faee93c15676d0284497573bbff38629a21114c#3faee93c15676d0284497573bbff38629a21114c"
 dependencies = [
  "aes-prng",
  "ampc-secret-sharing",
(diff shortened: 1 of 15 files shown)


carlomazzaferro:
## Genesis e2e was red — root cause was a 15s poll interval in the fixtures, now green in 12m08s

Recording this because the diagnosis went one way and the answer went another.

**Symptom:** job cancelled at exactly 20m40s (`timeout-minutes: 20`), bob and charlie looping on `Handshake failed: not accepted: rsp=[]` against alice. Every one of the 72 party-boots logged both barrier phases completing, and identical code had passed 83 minutes earlier — all of which looked like evidence the barrier was innocent. It wasn't. Re-running the identical commit reproduced the timeout to the second, which ruled out a flake.

**The barrier had eaten the suite's headroom.** Measured durations:

| run | has this PR? | duration |
| --- | --- | --- |
| `223a722b` (mihai/parallel-legacy-prune) | no | 17m45s |
| `7b025207` (mihai/genesis-remove-snapshot) | no | 18m37s |
| `d767a1e5` (this PR, green) | yes | 19m45s — 15s of margin |
| `219e5afb` (this PR, twice) | yes | 20m40s = the ceiling |

A timeout is a resource failure, not a logic one, and the cheapest diagnostic — compare durations against branches without the change — was the one worth running first.

**The actual lever.** Histogramming stalls in the failing run: 40 gaps ≥3s totalling 634s, of which **24 were exactly 14.998s — 6 minutes of an 18-minute suite.** A repeated machine-exact interval is a constant, and this one greps: all six genesis fixtures (`tests/resources/node-config/{docker,local}/node-{0,1,2}-genesis-0.toml`) set `http_query_retry_delay_ms = 15000`. A 15-second poll interval between three nodes on loopback. That is also why one extra barrier phase per boot was enough to blow the ceiling *here* specifically — a retry cycle costs 15s in this config versus ~1s in the others.

**Fix:** those six fixtures now use 250ms. The barriers stay genuinely exercised — they still have to converge, and a peer that never appears still fails on the unchanged 300s `startup_sync_timeout_secs`. **cpu-e2e now passes in 12m08s**, so this PR adds the visibility barrier to four binaries *and* leaves the genesis suite ~6 minutes faster than it was before, with headroom up from 15 seconds to about 8 minutes.

I did not raise `timeout-minutes`, and I'd push back on doing so: the ceiling was never the problem, and raising it would have preserved the six minutes of waste while hiding the 15s interval.

**Also in this PR, from the same investigation:** `iris-mpc/tests/utils/configs.rs` and `wait_conditions.rs` drop from a 1s to a 100ms poll cadence. That's the CPU Hawk harness rather than genesis, so it wasn't the fix for this failure, but the same reasoning applies and it's cheap.

**Left alone deliberately** (~2.5 min, build/infra rather than config, and not this PR's business): docker compose bring-up 38.5s, in-job compilation of `iris-mpc-upgrade-hawk` 37s and `iris-mpc-cpu` 17.5s, docker layer writes 36s + 19.5s. Worth a look if anyone wants the suite under 10 minutes.

carlomazzaferro:
## Green — 21/21. Follow-up: I broke `cpu-integration-tests` and reverted it

Correcting the previous comment. The poll-cadence change it mentions in passing (`configs.rs` / `wait_conditions.rs`, 1s → 100ms) was not harmless — it roughly doubled `cpu-integration-tests`, a separate workflow with a 30-minute cap:

| commit | contains | duration | result |
| --- | --- | --- | --- |
| `d767a1e5` | barrier only | 16m45s | pass |
| `219e5afb` | + rebase/re-pin | 17m04s | pass |
| `2b0b558e` | + cadence 1s→100ms | **30m46s** | hit the ceiling |
| `57a51b75` | cadence reverted | **20m07s** | pass |

Reverted in `57a51b75`. The genesis fixture change stays — that one is measured and load-bearing.

**Final position on this PR, all measured:**

| suite | baseline (no barrier) | ceiling | this PR | headroom |
| --- | --- | --- | --- | --- |
| genesis cpu-e2e | 17m45s / 18m37s | 20 min | **11m52s** | ~8 min |
| cpu-integration-tests | 16m45s / 17m04s | 30 min | **20m07s** | ~10 min |

Genesis comes out ~6 minutes faster than before the barrier existed. `cpu-integration-tests` is ~3 minutes slower, which is the barrier's real cost in a suite that keeps the 1s cadence and pays it per boot. No `timeout-minutes` change anywhere.

**The result worth keeping if you ever tune these harnesses:** in `cpu-integration-tests`, 1000ms → 20m07s but 100ms → >30m. Faster polling made it *dramatically slower*. Those three parties run in one process, so the poll loop competes with the code under test for the same runtime instead of just observing it — poll cadence there is a resource decision, not a latency one, and an intermediate value like 500ms isn't worth trying because the direction of the effect is wrong. Genesis tolerated 15s → 250ms only because its parties are in separate containers. The two suites also read different configs (`iris-mpc/tests/utils/configs.rs` vs the genesis TOML fixtures), which is what I got wrong the first time.

wojciechsromek (approved):

PR #2321, opened 2026-07-30
Cut redundant work in the graph mutation-apply path

mcalancea:
Two changes to the mutation-apply hot path. No effect on the stored format, checkpoint layout, checksums, or mutation semantics.

**Compaction scan: raw-degree guard.** `compact_batch` decoded and staleness-filtered every candidate neighborhood to find the oversized ones. Raw degree — a header read on the encoded form — upper-bounds active degree, so candidates at or under `M_limit` are now skipped without decoding. Identical output.

**Staleness-filter watermark.** `edit_neighborhood` ran the per-entry content-clock filter on every touch. `GraphMem` now tracks the seq_no of the last op that can invalidate existing edges (`RemoveNode`, or an `AddNode` re-minting a live serial); a neighborhood stamped at or after that watermark provably holds no stale edge, so the filter is skipped. The field is derived and in-memory only: loads seed it conservatively at `last_update_seq_no`, replay maintains it deterministically, and the existing `debug_assert` in `edit_neighborhood` guards the skip. `GraphMem` equality now compares semantic state only (derived fields excluded), so a conservatively seeded load compares equal to its mint.

Measured on a synthetic 17M-node graph on Graviton (x8g), single-threaded: applying a batch of 96 inserts with 320 bilateral links went from 1.64 s to 288 ms, and the compaction candidate scan from ~1.4 s to 1.3 ms. The benchmark harness lands separately.

### DIFF
--- a/iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
+++ b/iris-mpc-cpu/src/hnsw/graph/layered_graph.rs
@@ -179,9 +179,11 @@ pub struct GraphMem {
 
     /// Seq_no of the last op that can invalidate existing edges: a
     /// `RemoveNode`, or an `AddNode` re-minting a live serial. A neighborhood
-    /// stamped at or after it holds no stale edge, so
-    /// [`Self::edit_neighborhood`] skips the staleness filter. Derived and
-    /// in-memory only; `from_parts` seeds it conservatively.
+    /// stamped at or after it holds no stale edge — a precondition on loaded
+    /// state, guaranteed by `from_parts` seeding past the load point so every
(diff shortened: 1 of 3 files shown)


sdwoodbury (approved):

bgillesp (approved):
Concept looks clean!  Not clear to me how effective the `last_invalidation_seq` logic is at reducing stale edge filtering overhead -- my uncalibrated guess would be that most neighborhoods are older than the cutoff, since reauths aren't _that_ rare, but it could be that there's correlation between neighborhoods processed during different operations from a single request.  If the benchmarks show an improvement overall (and pretty significant!), it seems fine to me.

PR #2326, opened 2026-08-03
chore(prod): bump HNSW to v0.32.8 for the startup barrier

carlomazzaferro:
v0.32.7 predates #2318, so prod HNSW has been running without the mutual-visibility startup barrier. Parties 0 and 2 have been crash-looping with a clean exit 1 roughly hourly since v0.32.7 went out on 07-31 14:05Z; v0.32.8 contains the barrier.

No ampc-common bump needed, main already pins 3faee93 (#132).

Extra:
- anon-stats-server v0.32.6 -> v0.32.8, the other ampc consumer #2318 touched in this fleet

### DIFF
--- a/deploy/prod/common-values-ampc-hnsw-anon-stats-server.yaml
+++ b/deploy/prod/common-values-ampc-hnsw-anon-stats-server.yaml
@@ -1,4 +1,4 @@
-image: "ghcr.io/worldcoin/anon-stats-server:v0.32.6@sha256:bb88af42ca4a1c4af2803adb65f9ca09da6028011dad55d607b3c8affd6b0f6a"
+image: "ghcr.io/worldcoin/anon-stats-server:v0.32.8@sha256:fb42e8448cfdd2cee01db069a17ed4181318f3290457024f5662cc30562a20f7"
 
 environment: prod
 replicaCount: 1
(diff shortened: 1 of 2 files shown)


carlomazzaferro:
Timeline: v0.32.7 = `ad3bd9e9e` (07-30 13:50), #2318 merged 07-31 11:16Z, #2324 bumped prod to v0.32.7 at 07-31 14:05:47Z. The three HNSW pods started 14:06:32 / 14:08:13 / 14:08:23Z and 0 and 2 have been looping ever since, 68 restarts each at roughly one per 59 min, exit 1. Party 1 has been up 37h on the same image, which is why this reads as peer verification failing rather than a crash on boot.

Digests here are index digests off the ghcr manifest endpoint, not per-platform ones (`docker manifest inspect -v` returns the wrong value). Verified the method by resolving v0.32.6 and v0.32.7 for both images first and matching them against what is already committed in these files.

Left out on purpose:
- graph-persistence-job, still on iris-mpc-cpu v0.32.6: v0.32.6 -> v0.32.8 crosses #2302 (tiered checkpoint rotation), which is a behavior change and wants its own PR
- anon-stats / modifications retention reapers: mid dry-run ladder on POP-4134
- db-exporter: v0.31.11, separate cadence

One thing to flag: stage is pinned to `6429fa5c`, which also predates #2318, so the barrier has no stage soak and prod is the first place it runs. Given 0 and 2 are already down I would still take it, but worth knowing before merge.

PR #2334, opened 2026-08-06
allow peer restarts before all are ready

sdwoodbury:
Before the peers reach `ready`, rather than ensure a peer hasn't restarted (via a `uuid` check), ensure that the `SyncState` has not changed.

Tested on the dev cluster where
- one peer comes up later than the others
- one peer restarts during the db load (`init_hawk_actor`)

Integration tests added:
- a party restarts before the ready signal with unchanged SyncState (sync passes)
- a party restarts before the ready signal with changed SyncState (sync fails)
- a party restarts after the ready signal with unchanged SyncState (sync fails)

### DIFF
--- a/iris-mpc/src/server/startup_phase.rs
+++ b/iris-mpc/src/server/startup_phase.rs
@@ -1,4 +1,4 @@
-//! Startup phase tracking and the data-derived *epoch* that keys it.
+//! Startup phase tracking and the data-derived [`SyncStateDigest`] that keys it.
 //!
 //! # Why this exists
 //!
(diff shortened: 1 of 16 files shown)


iris-mpc/src/server/startup_phase.rs
```diff
@@ -0,0 +1,1098 @@
...
+    Hash,
+    Serialize,
+    Deserialize,
+    derive_more::Debug,
+    derive_more::Display,
+)]
+#[serde_as]
+#[display("{}", hex::encode(_0))] // _0 accesses the inner [u8; 32]
+#[debug("{self}")] // Delegates Debug directly to Display
+pub struct SyncStateDigest(#[serde_as(as = "Hex")] [u8; 32]);
```
mcalancea, line 128: *Generated by Claude Code*

`#[serde_as]` placed after `#[derive(Serialize, Deserialize)]` is a no-op: the derive expands first, so `as = "Hex"` never applies and the digest serializes as a JSON array of 32 integers rather than the hex string this doc comment promises. (The tell: `Hex` isn't imported — with the attribute active, that would be a compile error.)

```rust
use serde_with::hex::Hex;

#[serde_as] // must precede the derive
#[derive(Clone, Copy, /* … */ Serialize, Deserialize)]
pub struct SyncStateDigest(#[serde_as(as = "Hex")] [u8; 32]);
```

Verified against serde_with 3.20.0: the current order prints `[171,171,…]`, the reordered version `"abab…"`. The round-trip test can't catch this (both representations round-trip fine), so consider also asserting the serialized JSON is the 64-char hex string.

iris-mpc/src/server/startup_phase.rs
```diff
@@ -0,0 +1,1098 @@
...
+    /// Short form for log lines, where the full 64 hex chars are noise.
+    pub fn short(&self) -> String {
+        hex::encode(&self.0[..6])
+    }
+}
+
+/// Canonical digest of one party's [`SyncState`]: everything about its local
+/// state that the startup sync reads, and that must therefore be identical
+/// across a restart for a rejoin to be safe.
+pub fn hash_sync_state(state: &SyncState) -> SyncStateDigest {
```
mcalancea, line 144: *Generated by Claude Code*

`hash_sync_state` and `digest_modifications` read fields off `&SyncState` / `&Modification` individually, so a field added to either struct later would silently stay out of the digest. Exhaustively destructuring the structs here (the pattern `From<Config> for CommonConfig` already uses) turns that into a compile error instead. Coverage is complete today — this is just future-proofing.

mcalancea (approved):

PR #2347, opened 2026-08-13
Add deployable CPU linear-scan backend

philsippl:
## Summary

Adds the correctness baseline for a deployable ARM CPU exact-scan backend as a new `iris-mpc-linear-scan` target. It reuses the production Hawk request, MPC, persistence, and result pipeline, but replaces HNSW candidate search with the GPU actor's two-eye linear-scan semantics. Existing GPU and HNSW targets remain available and unchanged.

This PR deliberately contains the simple scan implementation: no cold-eye cache/prefetch layer and no mixed-plane, fused mirror, NEON/UMMLA, or dot/MPC pipeline optimizations.

Depends on [worldcoin/ampc-common#138](https://github.com/worldcoin/ampc-common/pull/138); all four AMPC crates are pinned to `fa95ff78a91ef3cfb3c8020afab0e50eae0dc88a`.

## Behavior and correctness

- Loads one configured eye for the process lifetime and performs an exact 31-rotation scan in 4,096-record chunks.
- Checks the second eye only for first-eye matches and known LUC/OR-rule/reauth candidates, matching the GPU cascade.
- Enforces `SMPC__MAX_BATCH_SIZE=1`; the resident eye changes only on a coordinated restart through `SMPC__FULL_SCAN_SIDE`.
- Preserves enroll, duplicate, mirror, deletion, reauth, reset/recovery, retained distance-share, and result-persistence behavior.
- Uses exact database versions for cold-eye reads and fails the request closed on a missing row instead of substituting zero shares.
- Keeps database and in-memory version advancement consistent for idempotent iris updates and safely handles hidden partial results.

CPU/GPU equivalence was checked on 3 × A40 for public results and retained distance shares. A deterministic 100-request full-server sequence covering enroll, duplicate, mirror, deletion, reauth, reset, and recovery also produced identical canonical CPU/GPU result files with a 5,000-record database, larger than one CPU scan chunk.

## Validation

- `cargo check -p iris-mpc-bins --bin iris-mpc-linear-scan`
- `cargo clippy -p iris-mpc-cpu -p iris-mpc -p iris-mpc-bins --all-targets -- -D warnings`
- Linear-scan search tests: 6 passed.
- Result-shaping and server configuration tests: 4 passed.
- DB-backed cold-eye integration test compiles with `db_dependent`.

## Deployment note

The additive binary, image workflow, and stage/prod manifests are introduced here, but production should deploy the complete stack through #2348. Do not run GPU and CPU consumers against the shared queues simultaneously; drain all three GPU consumers, then roll all three CPU parties together with the same full-scan side.

## Review guide

1. Start with the additive entry point in `iris-mpc-bins/bin/iris-mpc/server/iris_mpc_linear_scan.rs` and mode selection in `iris-mpc/src/server/mod.rs`.
2. Review the two-eye cascade and chunk result assembly in `iris-mpc-cpu/src/execution/hawk_main/search.rs`.
3. Review one-eye initialization and exact cold-eye loading in `worker_pool_initializer.rs` and `iris_worker.rs`.
4. Finish with the GPU parity, CPU E2E, LocalStack, and native ground-truth tests.

## Stack

- **this PR** — correctness baseline
- #2351 — cold-eye cache and prefetch
- #2348 — compute and pipeline performance


philsippl:
@codex review

carlomazzaferro:
@codex review

iris-mpc-cpu/src/execution/hawk_main/test_utils.rs
```diff
@@ -26,6 +26,14 @@ use super::{
...
 
 pub async fn setup_hawk_actors() -> Result<Vec<HawkActor>> {
+    setup_hawk_actors_with_mode(false).await
+}
+
+pub async fn setup_linear_scan_actors() -> Result<Vec<HawkActor>> {
+    setup_hawk_actors_with_mode(true).await
+}
+
+async fn setup_hawk_actors_with_mode(linear_scan: bool) -> Result<Vec<HawkActor>> {
```
dkales: nit: can this not use HawkSearchMode?

iris-mpc/src/server/mod.rs
```diff
@@ -396,36 +427,80 @@ async fn init_shutdown_handler(config: &Config) -> Arc<ShutdownHandler> {
...
     // Load batch_size config
     tracing::info!("Set max batch size to {}", config.max_batch_size);
+    Ok(())
+}
+
+fn validate_max_batch_size(max_batch_size: usize, search_mode: HawkSearchMode) -> Result<()> {
+    if search_mode == HawkSearchMode::LinearScan && max_batch_size != 1 {
+        bail!(
+            "exact CPU linear-scan mode requires max_batch_size=1; got {max_batch_size}. \
+             Requests must be serialized so intra-batch matching is unnecessary"
```
dkales: why is this needed, could this not just be reused from the existing impl?
I guess there is no speedup anymore with batching?

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -170,6 +213,33 @@ impl IrisPoolHandle {
         result
     }
 
+    async fn dot_product_irises_batch(
+        &self,
+        query: ArcIris,
+        targets: Vec<ArcIris>,
+    ) -> Result<Vec<RingElement<u16>>> {
+        let targets: Arc<[ArcIris]> = Arc::from(targets);
```
dkales: This still does a full copy of them, no? I guess it is usually small, but would making that type Arc<Vec<ArcIris>> not be strictly better?

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -458,28 +637,85 @@ fn worker_thread(ch: Receiver<IrisTask>, iris_store: SharedIrisesRef<ArcIris>, n
             } => {
                 let store = iris_store.data.blocking_read();
 
-                let iris_pairs = vector_ids
+                let targets: Vec<Option<ArcIris>> = vector_ids
                     .iter()
-                    .map(|v| store.get_vector(v).map(|iris| (&query, iris)));
+                    .map(|v| store.get_vector(v).map(ResidentIris::to_arc))
+                    .collect();
```
dkales: Just to clarify, this maps the resident layout back into the traditional one for this op, if it is indeed stored in this way internally.
I guess this path is then not intended to be used in the linear scan layout?

I think this might benefit from some cleanup as the layout introduction here is just adding a lot of complexity in an already very complex part of the system.

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -905,23 +1275,53 @@ impl IrisWorkerPool for LocalIrisWorkerPool {
...
+        let pool = self.clone();
+        let is_cold = self.cold_storage.is_some();
+        let task_size = self.full_rotation_task_size;
         Box::pin(async move {
-            let store = iris_store.data.read().await;
-            Ok(ids
-                .iter()
-                .map(|id| store.get_vector_or_empty(id).clone())
-                .collect())
+            let iris = {
```
dkales, line 1344: This is written very confusingly I feel. Should this not just be a ringbuffer based on a VecDeque?
(This was about the RollingLucCache)

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -757,6 +1045,67 @@ impl LocalIrisWorkerPool {
...
+        self
+    }
+
+    async fn fetch_irises_resident_or_cold(&self, ids: &[VectorId]) -> Result<Vec<ArcIris>> {
+        let Some(cold) = &self.cold_storage else {
+            let store = self.iris_store.data.read().await;
+            return Ok(ids
+                .iter()
+                .map(|id| store.get_vector_or_empty(id).clone())
+                .collect());
```
dkales, line 1122: just thinking: Since this is behind a trait anyways can the cold-storage one not be a completely new struct, maybe wrapping the old worker pool, instead of doing a lot of selecting based on this option here?
(This was about the `Option<ColdStorage>`) in the worker)

iris-mpc-cpu/src/execution/hawk_main/search.rs
```diff
@@ -131,6 +215,456 @@ pub async fn search<const ROTMASK: u32>(
...
+        candidates = candidate_count,
+        "Running candidate-only linear-scan stage"
+    );
+    metrics::counter!("linear_scan_second_eye_candidates_total").increment(candidate_count as u64);
+    let second_results = linear_scan_eye(
+        sessions,
+        search_queries,
+        &search_params,
+        LinearScanEyeContext {
+            eye: second_eye_side,
```
dkales, line 317: Probably does not take too long but this is the whole DB id set, right? Maybe the registries can just keep this around themselves instead of computing it here fresh?
(This was about the live_id creation block in linear_search_cascade)

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
philsippl: it's only copying the arc handles only though? but agree, seems strictly better

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
philsippl: addressed now in #2351

iris-mpc/src/server/mod.rs
philsippl: yeah that's quite dramatic way of handling this. Batching doesn't bring any benefit and for the equivalence testing the batching could introduce non-determinism, but it should work out of the box (assuming the existing logic works). It's more like a deployment recommendation. But I also didn't run any tests with batchsize>1

philsippl: The message is also confusing and implies that this saves anything, will adapt the comment, but leave the check for safety (because i didn't test this path and it's also a bit harder to test)

iris-mpc-cpu/src/execution/hawk_main/worker_pool_initializer.rs
```diff
@@ -173,25 +182,43 @@ impl WorkerPoolInitializer for LocalWorkerPoolInitializer {
...
+                ),
+            };
+            Arc::new(worker) as Arc<dyn IrisWorkerPool>
         });
 
         let post_load_checksums = [
-            iris_stores[LEFT].data.read().await.set_hash.checksum(),
-            iris_stores[RIGHT].data.read().await.set_hash.checksum(),
+            registries[LEFT].read().await.set_hash.checksum(),
+            registries[RIGHT].read().await.set_hash.checksum(),
```
dkales, line 221: Just a nit that this no longer checks the right set at all in cold mode, since it is just a copy of the left registry

iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
```diff
@@ -543,6 +585,29 @@ where
...
     }
 
+    #[instrument(level = "trace", target = "searcher::network", skip_all)]
+    async fn full_rotation_dot_shares(
+        &mut self,
+        query: &Aby3Query,
+        vectors: &[VectorId],
+    ) -> Result<Vec<Share<u16>>> {
+        eyre::ensure!(
+            self.distance_fn.mode == DistanceMode::MinRotation,
```
dkales: Does this make conceptual sense? it is not really MinRotation distance mode here?

philsippl: it didn't, addressed

leonanos8 (approved):

leonanos8:
Covered by: https://github.com/worldcoin/iris-mpc/pull/2409

PR #2348, opened 2026-08-17
Fused mirror scan, packed pair kernel, and lane pipelining for the CPU linear scan

philsippl:
## Summary

Adds the compute, memory-layout, networking, and scheduling optimizations for the exact CPU linear scan. It is stacked on the correctness baseline (#2347) and cold-eye cache/prefetch layer (#2351); it does not change thresholds, rotations, MPC result semantics, or the public server contract.

## Actual server benchmark

Measured with the real `iris-mpc-linear-scan` service and production client across 3 × AWS `r8g.24xlarge`: TLS MPC networking, S3/SNS/SQS ingestion through Moto, persistence, normal + mirror scans, batch size 1, 31 rotations, and a 1,048,576-record seeded database.

| Build | Synchronized comparisons/s | Wall/request |
|---|---:|---:|
| Production-complete pre-follow-up build (12 measured requests; one warm-up excluded) | 4.983 M/s median (4.950–5.028 M/s) | 421 ms median |
| Intermediate doubled-row/UMMLA build | 5.65 M/s | 371 ms |
| This PR, paired/fused pipeline (`tokio=14`) | **6.33 M/s** | **331.5 ms** |
| Plus AMPC threshold-buffer reuse (companion revision) | 6.47 M/s | 324.5 ms |

“Logical comparisons” counts normal and mirror work once rather than once per MPC party. In the 4.983 M/s run, median harness end-to-end throughput was 1.492 M comparisons/s because Moto SNS/SQS fan-out added about 985 ms; synchronized scan time is the production compute/network measurement. At 6.33 M/s, an 18M-record normal + mirror scan extrapolates to about **5.7 seconds**, excluding ingress and persistence.

## How it works

- Stores the resident eye in a contiguous ARM mixed-plane layout and scans 4,096-record chunks across all 31 rotations.
- Represents each query with doubled rows, turning rotations into aligned windows and reducing the prerotated query footprint from about 1.19 MiB to 153.6 KiB.
- Uses NEON/UMMLA cross-packed pair kernels, including padded-tail handling and byte-identical scalar/reference parity checks.
- Evaluates normal and mirror orientations while streaming the resident eye once; the per-orientation MPC sessions and transcripts remain independent.
- Overlaps dot-product work with threshold MPC through bounded per-lane and global lookahead queues, and uses multiple physical TLS flows to avoid AWS's per-flow bandwidth ceiling.
- Preserves the #2351 rolling LUC window, mutation overlay, asynchronous cold-eye prefetch, and TinyLFU cache. Known candidates are excluded before prefetch reservation, preventing unconsumed slots from accumulating.
- Emits per-eye comparisons/s, scan duration, candidate fraction, cache/prefetch, and networking metrics suitable for comparison with the server benchmark.

Capacity estimate: the resident mixed-plane eye costs about **38.55 kB per record**, including container, allocator, registry, and 20% store-reserve overhead. The 501-record rolling LUC window, 4,096-entry TinyLFU, and one 4,096-record prefetch chunk add about **0.32 GiB**. On an `r8g.24xlarge` with the configured 700 GiB pod limit, the recommended operational maximum is **18–18.5M records** (about 646.5–664.5 GiB); the roughly 19.49M mathematical ceiling leaves no safe runtime margin.

## Deployment

1. Merge/publish [worldcoin/ampc-common#138](https://github.com/worldcoin/ampc-common/pull/138), keep all four AMPC pins on the same revision, and build `ghcr.io/worldcoin/iris-mpc-linear-scan` using `.github/workflows/build-and-push-linear-scan-server.yaml`. Pin production to an immutable image digest.
2. Provision one dedicated on-demand `r8g.24xlarge` per party and use the supplied stage/prod common values plus party overlays. They request 92 CPUs without a CPU limit and select/tolerate the dedicated linear-scan nodes.
3. Apply `deploy/aws/r8g-24xlarge-linear-scan-user-data.mime` through the node class, launch template, or baked AMI. It keeps ENA interrupts on CPUs 0–10 and reserves CPUs 11–95 for dot-product workers; verify IRQ placement under load.
4. Preserve `hostNetwork: true`, allow direct cross-party TCP ports 4000–4002, and keep 16 physical connections so traffic is not capped by a single AWS 5-Gbit/s flow.
5. Verify `SMPC__MAX_BATCH_SIZE=1`, the same `SMPC__FULL_SCAN_SIDE` on every party, `SMPC__FULL_SCAN_SIDE_SWITCHING_ENABLED=false`, all three `/ready` endpoints, a real request, and the scan/cache metrics.
6. Drain/stop all three GPU consumers first, then deploy all three CPU parties together with `Recreate`. Never run mixed images or both queue consumers concurrently. Alternate the resident eye only by changing `Left`/`Right` on all parties and performing another coordinated restart.

## Review guide

1. Review paired cascade scheduling and bounded lookahead in `iris-mpc-cpu/src/execution/hawk_main/search.rs`.
2. Review resident layout and cache/prefetch interaction in `worker_pool_initializer.rs` and `iris_worker.rs`.
3. Review doubled-query and UMMLA kernels in `iris-mpc-cpu/src/protocol/ops.rs` plus fused MPC primitives in `ampc-common`.
4. Review server/network tuning and the real-server benchmark tooling.
5. Finish with packed-kernel parity, paired-vs-independent cascade, GPU parity, cold-eye, and deterministic full-server tests.

## Stack

- #2347 — correctness baseline
- #2351 — cold-eye cache and prefetch
- **this PR** — compute and pipeline performance


iris-mpc-cpu/src/protocol/shared_iris.rs
```diff
@@ -126,3 +126,148 @@ impl GaloisRingSharedIris {
...
+/// 8-byte-interleaved lo/hi plane representation of an iris share for the
+/// UMMLA-based exact-scan kernel: every 8 consecutive u16 coefficients are
+/// stored as `[lo0..lo7 | hi0..hi7]`. Same byte count as the u16 form; the
+/// original share is reconstructed exactly by [`MixedPlaneIris::to_iris`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MixedPlaneIris {
+    /// `id` of the code share (party_id + 1), preserved for reconstruction.
+    code_id: usize,
+    /// `id` of the mask share (party_id + 1), preserved for reconstruction.
+    mask_id: usize,
```
dkales: This is never out of sync or am I missing something? why are we storing this twice?

iris-mpc-cpu/src/hawkers/aby3/aby3_store.rs
```diff
@@ -68,6 +70,38 @@ pub type Aby3Query = QuerySpec;
 pub type Aby3DistanceRef<T = u32> = DistanceShare<T>;
 pub type RotationMatchIndices = Vec<Vec<usize>>;
 
+/// Both orientations' additive dot-product shares for one chunk.
+pub type PairDotContributions = [Vec<RingElement<u16>>; 2];
+
+/// A spawned task handle that aborts the task when dropped, so a lookahead
+/// chunk cannot keep running detached after the lane that requested it has
+/// failed or been cancelled.
+pub struct AbortOnDropHandle<T>(tokio::task::JoinHandle<T>);
```
dkales: This is in tokio-util: https://docs.rs/tokio-util/latest/tokio_util/task/struct.AbortOnDropHandle.html

iris-mpc-cpu/src/protocol/ops.rs
```diff
@@ -1099,6 +2184,138 @@ mod tests {
         }
     }
 
+    #[cfg(target_arch = "aarch64")]
```
dkales, line 2008: These will not run in the CI pipeline I guess?

iris-mpc-cpu/src/protocol/ops.rs
```diff
@@ -608,6 +659,51 @@ fn accumulate_scalar_target<const ROTATIONS: usize>(
     }
 }
 
+/// One rotation against four targets: the remainder tile after the 6-wide
+/// (and optional 4-wide) passes. Keeping the four targets in one pass reuses
+/// each query vector load for four MLAs instead of evaluating four
+/// independent dot products.
+#[cfg(target_arch = "aarch64")]
+#[inline(always)]
+fn dot_product_1x4_u16(query: &[u16], targets: [&[u16]; 4]) -> [u16; 4] {
```
dkales: These could all be one const-generic function, no? I would hope the compiler can unroll this and produce equivalent code.

iris-mpc-cpu/src/execution/hawk_main/iris_worker.rs
```diff
@@ -1951,7 +2228,7 @@ impl IrisWorkerPool for LocalIrisWorkerPool {
             // visible in the store immediately after this returns.
             let mut store = iris_store.data.write().await;
             for (vector_id, iris) in resolved {
-                store.insert(vector_id, iris);
+                store.insert(vector_id, ResidentIris::from_arc(iris, layout));
```
dkales: The from_arc does the conversion under the write lock here. I guess this is not really that much of an issue though

iris-mpc-cpu/src/protocol/ops.rs
```diff
@@ -839,6 +935,995 @@ pub fn non_existent_distance() -> Vec<RingElement<u16>> {
...
+        }
+    }
+
+    #[inline(always)]
+    unsafe fn query_window<const ROTATIONS: usize>(
+        query: &[u8],
+        row: usize,
+        rotation: usize,
+    ) -> *const u8 {
+        let amount = PrerotatedQueryRowMajorView::<ROTATIONS>::ROTATION_AMOUNTS[rotation];
```
dkales, line 1281: This could use a const assert that this is a multiple of 4.

iris-mpc-cpu/src/protocol/ops.rs
```diff
@@ -553,40 +593,51 @@ fn accumulate_component_tiled_6x4<const ROTATIONS: usize>(
...
-                        target_row,
-                        target_idx + target_offset,
-                        result_lane,
-                        additive_shares,
-                    );
-                }
             }
-            target_idx += 4;
+        } else {
+            // Missing vectors are uncommon in a full scan. Preserve their
```
dkales, line 616: I guess this is true, but maybe this is worth a metric emmission since this probably tanks performance if it happens often.

iris-mpc-cpu/src/protocol/ops.rs
philsippl: yeah there is the fallback impl, but i guess would be nice to actually test this in CI, pushing extra job

iris-mpc-cpu/src/protocol/ops.rs
philsippl: Will bench again to verify this, it grinded quite a bit on the dot product impl

philsippl: You're correct! This is not beating the compiler, I guess this was conflating improvements due to layout with "hand"written asm. Removed the custom impl as well as the CI job again.

leonanos8:
covered by; https://github.com/worldcoin/iris-mpc/pull/2409

PR #2382, opened 2026-09-04
INFRA-7144: add ampc-hnsw values for internal-tools dev

luafanti:
## Requestor/Issue:
INFRA-7144

## Tested (yes/no):
no — runtime validation happens on first workflow dispatch in cluster-apps

## Description/Why:
Chart values for the three ampc-hnsw participants in TFH internal-tools dev (eu-west-1). Path `deploy/internal-tools-dev/` is deliberately distinct from `deploy/dev/` (same cluster names, different AWS accounts).

Diffs vs `deploy/dev/`: new Route53 zones (`ampc-hnsw-<N>.dev.internal.worldcoin.org`), `attachSecurityGroupPolicy.enabled: false` (CIDR-based ingress from terraform-mpc-setup), env `internal-tools-dev`, region `eu-west-1`, non-snapshot Aurora URL, PPROF/DB-chunks knobs dropped.


jazzbee:
Those values were moved to repo with ArgoCD configuration

PR #2402, opened 2026-09-14
[POP-4319] Make complete exports generation-atomic

carlomazzaferro:
Complete exports now write every chunk under an immutable generation prefix and publish a versioned marker only after every chunk succeeds. The importer accepts both legacy and generation markers; once a generation exists, incremental requests safely fall back to another complete generation instead of reintroducing mutable chunks.

Tested with exporter unit/race tests, `go vet`, `cargo test -p iris-mpc-store`, and clippy.


carlomazzaferro:
good catch. fixed in `0c465c4`: v2 markers now use Unix nanoseconds for the ordering field, so invocation time wins; the random generation ID is only the deterministic fallback for an exact timestamp tie.

carlomazzaferro:
@codex review

carlomazzaferro:
Closing this after tracing the importer path end to end. The timestamp marker is the Aurora replay checkpoint. On startup, the loader reads S3 and then overwrites those records with every Aurora row modified since the selected marker, minus the safety overlap. If a complete export fails, the marker stays old, so replay remains conservative across any individually complete chunks written by that run.

This means #2392 and #2401 already preserve the required invariants: producer failures propagate, a failed stream cannot replace a good chunk, and the marker advances only after every batch succeeds. The generation protocol adds a second cross-language format and ordered rollout without improving correctness under the existing recovery model.

The current v2 marker also uses Unix nanoseconds for ordering, while the loader passes that value into the seconds-based last_modified_at predicate. That suppresses the Aurora replay the loader relies on, which is another reason not to merge this version.

